PR #1, opened 2015-04-14
JWK kid generator plus basic api endpoint

kalabiyau:


spec/vendor/kid_generator.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/docker/libtrust"
+	"os"
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Fprintf(os.Stderr, "Usage: kid_generator <public_key>\n")
+		os.Exit(1)
+	}
+	fmt.Println(os.Args)
```
flavio: This can be removed, I forgot it...

flavio:
Also let's remove `spec/vendor/kid_viewer`, we will build that on travis CI later.

spec/vendor/kid_generator.go
kalabiyau: That's I've added to look what is passed, but of course we can remove it

app/controllers/api/v2/tokens_controller.rb
```diff
@@ -0,0 +1,16 @@
+class Api::V2::TokensController < Api::BaseController
+
+  def show
+    token_request_params
+    token_request_params.permitted?
```
flavio: We totally ignore the return value of `permitted?`. Seems useless.

PR #3, opened 2015-04-15
Introducing user authentication

flavio:
* Use [devise](https://github.com/plataformatec/devise) to implement user authentication.
* Return an authentication token only to the valid users


app/controllers/users/registrations_controller.rb
```diff
@@ -0,0 +1,60 @@
+class Users::RegistrationsController < Devise::RegistrationsController
+ before_filter :configure_sign_up_params, only: [:create]
+# before_filter :configure_account_update_params, only: [:update]
+
+  # GET /resource/sign_up
```
kalabiyau: I think we can cleanup commented stuff, as there is not much value in them

spec/api/v2/token_spec.rb
```diff
@@ -3,21 +3,55 @@
 describe '/v2/token' do
 
   describe 'get token' do
+    before :all do
```
kalabiyau: I would suggest we use factory girl for creation of users, as that is way more convenient rather than updating all the resource creation calls after update of a signature of a resource

kalabiyau: More than that - you already have factory and you can use

```ruby
let(:user) { create(:user) }
```

just after describe

app/views/devise/registrations/edit.html.erb
```diff
@@ -0,0 +1,39 @@
+<h2>Edit <%= resource_name.to_s.humanize %></h2>
```
kalabiyau: Can we convert it to Slim to have only one templating language in the project?

kalabiyau: https://github.com/slim-template/html2slim

not perfect, but good approximation

kalabiyau: I am on it, so no worries

PR #5, opened 2015-04-15
Fixed re-auth attempt when jwt does not have enough scope to permorm action

kalabiyau:


config/routes.rb
```diff
@@ -5,6 +5,9 @@
 
   namespace :v2, module: 'api/v2', defaults: { format: :json } do
     resource :token, only: [ :show ]
+    resource :webhooks, only: [] do
+      resources :events, only: [ :create ]
```
kalabiyau: @flavio I've settled the events endpoint on `/v2/webhooks/events` - pure RESt

PR #6, opened 2015-04-16
Introducing Repositories, Images and Tags

flavio:
Introduce the Repositories, Images and Tags entities:
  * A repository has many images.
  * An image belongs to a repository.
  * An image has many tags.

The database data is populated by parsing the 'push' events sent by the
registry.


app/assets/javascripts/images.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
kalabiyau: Those guys we definitely need to drop

app/assets/stylesheets/scaffolds.scss
```diff
@@ -0,0 +1,69 @@
+body {
```
kalabiyau: If that one is also from scaffold - we should drop as new design is emerging

kalabiyau:
@flavio I would say that everything you didn't created intentionally - should be dropped. As that is just scaffolding noise, which gives zero benefit

flavio:
I'll take care of that.

app/controllers/api/v2/events_controller.rb
```diff
@@ -2,7 +2,10 @@ class Api::V2::EventsController < Api::BaseController
 
   def create
     event = JSON.parse(request.body.read)
-    ap event
+    push_events = event['events'].find_all do |e|
```
kalabiyau: I would suggest to build Event instance from params and then in it after init - perform all the logic which is here right now. So controller will be fatfree:

```rails
Event.new(JSON.parse(request.body.read))
head status: :accepted
```

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,23 @@
+class ImagesController < ApplicationController
+
+  # GET /images
+  # GET /images.json
+  def index
+    @images = Image.all
+
+    respond_to do |format|
+      format.html # index.html.erb
+      format.json { render json: @images }
```
kalabiyau: I think we don't need standard format closures but rather better to use respond_to :html and respond_with

gem called responders

app/controllers/api/v2/events_controller.rb
flavio: Sounds good to me. Should I write the `Event` class under the models' directory?

kalabiyau:
Other than those small notes - looks good

kalabiyau:
great job!

app/controllers/api/v2/events_controller.rb
kalabiyau: I would do so - same as for JwtToken class

app/helpers/repositories_helper.rb
```diff
@@ -0,0 +1,2 @@
+module RepositoriesHelper
```
kalabiyau: Helpers we don't need as well

app/models/image.rb
```diff
@@ -0,0 +1,32 @@
...
+  has_many :tags
+
+  def self.handle_push_event(event)
+    if event['target']['repository'].include?('/')
+      repo_name, image_name = event['target']['repository'].split('/', 2)
+    else
+      image_name = event['target']['repository']
+    end
+
+    match = %r|manifests/(?<tag>.*)$|.match(event['target']['url'])
```
kalabiyau: Better to have defined regexes extracted to constants an use them within code

db/schema.rb
```diff
@@ -11,7 +11,31 @@
...
+ActiveRecord::Schema.define(version: 20150416122030) do
+
+  create_table "images", force: :cascade do |t|
+    t.string   "name",          default: "", null: false
+    t.integer  "repository_id"
+    t.datetime "created_at",                 null: false
+    t.datetime "updated_at",                 null: false
+  end
+
+  add_index "images", ["name"], name: "index_images_on_name", unique: true
```
kalabiyau: we are missing index on repository_id

db/schema.rb
```diff
@@ -11,7 +11,31 @@
...
+    t.string   "name"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  add_index "repositories", ["name"], name: "index_repositories_on_name", unique: true
+
+  create_table "tags", force: :cascade do |t|
+    t.string   "name",       default: "latest", null: false
+    t.integer  "image_id",                      null: false
```
kalabiyau: Missing index on image_id

db/schema.rb
flavio: Fixed.

db/schema.rb
flavio: Fixed

flavio:
Ok, all the tests are passing now. I fixed all the issues you pointed out.

If you are fine with it I would like to do an interactive rebase inside of this branch, squash all the commits into a single one and then merge that single commit back into master.

PR #8, opened 2015-04-17
Use header and implement logout button

kalabiyau:


app/controllers/application_controller.rb
```diff
@@ -8,4 +8,8 @@ def after_sign_in_path_for(_resource)
     dashboard_url
   end
 
+  def after_sign_out_path_for(_resource)
+    new_user_session_url
```
flavio: Wouldn't it be better to redirect to the homepage of Portus?

flavio:
LGTM to me except for the single comment I made.

app/controllers/application_controller.rb
kalabiyau: If we would have one. We have now set of auth endpoints and a root path in app points to a dashboard
So we don't yet have a defined homepage, so for now it should be fine

PR #13, opened 2015-04-20
Add class to interact with a Docker Registry server

flavio:
Add Registry class. This can be used to fetch additional information
from a Docker registry v2 using its API.

Right now the Registry class can authenticate against the registry and
fetch the manifest of a Docker image available on the it.

### DIFF
--- a/app/models/registry.rb
+++ b/app/models/registry.rb
@@ -1,0 +1,110 @@
+class Registry
+  class NoBearerRealmException < RuntimeError
+  end
+
+  class AuthorizationError < RuntimeError
+  end
+
+  class ManifestNotFoundError < RuntimeError
+  end
+
+  class CredentialsMissingError < RuntimeError
+  end
+
+  def initialize(host, use_ssl=true, username=nil, password=nil)
+    @host = host
+    @use_ssl = use_ssl
+    @base_url = "http#{'s' if @use_ssl}://#{@host}/v2/"
+    @username = username
+    @password = password
+  end
+
+  def has_credentials?
+    @username && @password
+  end
+
+  def manifest(repository, tag='latest')
+    res = get_request("#{repository}/manifests/#{tag}")
+    if res.code.to_i == 200
+      JSON.parse(res.body)
+    elsif res.code.to_i == 404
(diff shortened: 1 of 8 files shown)


app/models/registry.rb
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env ruby
```
kalabiyau: I think as that probably will be a gem at some point - it is better to settle it in the vendor

app/models/registry.rb
```diff
@@ -0,0 +1,113 @@
...
+      fail ManifestNotFoundError, "Cannot find manifest for #{repository}:#{tag}"
+    else
+      fail "Something went wrong while fetching manifest for #{repository}:#{tag}:" \
+        "[#{res.code}] - #{res.body}"
+    end
+  end
+
+  def get_request(path, request_auth_token=true)
+    uri = URI.join(@base_url, path)
+    req = Net::HTTP::Get.new(uri)
```
kalabiyau: You can sit on top of RestClient for ruby which gives you all the goodies out of the box

app/models/registry.rb
flavio: Yeah, I thought that as well.

Thanks for the catch, I forgot to remove these lines.

app/models/registry.rb
flavio: I tried to use it, unfortunately I wasn't able to get the headers out of the first unauthorized error response...

app/models/registry.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  def request_auth_token(unhauthorized_response)
+    bearer_realm, query = parse_unhauthorized_response(unhauthorized_response)
+
+    uri = URI("#{bearer_realm}?#{query.to_query}")
+
+    req = Net::HTTP::Get.new(uri)
+    req.basic_auth(@username, @password) if has_credentials?
+
+    res = Net::HTTP.start(
```
kalabiyau: Same here better to use solid solution as this is a golden stardard in Ruby more or less

kalabiyau:
@flavio anyhow, those are just nitpicks, other than coverage below 100% it is good to go, IMO

Issue #16, opened 2015-04-22
Namespace should have different ACL for pull and push requests


flavio:
Right now a namespace will accept push and pull requests only from the owner of the team associated with the namespace.

It must be possible to allow pulls also from registered users outside of the team.

flavio:
> Right now a namespace will accept push and pull requests only from the owner of the team associated with the namespace.

Implemented.

The only missing piece is: add a bool flag to the namespace to allow RO access from users that are not members of the team.

PR #29, opened 2015-04-29
Team roles

flavio:
This PR introduces the concept of user roles within a team, as requested by the 1st milestone (see issue #16).

A team can have three types of users:
  * viewers: RO access
  * contributors: RW access
  * owners: like contributors, but have also access to the admin panel of the team

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -4,10 +4,13 @@ class Team < ActiveRecord::Base
 
   has_many :namespaces
 
-  # Users & owners
   has_many :team_users
   has_many :users, through: :team_users
-  has_many :owners, -> { where 'team_users.owner': true },
+  has_many :owners, -> { where 'team_users.role' => TeamUser.roles['owner'] },
+    through: :team_users, source: :user
+  has_many :contributors, -> { where 'team_users.role' => TeamUser.roles['contributor'] },
+    through: :team_users, source: :user
+  has_many :viewers, -> { where 'team_users.role' => TeamUser.roles['viewer'] },
     through: :team_users, source: :user
 
   before_create :downcase?
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -15,6 +15,10 @@ class Team < ActiveRecord::Base
 
   before_create :downcase?
 
+  def create_team_namespace!
+    Namespace.find_or_create_by!(team: self, name: name)
+  end
+
   private
 
   def downcase?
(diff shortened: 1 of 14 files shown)


flavio:
@mssola what do you think about that, do you have any suggestion before I merge the code?

app/models/team.rb
```diff
@@ -4,14 +4,21 @@ class Team < ActiveRecord::Base
...
 
-  # Users & owners
   has_many :team_users
   has_many :users, through: :team_users
-  has_many :owners, -> { where 'team_users.owner': true },
+  has_many :owners, -> { where 'team_users.role' => TeamUser.roles['owner'] },
+    through: :team_users, source: :user
+  has_many :contributors, -> { where 'team_users.role' => TeamUser.roles['contributor'] },
+    through: :team_users, source: :user
+  has_many :viewers, -> { where 'team_users.role' => TeamUser.roles['viewer'] },
```
mssola: I believe that `viewers` won't be used. That's because any member of the team has effectively pull access, so the viewers relationship is redundant here (see the `NamespacePolicy::pull?` method from your PR).

db/migrate/20150429102443_add_role_to_team_user.rb
```diff
@@ -0,0 +1,6 @@
+class AddRoleToTeamUser < ActiveRecord::Migration
+  def change
+    add_column :team_users, :role, :integer, default: 0
```
mssola: IIRC, you can set the default to:
```ruby
TeamUser.statuses[:viewer]
```
It would be safer, imho

app/models/team.rb
flavio: Good point, they are useless.

I'm going to remove them.

Anything else? In the meantime thanks

mssola:
Besides my nitpicking, LGTM :+1:

db/migrate/20150429102443_add_role_to_team_user.rb
flavio: Good point

flavio:
I thought a bit more about the `viewers` role. I think we should leave it because it will make things easier inside of the UI. Let me explain a bit.

Scenario: allow user X to pull repositories from Team Y namespace.

As a team owner for team Y I go to the team administration page
and I go to edit members page
and I click add new user
and I enter X as username
and I select 'viewer' as role

When adding a new user to a group the UI will show three possible values: owner, contributor and viewer.

mssola:
Makes sense :+1:

PR #32, opened 2015-05-06
Implement UI for namespaces and teams

flavio:
This PR adds some missing UIs plus some changes to the sidebar and the dashboard page.

## Sidebar

Entries of the sidebar:
  * dashboard: still points to the dashboard of Portus
  * Namespaces: points to the page showing all the namespaces the user has access to. That include its personal namespace, the ones controlled by teams he's member of and also the public ones.
  * Teams: points to the "list all the teams page"; see below.

sidebare

## Dashboard

Nothing fancy to show. The old page contained just a message "Servus!". Now we just write how many teams the user is member of. We will provide more meaningful contents on another iteration.

## Team

All these operations keep into account the user relation with existing teams
and prohibit unauthorized from messing with the system.

### List all the teams

This page lists all the teams the current user is member of.

team index

### Show details of a team

team show

### Create a new team.

team create

### Add new users to a team.

This is possible from the page of the team. There's no fancy auto-completion for the usernames yet. 

team add user

### Edit members of a team.

Changing the role of a team member is done from the team page. Portus won't allow the only owner of the team to change its role to something else.

team change user role

Removing a user from the team is just a matter of clicking the trash button. There's a confirmation dialog that is not yet styled.

## Namespaces

Again, user rights have been taken into account.

### List all the namespaces

This list all the namespaces that are either public or are controlled by a team to which the user belongs.

namespace index

### Create new namespaces

Since namespaces must belong to a team they can be created only from the team page.

namespace add

### Change the 'public' attribute

This is done by toggling the highlighted switch. This operation can be done only by owners of the team controlling the namespace; other users cannot click the switch.

namespace public switch


flavio:
@mssola, @kalabiyau: can you take a look at it?

There's one thing I do not understand: when rails runs in development mode the `teams.coffeescript` file is not served as a separate asset. Rails serves one single js file containing all the js unminified code. I wasn't able to figure out what is causing that.

mssola:
Good work! :+1: 

First, let me focus on the UX before diving into the code :)

Let me start with a very subtle thing. In the Namespaces tab, the elements of the table are not vertically-aligned properly. So we have this example:
miss
The problem is that `td` elements are vertically aligned to the top, but the switch apparently is not. This can be easily solved by setting `vertical-align: middle;` to these elements. Then it looks like this:
fixed
As far as I can see, this only happens in this table (e.g. the tables on the team edit page are just fine)

I have some issues with the /teams/new page:

create

First of all, the button should be blue, since this is the case for the rest of the submit buttons in this application. Moreover, all the text seems a bit redundant. Personally I would:
1. Remove the `Name` label.
2. Replace the current placeholder from "Enter name" to "Name".
3. Rename the submit button to "Create".
What do you think about this?

More subtilities :) When adding a new member to a team, the `Role` label and the select element are misaligned. See:

aligned

I can reproduce this with both Firefox and Chrome. Moreover, in this same form, the placeholder for the username input should be capitalized, so it's consistent with the other placeholders in this application. I'd suggest renaming it from `username` to just `Name`, since the "user" part is in its label already (and plus, this way it would be consistent with the placeholder of the form that creates namespaces ;) ). By the way, the placeholder of the form for creating namespaces should also be capitalized, so it's consistent with the rest of the application.

Another area for improvement, in my opinion, is the table for managing members of a team. First of all, when we click the edit button we get this:
edit
This has some minor issues:
1. It feels "too crowded". I'd hide the current role, since it's already the default value of the select element being shown. But remember to show it again when getting out of "edit mode" ;)
2. There should be a (tiny) margin between the label and the select box.
3. I believe that the Edit button should reflect in some way that we are in editing mode. Or at least there should be an instantly obvious way to quit "edit mode".
4. As a note: the "Save" button is probably not really needed, we can change the value by listening to the `change` event. That being said, I understand that this can lead to dangerous situations, so I'm fine with it.

mssola:
@flavio for your problem with the teams.coffee file, maybe you can try setting `config.assets.debug = true` in the `config/environments/development.rb` file?

app/assets/javascripts/teams.js.coffee
```diff
@@ -0,0 +1,24 @@
+$(document).on "page:change", ->
+  $('#add_team_user_btn').on 'click', (event) =>
+    $('#add_team_user_form').fadeToggle 400, "swing", ->
+      if $('#add_team_user_form').css("display") == "block"
```
mssola:
```js
if $('#add_team_user_form').is(':visible')
```

app/assets/javascripts/teams.js.coffee
```diff
@@ -0,0 +1,24 @@
...
+        $('#add_team_user_btn i').removeClass("fa-chevron-up")
+        $('#add_team_user_btn i').addClass("fa-chevron-down")
+
+  for btn_edit_role in $(".btn-edit-role")
+    $(btn_edit_role).on 'click', (event) =>
+      $('#change_role_team_user_' + event.currentTarget.value).fadeToggle()
+
+  $('#add_namespace_btn').on 'click', (event) =>
+    $('#add_namespace_form').fadeToggle 400, "swing", ->
+      if $('#add_namespace_form').css("display") == "block"
```
mssola: same as my previous note

mssola:
And that's all from me for now. Nicely done! Just take a look at the small issues that I've raised :)

flavio:
> @flavio for your problem with the teams.coffee file, maybe you can try setting config.assets.debug = true in the config/environments/development.rb file?

That fixed the issue. Thanks.

flavio:
I fixes all the glitches you mentioned. Thanks for the feedback!

mssola:
I have a couple of small issues when adding namespaces. First of all, the placeholder should be capitalized. Second of all, the "Add namespace" element is broken. It works when the form is hidden, but it no longer works once the form has been shown.

app/assets/javascripts/teams.js.coffee
```diff
@@ -0,0 +1,32 @@
...
+        $('#add_team_user_btn i').addClass("fa-chevron-up")
+        $('#add_team_user_btn i').removeClass("fa-chevron-down")
+      else
+        $('#add_team_user_btn i').removeClass("fa-chevron-up")
+        $('#add_team_user_btn i').addClass("fa-chevron-down")
+
+  for btn_edit_role in $(".btn-edit-role")
+    $(btn_edit_role).on 'click', (event) =>
+      $('#team_user_' + event.currentTarget.value + ' td .role').fadeToggle()
+      $('#change_role_team_user_' + event.currentTarget.value).fadeToggle()
```
mssola: I think it looks much nicer if you replace these two `fadeToggle` with simply `toggle`

app/assets/javascripts/teams.js.coffee
```diff
@@ -0,0 +1,32 @@
...
+      if $('#add_namespace_form').css("display").is(':visible')
+        $('#add_namespace_btn i').addClass("fa-chevron-up")
+        $('#add_namespace_btn i').removeClass("fa-chevron-down")
+      else
+        $('#add_namespace_btn i').removeClass("fa-chevron-up")
+        $('#add_namespace_btn i').addClass("fa-chevron-down")
+
+  $('#add_team_btn').on 'click', (event) =>
+    $('#add_team_form').fadeToggle 400, "swing", ->
+      if $('#add_team_form').css("display") == "block"
```
mssola: .is(':visible') ;)

app/assets/javascripts/teams.js.coffee
```diff
@@ -0,0 +1,32 @@
...
+        $('#add_team_user_btn i').addClass("fa-chevron-down")
+
+  for btn_edit_role in $(".btn-edit-role")
+    $(btn_edit_role).on 'click', (event) =>
+      $('#team_user_' + event.currentTarget.value + ' td .role').fadeToggle()
+      $('#change_role_team_user_' + event.currentTarget.value).fadeToggle()
+
+  $('#add_namespace_btn').on 'click', (event) =>
+    $('#add_namespace_form').fadeToggle 400, "swing", ->
+      if $('#add_namespace_form').css("display").is(':visible')
```
mssola: AFAIK you don't have to call css('display'). The following should simply work:
```coffee
if $('#add_namespace_form').is(':visible')
```

app/assets/javascripts/teams.js.coffee
flavio: Stupid copy & paste....

app/assets/javascripts/teams.js.coffee
flavio: Good point

flavio:
Ok, I fixed all issues you mentioned + another issue created by @eotchi

Issue #37, opened 2015-05-06
Introduce "owned by" team in the namespace detailed view


eotchi:
By clicking on a namespace presented in the namespace tab, the name of the namespace will be presented in the header of the table:

![screenshot from 2015-05-06 19 35 04](https://cloud.githubusercontent.com/assets/576621/7499318/19ef70a2-f427-11e4-9e89-679e386e9b15.png)

This should be accompanied by a "owned by" followed by the name of the team to which the namespace belongs to.
In the best case (and not strictly necessary), for personal namespaces this could be omitted, not to display entries like 
**Namespace federica owned by federica**

mssola:
@cyntss what do you think about this. What's the best way to accomplish this from a UI/UX point of view ?

cyntss:
Fixed in: ec757f349257ba6b4e6101ea6f8502e7ae3ef072

PR #41, opened 2015-05-07
Show the 'owned by' information when displaying namespaces

flavio:
This fixes issue #37 

## Namespaces index

I don't know how to hide the owner for personal repositories in a nice way...

namespace index

### Personal namespace 

No information is shown

personal namespace

### Non personal namespace

Information is shown. I am not really happy about how we are showing it though,

team namespace

### DIFF
--- a/app/views/namespaces/_namespace.html.slim
+++ b/app/views/namespaces/_namespace.html.slim
@@ -1,5 +1,7 @@
 tr[id="namespace_#{namespace.id}"]
   td= link_to namespace.name, namespace
+  - if show_owner
+    td= link_to namespace.team.name, namespace.team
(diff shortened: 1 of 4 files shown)


flavio:
@eotchi do you have any idea?

mssola:
One idea: since a namespace can only be owned by one team, why don't we just put it in a title just like in the image below?
snapshot1
This is a quick example, but the name of the team in my example should be a link (but even if it's a link I would keep the grey color). What do you think?

flavio:
I fear this could lead to some confusion. Users might think they have to do `docker pull team/namespace`.

Right now @cyntss is looking into the whole UI of Portus. I think we can close this bug.

PR #46, opened 2015-05-07
UI: review how to change the public attribute of a namespace

flavio:
Made everything more consistent with the rest of the UI.

## When showing namespaces

The 'edit' button is enabled only for the users with the appropriate rights.

namespace edit index

## Edit mode enabled

The old switch is shown

namespace edit mode

## Switch clicked

A notification is shown (like before). The switch disappears and the "eye" icon is shown, the value of the eye reflects the new status of the namespace.


flavio:
@eotchi what do you think about it?

mssola:
@flavio: besides being outdated, I don't think that this change is relevant anymore, no?

flavio:
You are right, let's close it.

Issue #44, opened 2015-05-07
Sanitize the namespace automatically created upon team creation


flavio:
Creating a team named "spaces are not allowed" results into the creation of a namespace named "spaces are not allowed".

flavio:
Done

PR #52, opened 2015-05-11
Ensure User#username and Namespace#name are always valid

### LINKED ISSUE #42: Namespace creation UI
No validation is being done apparently (e.g.: usernames with spaces are accepted).

Docker validates namespace's name using this regexp: `[a-z0-9-_]`

flavio:
The Docker daemon performs some validations for the User's username and
also for the Namespace's name. We have to apply the same policies
inside of Portus.

Fixes #42 and #44

### DIFF
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -1,8 +1,18 @@
 class Namespace < ActiveRecord::Base
 
+  NAME_ALLOWED_CHARS = 'a-z0-9\-_'
+
   has_many :repositories
   belongs_to :registry
   belongs_to :team
-  validates :name, presence: true
+  validates :name,
+            presence: true,
+            format: {
+              with: /\A[#{NAME_ALLOWED_CHARS}]+\Z/,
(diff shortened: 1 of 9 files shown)


flavio:
@mssola what do you think about it?

app/models/namespace.rb
```diff
@@ -1,8 +1,15 @@
 class Namespace < ActiveRecord::Base
 
+  NAME_ALLOWED_CHARS = 'a-z0-9\-_'
+
   has_many :repositories
   belongs_to :registry
   belongs_to :team
-  validates :name, presence: true
+  validates :name, presence: true,
+                   format: { with: /\A[#{NAME_ALLOWED_CHARS}]+\Z/, message: 'Only allowed letters: [a-z0-9-_]' }
```
mssola: Nitpicking: maybe you could write the `message` key on the next line, just like you did in the user.rb file :)

mssola:
LGTM :+1:

flavio:
Fixed, thanks for the feedback

PR #55, opened 2015-05-12
review gems for rpms

jordimassaguerpla:
update loofah and minitest gem to latest version because that is the
version we have under the repo obs://devel:languages:ruby:extensions

(obs=opensuse build service . see build.opensuse.org)

It is easier to update the gems than to have to version of the same gem
in the RPMs.


flavio:
LGTM

PR #56, opened 2015-05-12
review Gemfile for packaging

jordimassaguerpla:
When packaging this application as for example an RPM, we need to know
which gems are needed, which means we don't want to have devel or test
gems.

While you can run "bundle install --without development test", this
does not update the Gemfile.lock, meaning the develoment and test gems
are still in Gemfile.lock.

If you infere the gem dependencies from the Gemfile.lock, you would get
the development and test gems, unless you exclude them inside the
Gemfile file.

This commits adds a "unless ENV['PACKAGING'] == 'yes'" that excludes the
gems in the Gemfile thus, if you run

PACKAGING=yes bundle list

you will get the Gemfile.lock that you need without the test and
development dependencies and thus you can use that for infering which
gems you will need in production and thus set the package requirements.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,35 +25,45 @@ gem 'rails-observers'
 #TODO remove as soon as we migrate to PostgreSQL
 gem 'sqlite3'
 
-group :development do
-  gem 'quiet_assets'
-  gem 'pry-rails'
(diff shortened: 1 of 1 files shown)


flavio:
LGTM

PR #61, opened 2015-05-14
Audit / activities

flavio:
Introduced the ability to track the following events:
  * Team:
    - create
    - add new member
    - remove member
    - change role member
  * Naspace:
    - create
    - public -> private
    - private -> public
  * Repository:
    - push tag

All these events are stored inside of the database.

Non-admin users can watch the stream of events on their landing page (I've re-estabilished
the dashboard page). The flow is tailored to the user; hence only events related
with namespaces and teams he's involved with. Events happening inside of public
namespaces are always reported. Also events triggered by a change of the `public`
attribute of a namespace are shown to everybody.

I have also changed Admin's dashboard page to show the events of the last 24 hours.
A full list of the events can be see on a new dedicated page. There's also the
possibility to download all the events as a CSV file.

## Mandatory screenshot

This is just a small extract from the admin page, the user page is similar.

screenshot-localhost 5000 2015-05-14 12-24-05

## Missing

Caching, this is coming with another PR.

## Heads up

@jordimassaguerpla: once this PR gets merged into master you will have to package two new gems.

@mssola: as always, feedback is welcome


mssola:
The code looks good to me. For the UI I have a couple of minor issues:

1. In the Dashboard, page, the title should be "Recent Activities" instead of "Dashboard". This way the current behavior is more clearly reflected in the UI. Moreover, this is consistent with the admin's private dashboard.
2. In the admin's dashboard, there should be a margin between the title "Recent Activities" and the link "[All]". See the following snapshot:

snap

flavio:
I fixed the UI glitches. Thanks for your help!

flavio:
@jordimassaguerpla : you have two new gems to package ;)

PR #63, opened 2015-05-15
Introducing the special 'portus' user

flavio:
Soon we are going to fetch informations from the registry itself (like
repositories' manifests). When doing that we have to authenticate
against the registry like all normal clients. Basically Portus is going to
authenticate against itself. :)

The client library used to interact with the registry is already merged
in master (see `app/models/registry_client.rb`). However we still have
to provide a username and password. Hence we need to send the password
as a client, and then verify itself inside of the token controller.

We are going to authenticate as 'portus' user providing a time based one
time only password (aka TOTP).
The TOTP is generated using a secret that changes every time the Rails process
starts (see `config/application.rb`).

The policy checks inside of the token controller have been disabled for the
'portus' user because he can access to everything.s

I've also reserved 'portus' as username to avoid misuses from the users.

### DIFF
--- a/spec/api/v2/token_spec.rb
+++ b/spec/api/v2/token_spec.rb
@@ -24,20 +24,54 @@ describe '/v2/token' do
     context 'as invalid user' do
 
       it 'denies access when the password is wrong' do
-        get v2_token_url, { service: 'test', account: 'account', scope: 'repository:foo/me:push' }, invalid_auth_header
+        get v2_token_url, { service: registry.hostname, account: 'account', scope: 'repository:foo/me:push' }, invalid_auth_header
         expect(response.status).to eq 401
       end
 
       it 'denies access when the user does not exist' do
-        get v2_token_url, { service: 'test', account: 'account', scope: 'repository:foo/me:push' }, nonexistent_auth_header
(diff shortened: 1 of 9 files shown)


flavio:
@mssola as usual feedback is welcome :)

flavio:
@jordimassaguerpla : please monitor this PR, once it gets merged there is going to be another gem to package,

spec/api/v2/token_spec.rb
```diff
@@ -24,22 +24,56 @@
...
 
       it 'denies access when basic auth credentials are not defined' do
-        get v2_token_url, { service: 'test', account: 'account', scope: 'repository:foo/me:push' }
+        get v2_token_url, { service: registry.hostname, account: 'account', scope: 'repository:foo/me:push' }
         expect(response.status).to eq 401
       end
 
     end
 
+    context 'as the special portus user', bug: true do
```
mssola: I guess that `bug: true` is a tag for RSpec. Please clean it ;)

mssola:
Besides that small note, the code looks good to me :+1:

flavio:
Fixed.

PR #69, opened 2015-05-18
Convert from less to sass

flavio:
This makes possible to drop the `therubyracer` gem, which makes packaging a lot easier.


flavio:
According to my tests everything looks good.

@jordimassaguerpla ping

jordimassaguerpla:
lgtm

Issue #54, opened 2015-05-12
Special handling of personal namespaces


flavio:
Whenever a new user is create a personal namespace is being created. That involves the creation of a team named like the user, which in turn creates a namespace named as the team itself (and hence as the user).

These personal namespaces should be treated in a special way:
  * the associated team should be hidden
  * the page of the associated team should not be accessible
  * no users can be added/removed from the team
  * it is possible only to toggle their public status

flavio:
This has been implemented with #70 

I think we can just ignore these points:
  * the page of the associated team should not be accessible
  * no users can be added/removed from the team

These operations are not possible because the required UI with the links/buttons is not shown. However the owner of the team or an administrator can still perform these operations by making curl requests. I think we can ignore that.

@eotchi feel free to open the issue again if you are not comfortable with that.

PR #70, opened 2015-05-18
Global namespaces support, plus better handling of teams

flavio:
This PR introduces the concept of global namespace.

## Global namespaces

A global namespace is identified by a boolean flag and it's associated with a Registry instance. Global namespaces are **always** public, meaning everybody can pull from them. Only admin users can push to them. As usual there's a team behind the global namespace, so an admin user could add contributors to it to have more people with pull rights.

## Hidden teams

The team model now has a new boolean attribute: hidden. This is used to hide certain teams like: the one associated to a global namespace and the one associated with a personal user namespace.

### Namespaces index

The namespaces page now has to sections: special namespaces and normal ones. The special section contains the global namespaces plus the personal namespace of the user (this addresses some of the points of issue #54).

namespaces index

### Teams index

The team page hides all the teams that have `hidden` set to `true`. This fixes some of the points of issue #54.

team index

In the above screenshot the user is part of 0 teams as opposed to before when the team of the personal namespace would have been shown.

### DIFF
--- a/app/controllers/admin/namespaces_controller.rb
+++ b/app/controllers/admin/namespaces_controller.rb
@@ -1,6 +1,6 @@
 class Admin::NamespacesController < Admin::BaseController
   def index
-    @namespaces = Namespace.all
-    render template: 'namespaces/index'
+    @special_namespaces = Namespace.where(global: true)
+    @namespaces = Namespace.where(global: false)
   end
 end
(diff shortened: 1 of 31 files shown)


flavio:
@mssola as usual feedback is welcome :)

mssola:
I get an exception when accessing the admin dashboard and clicking "Namespaces". In particular, the exception happens in the file `app/views/namespaces/index.html.slim` in the line 20, because the `@special_namespaces` instance variable is nil.

mssola:
Everything else LGTM :+1:

flavio:
@mssola thanks for the testing and the notes. I fixed the issue you reported by creating admin-only pages for the namespaces and for the teams.

I'm going to merge the code, but feel free to comment if you do not like something

PR #75, opened 2015-05-21
Initial version of the search feature

flavio:
Added a simple search feature that allows to find repositories.

How it works:
  * user enters a string inside of the search box
  * a fulltext search is performed against both the repositories.name and the namespaces.name columns using the OR operator
  * results are shown inside of the new 'search' page

Mandatory s
search
creenshot

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: ruby
 rvm:
   - 2.1.3
+before_install:
+  - |
+    bash -c " # Install MariaDB
+    sudo bash .travis_mariadb.sh
+    "
 before_script:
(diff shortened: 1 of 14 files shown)


flavio:
@mssola feedback is welcome as usual

flavio:
Ok, just ignore the build errors right now. The tests are failing because the mysql db provided by travis does not support fulltext search indexes on the default engine. I'll change the travis code to use latest stable version of MariaDB instead of some old MySQL version.

mssola:
LGTM :+1: Great Work! :)

flavio:
Thanks

Issue #74, opened 2015-05-21
Team name should be unique


flavio:
This is not enforced right now.

flavio:
Fixed

Issue #68, opened 2015-05-17
Ensure users can be created without conflicts 


flavio:
When creating a new user we must ensure their username is unique (already done) but also that we can create a namespace with the same name. If not then we should block user creation.

flavio:
Fixed

PR #77, opened 2015-05-22
Fix db constraints

flavio:
Ensure:
  * a user cannot be created when there's already a namespace with the
    same name
  * namespace names must be unique relatively to the registry they
    belong to
  * repositories names must be unique relatively to the namespace they
    belong to
  * tags names must be unique relatively to the repository they belong
    to
  * team name must be unique

We are now verifying the format of the name used by repositories and
tags. As usual we follow docker's directives.

This fixes issue #74 and issue #68

### DIFF
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -8,6 +8,7 @@ class Namespace < ActiveRecord::Base
   belongs_to :team
   validates :name,
             presence: true,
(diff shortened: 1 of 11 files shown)


flavio:
@mssola could you please double check these changes? I don't want to miss something

mssola:
LGTM :+1:

Issue #73, opened 2015-05-20
RFC: do not create a namespace upon team creation


flavio:
Right now when a new team is created an associated namespace is created as well.

Given:
  * we do not have a way to delete teams yet
  * this consumes the possible usernames to pick (because the team's name is used to created the associated namespaces, the same thing happens with users)

I would propose to stop doing that. That would also close issue #72

flavio:
@eotchi what do you think about it?

eotchi:
As discussed elsewhere, I am completely fine with that

Issue #53, opened 2015-05-12
When creating a new team the name of the associated namespace should be shown


flavio:
We should show the name of the namespace associated with the team during its creation. We should also allow the user to change the proposed name.

lsamayoa:
What is the status of this after #78 was merged ?

flavio:
Yeah, this is no longer valid. Closing. Thanks!

PR #78, opened 2015-05-25
Don't allocate automatically a namespace upon team creation

### LINKED ISSUE #72: Look into issue triggered by creation of a team with a uppercase name
I suspect something goes wrong when creating the associated namespace.

flavio:
When creating a new team no namespace will be automatically created.
This implments what discussed inside of #73. It also fixes #72 and
makes #53 useless.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -20,10 +20,20 @@ class User < ActiveRecord::Base
     end
   end
 
-  def create_personal_team!
-    if Team.find_by(name: username).nil?
-      Team.create!(name: username, owners: [self], hidden: true)
+  def create_personal_namespace!
+    # the registry is not configured yet, we cannot create the namespace
+    return unless Registry.any?
+
+    team = Team.find_by(name: username)
+    if team.nil?
+      team = Team.create!(name: username, owners: [self], hidden: true)
     end
+
+    Namespace.find_or_create_by!(
+      team: team,
+      name: username,
+      registry: Registry.last # TODO: fix once we handle more registries
+    )
   end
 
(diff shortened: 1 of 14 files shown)


flavio:
@mssola as usual, feedback is welcome

app/models/user.rb
```diff
@@ -20,10 +20,22 @@ def private_namespace_available
...
   end
 
-  def create_personal_team!
-    if Team.find_by(name: username).nil?
-      Team.create!(name: username, owners: [self], hidden: true)
+  def create_personal_namespace!
+    if Registry.count == 0
+      # the registry is not configured yet, we cannot create the namespace
+      return
     end
```
mssola:
```ruby
if Registry.count == 0
  return
end
```
should be:
```ruby
return unless Registry.any?
```

app/observers/registry_observer.rb
```diff
@@ -2,6 +2,14 @@ class RegistryObserver < ActiveRecord::Observer
...
   def after_create(registry)
     registry.create_global_namespace!
+
+    # Now it is possible to create the private namespaces
+    # of all the users that signed into portus _before_
+    # the 1st registry was created
+    # TODO: change code once we support multiple registries
+    User.all.each do |user|
+      user.create_personal_namespace!
+    end
```
mssola: May be you could put it into a single line (through { }) while we don't support multiple registries.

mssola:
LGTM :+1:

flavio:
I fixed the code the way you suggested. Thanks for the feedback.

PR #85, opened 2015-06-04
add integration test data

jordimassaguerpla:
This data is for testing the appliance

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,0 +1,19 @@
+if ENV['INTEGRATION_TESTS']
+  puts 'Adding user username31'
+  if (User.find_by_username('username31'))
+    puts 'User already exists. Please drop the database first'
+    exit -1
+  end
+  User.create(
+    username: 'username31',
+    email: 'a1@b.com',
+    password: 'test-password',
+    admin: true
+  )
+
+  puts 'Adding registry portus.suse.example.com:5000'
+  Registry.create(
+    name: 'portus.suse.example.com:5000',
+    hostname: 'portus.suse.example.com:5000'
+  )
+end


flavio:
The right location to add this sort of stuff is inside of `db/seeds.rb`. Please move the code inside of this file and wrap it by something like:

```ruby
if ENV['INTEGRATION_TESTS']
# code
end
```

Then you just have to do something like:
```
RAILS_ENV=production INTEGRATION_TESTS=1 rake db:seed
```

db/seeds.rb
```diff
@@ -0,0 +1,23 @@
...
+    exit -1
+  end
+  u = User.new
+  u.username = "username31"
+  u.email = "a1@b.com"
+  u.password = "test-password"
+  u.admin = true
+  u.save
+
+  puts "Adding namespace for user username31"
```
flavio: This is not needed, the user namespace is automatically created as soon as the Registry is set.

db/seeds.rb
```diff
@@ -0,0 +1,23 @@
...
+  namespace = Namespace.new
+  namespace.name = "username31"
+
+  puts "Adding registry porter.suse.example.com:5000"
+  r = Registry.new
```
flavio:
```ruby
Registry.create(name: 'foo', hostname: 'bar')
```

db/seeds.rb
```diff
@@ -0,0 +1,23 @@
+if ENV['INTEGRATION_TESTS']
+  puts "Adding user username31"
+  if (User.find_by_username("username31") != nil)
+    puts "User already exists. Please drop the database first"
+    exit -1
+  end
+  u = User.new
```
flavio:
```ruby
User.create(username: 'foo', email: '...',...)\
```

db/seeds.rb
```diff
@@ -0,0 +1,23 @@
+if ENV['INTEGRATION_TESTS']
+  puts "Adding user username31"
+  if (User.find_by_username("username31") != nil)
```
flavio: You can drop the `nil` check.

flavio:
LGTM

flavio:
:+1:

PR #93, opened 2015-06-09
add landing page for appliance

jordimassaguerpla:
appliance is being built under
obs://Virtualization:containers:Portus/PortusAppliance

The appliance contains a landing page.

I don't know where to put this so I am creating a directory for it and
later on we can move it to whatever location we think is best.

### DIFF
--- a/packaging/appliance/index.html.in
+++ b/packaging/appliance/index.html.in
@@ -1,0 +1,54 @@
+<html>
+<head>
+</head>
+<body>
+
+  <h1>Welcome to the Portus Appliance!</h1>
+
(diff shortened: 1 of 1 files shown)


flavio:
It looks good, I would fix the indentation a bit if possible. But I'm fine with merging it right now

jordimassaguerpla:
thanks

mssola:
It looks good to me. Indentation would be nice, but not too urgent since the designer will most likely redo the html anyways.

As for the directory, it's fine to leave it there for now, but most probably it will end up being at `public` :)

PR #99, opened 2015-06-11
Front end fixes

cyntss:
Mayor fixes in the UI:

image

image

image


mssola:
Nice! Some issues though:

1. The search item does not autofocus for me.
2. Tests are failing. This happened because you've updated the text of the sign up button and the sign up link. This is fine, it only means that the test suite has to be updated.

From my side, just fix the autofocus thing and once it gets merged, I can update the tests myself.

I'd say that you should also wait to hear the opinion of both @flavio and @eotchi :)

flavio:
I :heart: it!

cyntss:
thank you @flavio  :)

@mssola excellent! I will check why it doenst autofocus for you and fix it now :+1:

mssola:
Ok, not the autofocus works for me. Thanks !

Issue #81, opened 2015-06-02
Running in docker


zachaller:
Just wondering if running this in docker will ever be a thing or something people would be interested in.

flavio:
I think it would be nice to have a docker image for portus. Right now it's not on our roadmap.

arabold:
I've started to play with a `Dockerfile` for Portus. It starts up now but still needs some work:
https://github.com/arabold/Portus/tree/feature/add-dockerfile

At this time I'm only playing with it to see what I can do with Portus. But feel free to take a look and do some improvements to the Dockerfile.

kikicarbonell:
:+1:

flavio:
Done

PR #116, opened 2015-06-11
This is a first version of a working Docker container with Portus

arabold:
I leave this here for comments and in case somebody else wants to take a look. I wouldn't consider this pull request to be ready for merge yet! However, the Docker container should work right out of the box and give you a _working_ Portus setup which you can connect to an external registry. It's noteworthy that the database is bundled directly into this image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+FROM ruby:2.1
+
+RUN echo mysql-server-10.0 mysql-server/root_password password 'PASS' | debconf-set-selections
+RUN echo mysql-server-10.0 mysql-server/root_password_again password 'PASS' | debconf-set-selections
+RUN DEBIAN_FRONTEND=noninteractive \
+  apt-get update && apt-get install -y \
+  nodejs \
+  mariadb-server \
+  apache2 \
+  apache2-dev
+
+COPY . /portus
+COPY docker-entrypoint.sh /
+
+WORKDIR /portus
+RUN bundle config build.nokogiri --use-system-libraries
+RUN bundle install
+
+RUN gem install passenger -v 5.0.7
+RUN passenger-install-apache2-module -a
+
+COPY docker/config/sysconfig_apache2    /etc/sysconfig/apache2
+COPY docker/config/httpd.conf.local     /etc/apache2/mods-enabled/passenger.conf
+COPY docker/config/portus.test.lan.conf /etc/apache2/conf-enabled/portus.conf
+
+ENTRYPOINT ["/docker-entrypoint.sh"]
+CMD ["/usr/sbin/apache2ctl", "-D", "FOREGROUND"]
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Start DB server and reset password to empty
+service mysql start >/dev/null
+mysql -uroot -pPASS -e "SET PASSWORD = PASSWORD('');"
+
+# Setup Portus database
+bundle exec rake db:create
+bundle exec rake db:migrate
+
+chown -R www-data:www-data /portus
+
+exec "$@"
--- a/docker/config/httpd.conf.local
+++ b/docker/config/httpd.conf.local
@@ -1,0 +1,5 @@
+LoadModule passenger_module /usr/local/bundle/gems/passenger-5.0.7/buildout/apache2/mod_passenger.so
+<IfModule mod_passenger.c>
+  PassengerRoot /usr/local/bundle/gems/passenger-5.0.7
+  PassengerDefaultRuby /usr/local/bin/ruby
+</IfModule>
--- a/docker/config/portus.test.lan.conf
+++ b/docker/config/portus.test.lan.conf
@@ -1,0 +1,14 @@
+<VirtualHost *:80>
+   ServerName portus.test.lan
+   # !!! Be sure to point DocumentRoot to 'public'!
+   DocumentRoot /portus/public    
+   <Directory /portus/public>
+      # This relaxes Apache security settings.
+      AllowOverride all
+      # MultiViews must be turned off.
+      Options -MultiViews
+      # Uncomment this if you're on Apache >= 2.4:
+      Require all granted
+      PassengerAppEnv development
+   </Directory>
+</VirtualHost>
--- a/docker/config/sysconfig_apache2
+++ b/docker/config/sysconfig_apache2
@@ -1,0 +1,280 @@
+## Path:	Network/WWW/Apache2
+## Description:	Configuration for Apache 2
+
+## Type:	string
+## Default:	""
+## ServiceRestart: apache2
+#
+# Here you can name files, separated by spaces, that should be Include'd from 
+# httpd.conf. 
+#
+# This allows you to add e.g. VirtualHost statements without touching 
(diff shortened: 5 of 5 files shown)


flavio:
Thanks for your contribution, this is a good starting point to address issue #81

adamlc:
Is this going to get merged? It seems crazy that a product developed for Docker doesn't have its own Dockerfile to run a container at the moment!

flavio:
@adamlc as you can see from the initial comment this Dockerfile is far from ready to be merged.

We are busy with other tasks, but we plan to tackle this issue as soon as possible.

Obviously PRs are welcome ;)

rtrauntvein:
This is a good starting point.

There are a few things that I would like to see, first of which is splitting out the web server and mysql services into their own containers.

DirkHeinke:
Is there a reason why you want to split out the services? In my opinion there is no reason not to have it all in one container, but it would be much easier to deploy and use if you have it all in one container.

For my tests I used basically the Dockerfile of this PR. I planned to move all the Files related to docker into the docker folder and add a build and a runscript.
That would be my approach and I can create another PR if this is the way we can agree on.

rtrauntvein:
@DirkHeinke My opinion on multiple containers is mostly about thinking how this could be deployed into a production environment with Docker.  I agree that what we have here is probably sufficient for development environment. It is great to get contributers and PR's going quickly.

The reason I would split out things like MySQL, is mainly in order to stick with best practices in a micro-services architecture, and to have one purpose per container. http://microservices.io/patterns/deployment/service-per-container.html

It also allows  more flexibility in the database technology. For example, in one deployment scenario, we may wish to use something else, such as Amazon RDS rather than what is provided. We can also independently scale the web front ends by having the services in separate containers.

Orchestrating this is rather simple with a [docker-compose.yml](https://docs.docker.com/compose/). Here is an example:

```yaml
portus:
  build: .
  links:
    - mysql:mysql
  ports:
    - 80:80
mysql:
  image: mysql:latest
```

Ideally we would also set the MySQL Host, DB, User, and Password for the Portus container via environment variables at runtime. I didn't get far enough into Portus yet to see where this is configured exactly.

DirkHeinke:
It's always a questions of the point of view. The single purpose could also be "provide a docker index". Even Docker wrote about that: https://docs.docker.com/articles/using_supervisord/
And just to do something, because some patterns say so...

I see your points in scaling and exchangeability, but to be honest, I don't see a scenario where you have to scale the frontend for a private index? Not at the current state of this project.

I totally agree on your point with the environment variables and we should also keep in mind, that there are development secrets in the config/secrets.yml. If we build a container, we should point that out more in the Documentation (for the container). 

Perhaps also add a volume for the certificates?

rtrauntvein:
I wouldn't recommend to use microservices just because a pattern says so, I recommend it because I have seen the pattern be proven to work.  If it is just a monolithic image, say I want to use something other than the provided mysql (perhaps RDS, or a Galera cluster), then I need to fork the project, re-factor the Dockerfile and maybe other parts to remove that MySQL instance.

I actually forgot to mention supervisord in my earlier reply, it does work well for multiple services in a container. This would be the way to go to launch multiple services.

Many other docker image projects out there do provide an included MySQL and even other components managed by supervisord.  The best ones have logic to detect when the needed services are linked as separate containers, and then do not setup and start the included services.  

The (docker-gitlab)[https://github.com/sameersbn/docker-gitlab] project does this internal / external service dependency handling quite well. This gives the best of both worlds.  Simple to get up and running for dev, and the resiliency and portability that operations requires.  So maybe that is the answer?

re: certificates. - A volume for them works well.  Can then be built into a base image or added via configuration management.

arabold:
I consider the MySQL setup as something internal that usually wouldn't need to be exposed. Portus could even use sqlite (and maybe should?) to achieve the same - it's just the simplest way to store and query settings and data. In fact it's just an implementation detail in my eyes. There is no concurrency or load balancing that would require a central, separated MySQL server - that looks like overkill to me.

Long story short, I'd continue to include it into the docker image instead of linking a separate container. It adds unnecessary complexity at this stage of development.

rtrauntvein:
I do agree with you, at this stage in development it is not needed. I wouldn't want to hold up this PR as it is great to people get up and running for development. So +1 in that respect.

I do though want to continue looking towards what it would take to eventually run this as a production service depended upon by large teams of developers.  To our team that means if all of the configuration settings in data are stored in MySQL... We would prefer to have the ability to use something that could be replicated and interchanged as needed. 

sqlite also has it's issues in docker, like a way to facilitate configuration settings import/export between container removal. It is simple but we'll likely end up just mounting it in the sqlite database file via bind mount, which is not really ideal for large scale production setup either. 

We can continue this discussion at a later point and see where the PR's take us ;)

adamlc:
Sorry if I sounded hostile earlier, wasn't my intention :smiley: 

The Dockerfile could use a bit of work but its pretty difficult for people to contribute until its merged or at the very least in its own branch. At the moment this is a good starting point :+1: 

As regards the multiple container argument I personally think its fine to keep everything in the same container. As long as a volume is exposed for persistence and the init scripts are smart enough to figure out existing volumes and migrations etc

arabold:
Before merging the pull request should be cleaned up a bit and probably combine the vagrant and the docker configuration files in a single location - there's too much duplication for my taste. And it needs some documentation about the key files, etc. 

Unfortunately I won't have time to work on this in the next few weeks as we're using the official docker hub's private repos instead of self hosting for now. I wanted to have this docker file mainly to be able to test Portus easily (and I'm not a fan of vagrant).

rtrauntvein:
I like the idea of combining the docker / vagrant conf files.  I tried taking a stab at that today but there is some places where /vagrant/ us being used in the conf files like for the web root.

Do we want to mimic the vagrant behavior of mounting the project working directory into the Docker container? Therefore allowing "live" editing.

adamlc:
Well I think the docker container will be used 2 lots of people. 1 for development purposes and 2 for end users to test / run the product. 

I think it would make the most sense to copy the project into the actual container. In that case the end users can test it. If people want to develop using the container than they can just mount the working directory as a volume over the code in the container if that makes sense?

dmacvicar:
A Docker container is not enough for an Vagrantfile that launches 3 services.

What about using Docker + Vagrantfile like I did here? https://github.com/dmacvicar/garita/blob/master/Vagrantfile

Otherwise we would need to add something like docker-compose to the mix.

flavio:
I'm fine having a Docker file for test/production usage. However I would prefer to have something taking advantage of what is already on the Docker Hub and also following Docker's best practices.

Hence I would propose to create two images:
  * portus: based on https://registry.hub.docker.com/_/rails/
  * db: based on https://registry.hub.docker.com/_/mariadb/

They should be orchestrated using docker-compose.

Then we, for development purposes, can start using them either via Vagrant or just by using docker-compose. We can do a smooth transition from the current development model to a new one.

I would say, let's just focus on the end-user use case for now.

flavio:
Initial docker-compose implementation here: https://github.com/SUSE/Portus/pull/212

flavio:
Thanks for your contributions, both in terms of code and ideas. I'm closing this PR now that PR https://github.com/SUSE/Portus/issues/212 has been merged in master.

adamlc:
Fantastic! Great work :+1:

flavio:
One final note, issue https://github.com/SUSE/Portus/issues/217 might be interesting for you guys.

PR #136, opened 2015-06-16
add stylesheet information to landing page

jordimassaguerpla:
This works without having to modify the apache2 configuration. @flavio : good idea :) !


mssola:
The link won't work. Try `/assets/aplication.css`.

flavio:
Also what happens when the assets are changed and we rebuild the appliance? The link contains the hashed value of the assets inside of its name. This should be determined during the 1st boot

jordimassaguerpla:
also in the rpm since an update would retrigger the precompilation of the assets

jordimassaguerpla:
https://github.com/SUSE/Portus/blob/master/packaging/suse/Portus.spec.in#L85
I could use some "ls ..." + "sed " in the landing.html.in

flavio:
Yes, or simply something involving "find"

mssola:
LGTM :+1:

flavio:
:+1:

PR #209, opened 2015-07-08
Adapt the temporary layout to Portus' interface

eotchi:
Just a couple of changes. 
It will give the following results:

adapted_layout


flavio:
I think it's fine, we should add some spacing between the logo and the 3 buttons (View on github, download as zip,...)

mssola:
+1 with what @flavio says. Plus, you've removed the tagline "Claim control of your Docker images!". I liked that :P

eotchi:
Something like this?
with_space

flavio:
Yes, that's definitely better. Can you add back the "Claim control of your Docker images!" text?

stylesheets/stylesheet.css
```diff
@@ -111,8 +111,8 @@ a {
 .main-content h1, .main-content h2, .main-content h3, .main-content h4, .main-content h5, .main-content h6 {
   margin-top: 2rem;
   margin-bottom: 1rem;
-  font-weight: normal;
-  color: #159957; }
```
mssola: why? I actually like this.

eotchi:
Yep. At this point, I'd suggest this:
font

flavio:
:+1:

mssola:
:+1:

stylesheets/stylesheet.css
eotchi: That's because there is no green in Portus. 
#383e41 is a "gimp colour pick" of the darker blue used in the logo

PR #212, opened 2015-07-08
Run Portus using docker-compose

flavio:
This commit adds the possibility to run Portus and its database using docker-compose. This is a different take for issue #81.

I'll add the documentation tomorrow. In the meantime feel free to comment on it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git*
+tmp
+vagrant
+packaging
+log/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM rails:4.2.2
+MAINTAINER Flavio Castelli <fcastelli@suse.com>
+
+RUN mkdir /portus
+WORKDIR /portus
+ADD . /portus
+RUN bundle install
+RUN mv /portus/config/database-docker-compose.yml /portus/config/database.yml
+
+EXPOSE 3000
--- a/README.md
+++ b/README.md
@@ -30,6 +30,55 @@ If Portus is unreachable when a new image is being pushed to the Docker registry
 
 ## Development environment
 
+### Docker
+
+It is possible to use docker-compose to spin up a small development/demo
+environment.
+
+The environment consists of two Docker containers linked together:
+
+  * web: this is the container running Portus. It's based on the
+    [official rails](https://registry.hub.docker.com/_/rails/) Docker image.
+
+  * db: this is the container running the database required by Portus. It's
+    based on the
+    [official mariadb](https://registry.hub.docker.com/_/mariadb/) Docker image.
+
+First of all ensure you have [docker-compose](https://www.docker.com/docker-compose)
+installed. Then do:
+
+```
+docker-compose up
+```
+
+This will:
+  * download the `rails` Docker image from the Docker Hub
+  * build the `web` Docker image
+  * download the `mariadb` docker image from the Docker Hub
+  * start the `db` container and link it against a running insitance of the `web` container.
+
+The next step is to create Portus' database:
+
+```
+docker-compose run web rake db:create
+```
+
+Then populate the database by running Rails' migrations (**note well:** this
+might need to be done from time to time, when new migrations are added):
+
+```
+docker-compose run rake db:migrate
+```
+
+Now everything is ready to be used. Portus' UI will be accessible on
+[http://localhost:3000](http://localhost:3000).
+
+#### Demo
+
+[![asciicast](https://asciinema.org/a/23185.png)](https://asciinema.org/a/23185)
+
+### Vagrant
+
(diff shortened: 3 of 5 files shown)


rtrauntvein:
Cool!

I had to add `MYSQL_DATABASE: portus_development` to the environment variables for the web container in the `docker-compose.yml` - then ran `docker exec -it portus_web_1 rake db:migrate RAILS_ENV=development` which setup the DB.

jordimassaguerpla:
It is simpler than #116 and I like that!

I see this is for development, right? Otherwise we should think about SSL and secrets configuration and apache+passenger.

DB configuration:  What's the plan there? Should the user add the portus user and the database to the mariadb container manually?

Add "RUN rake db:migrate"

I think you are missing installing docker-distribution ,right?

may it be this line?

RUN apt-get update -qq && apt-get install -y

and then I would also add a 

RUN apt-get update && apt-get upgrade

so that there are the latest security fixes

but the idea looks great!

rtrauntvein:
I would say this is pretty sufficient for getting this off the ground for development. Having the ability to specify DB host, user, port, and name  secrets via `ENV` variables on the Portus container would be a nice next step towards making it closer to production ready.  

SSL maybe could be accomplished by adding haproxy, or nginx reverse proxy in front. Otherwise just documenting how to mount in certs (via `docker -v`)  and providing a config for SSL is fine. Maybe even generate a pair of self signed certs that are used if custom certs are not specified.

By adding the `MYSQL_DATABASE` ENV variable to the mariadb section of `docker-compose.yml` it bring the DB setup one step closer. 

The `rake db:migrate` would likely need to be added to an init script for the portus container image if you want to setup the rest automatically. It wouldn't work in a `RUN` during build of the portus image, as the DB container would not be linked at that time.

Per https://docs.docker.com/articles/dockerfile_best-practices/ doing `apt-get upgrade` in a Dockerfile is not reccomended.

jordimassaguerpla:
regarding migrations, we could reuse this

https://github.com/SUSE/Portus/blob/master/config/environments/production.rb#L80

we should make sure thought database is configured when we start rails....

I wasnt' aware of running apt-get upgrade is not recommended but now I understand why so many images in docker hub have insecure packages...

How would you set the ENV variables in production?

flavio:
> I see this is for development, right? Otherwise we should think about SSL and secrets
> configuration and apache+passenger.

Yes this is for development purposes right now. We can iterate later over it to make it ready for production.

> DB configuration: What's the plan there? Should the user add the portus user and the database to > the mariadb container manually?

The mariadb automatically creates a root user with the password mentioned inside of the `docker-compose.yml` file. Then the db is setup up manually by the user. This is part of the documentation I'm writing.

> Add "RUN rake db:migrate"

We cannot do that because at build time the ports container is not linked against the db one. This has to be run manually. This is part of the doc I'm writing.

> I think you are missing installing docker-distribution ,right?

No, that's on purpose. We can take advantage of the official registry container for that. I have to figure out how to add it to the mix.

> may it be this line?
> RUN apt-get update -qq && apt-get install -y
> and then I would also add a
> RUN apt-get update && apt-get upgrade
> so that there are the latest security fixes

These are leftovers , I'm going to drop them as proposed inside of another comment.

flavio:
> I would say this is pretty sufficient for getting this off the ground for development. Having the ability > to specify DB host, user, port, and name secrets via ENV variables on the Portus container would > be a nice next step towards making it closer to production ready.

Yep, we can focus on that with a dedicated iteration.

> By adding the MYSQL_DATABASE ENV variable to the mariadb section of docker-compose.yml it > bring the DB setup one step closer.

What would be the purpose of that?

> Per https://docs.docker.com/articles/dockerfile_best-practices/ doing apt-get upgrade in a Dockerfile is not reccomended.

Point taken, I'm going to drop them

flavio:
> I then ran docker exec -it portus_web_1 rake db:migrate RAILS_ENV=development which setup the DB.

You aren't supposed to do that. You should use:
```
docker-compose run web db:migrate
```

This is documented inside of the doc I'm currently writing.

flavio:
Added documentation and asciicast

[asciicast](https://asciinema.org/a/23185)

rtrauntvein:
I missed that there was a `db:create` available. The `MYSQL_DATABASE` variable was just how I pre-setup the Portus db in the mariadb container. 

`docker-compose run web db:migrate` makes sense, exec was just how I did it before your docs were available.

So all is good here ;)

kevinsimper:
Do you imagine that it would be possible also to start the `registry:2.0` in docker-compose as well?

mssola:
LGTM

flavio:
> Do you imagine that it would be possible also to start the registry:2.0 in docker-compose as well?

Yes, I plan to do that with a separate pull request

PR #218, opened 2015-07-10
Changed the README after some of its contents have been moved to the wiki

mssola:


flavio:
The build failures are totally unrelated with this change. @mssola please take care of them

PR #228, opened 2015-07-22
Fix duplicated namespaces on the GET /namespaces

flavio:
Namespaces that are both accessible to the user and public used to be repeated twice.


mssola:
LGTM, just wait for Travis

PR #232, opened 2015-07-23
Add pagination

### LINKED ISSUE #135: Support pagination
Right now Portus does not support pagination. We should implement it in several places like: namespaces, repositories, users,...

lsamayoa:
Fixes #135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Upcoming Version
+
+- Teams page is now paginated
+- Teams namespaces and users are now paginated in team detail page
+
 ## 1.0.1
 
 - Fixed regression where namespaces could not be created from team page
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## Upcoming Version
 
-- Teams page is now paginated
-- Teams namespaces and users are now paginated in team detail page
+- Added pagination feature
 
 ## 1.0.1
 
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,9 @@ gem 'rotp'
 gem 'mysql2'
 gem 'search_cop'
 
+# Pagination
+gem 'kaminari'
+
 # In order to create the Gemfile.lock required for packaging
 # meaning that it should contain only the production packages
 # run:
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -127,6 +127,9 @@ GEM
     json-schema (2.5.1)
       addressable (~> 2.3.7)
     jwt (1.5.0)
+    kaminari (0.16.3)
+      actionpack (>= 3.0.0)
+      activesupport (>= 3.0.0)
     launchy (2.4.3)
       addressable (~> 2.3)
     loofah (2.0.2)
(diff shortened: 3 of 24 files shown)


lsamayoa:
Fixes #232

mssola:
It does not work for me :/ The `paginate` helper method does  not generate anything in the resulting HTML for me. Am I missing something ?

lsamayoa:
maybe `bundle install`

lsamayoa:
I see. It is not generating because there is not enough data.
It does not display anything if pagination is not needed.

flavio:
@mssola I'm going to look into that this morning

flavio:
I can see the changes working. This is a summary of how pagination interacts with Portus.

## dashboard

### INDEX

The list of activities is shown there.

No pagination is done here, however we already limit the number of items to 20. So I think nothing should be done.

## Namespaces

### INDEX

We have two tables here:

  * special namespaces: default pagination -> 25 per page
  * "normal" namespaces: default pagination -> 25 per page

#### Proposed change
I think it's useless to pagine the special namespaces, I propose to not do that. There are going to be only 2 special namespaces as long as we do not make it possible to handle more than 1 registry.

Note well: the pagination for the special repositories is also broken, but we do not care about that if we remove the feature.

### SHOW
  
There's a single table showing all the repositories that are part of a namespace. They are paginated using the default value -> 25 per page.

#### Proposed change

Nothing, everything looks good.

## Teams

### INDEX

There's a single table showing all the teams a user is part of. The default pagination is being used -> 25 per page

#### Proposed change

Nothing, everything looks good.

### SHOW

The page has two tables:

  * users: paginated by 5
  * namespaces: paginated by 5

#### Proposed changes

Change paginations values that way:

  * users: 10 per page
  * namespaces: 15 per page

## admin/dashboard

### INDEX

The list of activities is shown there.

No pagination is done here, however we already limit the number of items to 20. So I think nothing should be done.

## admin/activities

### INDEX

There's the list of all the activities happening on the system

They are paginated using the default value -> 25 per page.

#### Proposed changes

Nothing, everything looks good.

## admin/registries

### INDEX

All the available registries are shown. No pagination is done.

#### Proposed changes

Nothing, we just support one registry right now.

## admin/users

### INDEX

The list of users is shown. We use the default value for pagination -> 25 per page

#### Proposed changes

Nothing, everything is fine.

## admin/teams

### INDEX

All the teams on portus. Default pagination value is used -> 25 per page

#### Proposed changes

Nothing, everything is fine

## admin/namespaces

### INDEX

Two tables are shown:

  * special namespaces: only global namespaces are shown. Default pagination -> 25 per page
  * normal namespaces: all the other ones, including the personal ones of the users. Default pagination -> 25 per page

#### Proposed changes

Drop pagination for the special namespaces, we just support 1 global namespace ATM.

flavio:
I have also some changes locally that switch the pagination theme from the default one:

screenshot-localhost 5000 2015-07-27 10-14-45

To one based on bootstrap3:

screenshot-localhost 5000 2015-07-27 11-06-15-bootstrap

mssola:
I agree to the proposed changes

flavio:
Ok, I'll merge the PR and fix the issues straight on master

flavio:
@lsamayoa thanks a lot for your contribution!

PR #235, opened 2015-07-27
ISSUE-234: namespaces name validation now also adds an error to the model now to fix #234

### LINKED ISSUE #234: Suspected validation error when creating a new team
How to reproduce:
  * Create a new team named "QA"
  * The team is listed inside of the page
  * Reload the page: the team is gone

Expected result: the team is not deleted.

lsamayoa:
Fixes #234

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -27,6 +27,8 @@ class Team < ActiveRecord::Base
   private
 
   def downcase?
-    name.downcase == name
+    is_downcase = name.downcase == name
+    errors.add(:name, "namespace names cannot contain uppercase letters") unless is_downcase
+    is_downcase
   end
 end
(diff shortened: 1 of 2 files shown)


flavio:
Hi, I suspected something similar was happening. In the past as soon as a team was created we used to create also a namespace named after it. This is no longer happening, so we can probably remove some validations checks from the team object.

Could you please look into that? Also it would be required to provide a test case reproducing this issue

lsamayoa:
TODO:
- [ ] remove useless validation from teams model
- [ ] Add capybara test case for this regression

flavio:
> Add capybara test case for this regression

I think a simple functional test should be enough.

lsamayoa:
@flavio Done ?

mssola:
LGTM

Issue #200, opened 2015-07-06
Take advantage of the Manifest API


mssola:
As described [here](https://github.com/docker/distribution/blob/master/docs/spec/api.md#pulling-an-image-manifest) there's a manifest associated to an image. We should take advantage of this and fetch as much as we can from it.

Jean-Baptiste-Lasselle:
hi @mssola ... I maybe am not aware of something here : why is this issue still open...?

Issue #199, opened 2015-07-06
Take advantage of the Catalog API


mssola:
Right now Portus' database is filled in using the data sent by distribution via the `webhook API`. Glitches inside of Portus could misalign the contents of its database in regards of what is actually inside of the registry.

The next version of distribution (not yet released) is going to provide a [search API](https://github.com/docker/distribution/issues/206) that we can use to periodically sync our database with the real contents of the registry.

Portus will call the remote API exposed by the registry, parse the response and populate the database accordingly. We already have a client library that can be used to talk with registry. Note that registry's code is already in master.

lsamayoa:
[Catalog API] was already merged

[Catalog API]: https://github.com/docker/distribution/pull/653

mssola:
I just assigned myself to denote that I'm already doing some research on this.

mssola:
I've decided since PR #241 that this issue should also serve as a discussion about how we can implement this feature in the cleanest way.

mssola:
In my opinion, there are two strategies around this:

1. We can use an existing gem that deals with background processes like [Sidekiq](http://sidekiq.org/) that continuously performs the sync if needed (e.g. after 5min, sync if needed). Another similar approach, proposed by @lsamayoa is to use [crono](https://github.com/plashchynski/crono). Crono has the benefits that it does not require Redis to actually work.
2. We can write a standalone program for that. A simple script or application dedicated just for this.

It might seem crazy at first, but I'd go for the second option. There are two main reason why I wouldn't go for the first option:

- I'm really skeptical when it comes to background jobs in Ruby. This is mainly due the limitations of Ruby (MRI at least) on concurrency.
- In my mind what the Rails application has to implement is the web UI and the authorization service. Anything beyond that is out of the scope of the application itself in my opinion. Therefore I would write and independent script/application that deals with this.

That being said, I can see why the crono gem can be appealing, but at first glance I don't like that it touches the database to persist the state of the running job. Moreover, we don't need to worry about using platform-dependent stuff.

lsamayoa:
I don't think creating a standalone script would be a good idea doe, I
don't think we should reinvent the wheel.
I would advice to use ActiveJob to have multiple backends available
(crono/sidekiq/etc). And leave scheduling to consumers, maybe just make a
rake task available to sync the app.

With crono background jobs are proccesed in a diferent proccess, just like
with sidekiq.

Sidenote: crono does not use platform dependent stuff.

[...]

PR #241, opened 2015-07-31
WIP: Implement Catalog/Manifest API synchronization

lsamayoa:
Adding this for early review.

Fixes:
- #199 
- #200

Missing:
- [ ] Uni tests for synchronize methods

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,9 @@ gem 'search_cop'
 # Pagination
 gem 'kaminari'
 
+# Memoization
+gem 'memoist'
+
 # In order to create the Gemfile.lock required for packaging
 # meaning that it should contain only the production packages
 # run:
--- a/app/models/registry.rb
+++ b/app/models/registry.rb
@@ -31,11 +31,15 @@ class Registry < ActiveRecord::Base
 
   def synchronize!
     # 1. Fetch repositories names from Catalog API /v2/_catalog
-    catalog_repositories = client.catalog['repositories']
+    catalog = client.catalog
+    update_from_catalog!(catalog)
+  end
+
+  def update_from_catalog!(catalog)
     # 2. Figure out namespace/repo K/V
-    repository_addresses = catalog_repositories.reject(&:nil?).map do |repository_fullname|
+    repository_addresses = catalog['repositories'].reject(&:nil?).map do |repository_fullname|
       logger.debug "Detected repo for synchronization: #{repository_fullname}"
-      return  { 'repository_name' => repository_fullname } unless repository_fullname.include? '/'
+      next { 'repository_name' => repository_fullname } unless repository_fullname.include? '/'
       EXPLODE_REPO_NAME_REGEXP.match(repository_fullname)
     end
 
@@ -44,10 +48,10 @@ class Registry < ActiveRecord::Base
       .reject { |address| address.nil? || address[:repository_name].nil? }
       .map do |address|
         logger.debug "Loading repo for sycnhronization: #{address[:namespace_name]}/#{address[:repository_name]}"
-        return Repository.find_or_create_by!(name: address['repository_name'], namespace: global_namespace) unless address[:namespace_name].nil?
+        next Repository.find_or_create_by!(name: address['repository_name'], namespace: global_namespace) if address['namespace_name'].nil?
         namespaces
           .find_or_create_by!(name: address['namespace_name'])
-          .find_or_create_by!(name: address['repository_name'])
+          .repositories.find_or_create_by!(name: address['repository_name'])
       end
 
     # 4. Synchronize each repository in catalog
--- a/app/models/registry.rb
+++ b/app/models/registry.rb
@@ -24,12 +24,50 @@ class Registry < ActiveRecord::Base
     Namespace.find_by(registry: self, global: true)
   end
 
+  def client
+    RegistryClient.new(hostname, use_ssl)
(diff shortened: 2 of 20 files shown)


app/models/registry.rb
```diff
@@ -20,12 +22,47 @@ def global_namespace
     Namespace.find_by(registry: self, global: true)
   end
 
+  def client
+    RegistryClient.new(hostname, use_ssl)
+  end
+  memoize :client
```
flavio: We are cautious when it comes to introducing new dependencies: it takes time and resources to keep them up-to-date and secure. 

In this case I would drop this gem and dome something simple like:

```ruby
def client
  @client = RegistryClient.new(hostname, use_ssl) unless @client
  @client   
end
```

db/migrate/20150728224900_add_use_ss_to_registries.rb
```diff
@@ -0,0 +1,5 @@
+class AddUseSsToRegistries < ActiveRecord::Migration
```
flavio: Typo, it should be `AddUseSslToRegistries`

flavio:
I think it would be better to address the issues with two separated PRs to make the review process easier.

I'm not particularly convinced by `Registry#update_from_catalog!`. The method is pretty long and it takes time to understand what is going on. I think it would be better to split it into smaller pieces.

As for the image layers, I was planning to store them using either [closure_tree](https://github.com/mceachen/closure_tree) or [awesome_nested_sets](https://github.com/collectiveidea/awesome_nested_set) (more about the latter [here](http://www.mikeperham.com/2012/12/01/12-gems-of-christmas-12-awesome_nested_set/)). I'm think I was more inclined toward the 1st one, but I looked into them many months ago. However in both cases we will have to store the data in a different way.

Gemfile
```diff
@@ -26,6 +26,9 @@ gem 'search_cop'
 # Pagination
 gem 'kaminari'
 
+# Memoization
+gem 'memoist'
```
mssola: I agree with @flavio, we are careful when requiring new gems. This looks avoidable.

mssola:
I have some problems with this PR. Let's see:

1. As @flavio already pointed out, it tries to do too much. These are two non-trivial features, you shouldn't merge them in the same PR.
2. Minor: you claim to fix both issues, but then in the commit message you always write "ISSUE-199", which makes hard to know which commit fixes things for the issue #200 .

Besides that, I don't see in your code when do you sync with the Catalog API. By that I mean that I can see in your code where you do the synchronization but I'm not sure when is that code actually executed. In my opinion there are two strategies here:

1. We can use a separate process or something like [Sidekiq](http://sidekiq.org/) that continuously performs the sync if needed (e.g. after 5min, sync if needed). You're not doing this in this PR.
2. We can sync on push. That is, when a new tag gets pushed, sync. Something is telling me that you're striving for this strategy, but I don't see where you hook the synchronization mechanism on push either. Anyways, this strategy has the same flaw as in the current webhook mechanism.

In regards to the Catalog API, to be honest, I was already thinking on this and I'm still not decided on how is the best way to implement this feature. Note however that I'd prefer a solution that goes on the lines of my first point, since it's the safest way to guarantee that both the registry and the database will be in sync. Since we are asking you to try to divide this PR, if I were you I'd choose to just implement the Manifesto API, since I'm already researching on Catalog API. This doesn't necessarily mean that I "hate" your PR in regards to the Catalog API, but I'd prefer to keep things simple and allow you to work on the Manifesto API.

In regards to the Manifesto API, in general I like it, and I don't really have many thoughts put in this. I'd appreciate if you could consider @flavio's ideas ;)

Another issue that maybe is out of scope of these two issues but that I want to point out anyways: where and how should we show all the results that we gather from the Manifest API? I think that there should be a new issue for that, but we should also think about this as soon as possible because this affects to how we store the data from the manifest.

mssola:
Also @lsamayoa , note that we are updating the CONTRIBUTING.md file in the PR #248 . Since this is just the PR, this doesn't mean that these are the final contents of this file. This should help to avoid awkward situations like the one that just happenned of "I was already working on this" ;)

Gemfile
lsamayoa: I used it so we can memoize a RegistryClient inside the Registry model. I can do the memoization manually doe.

lsamayoa:
@mssola Catalog API is not that complex, it just lists the repos, so you have to actually use the Manifest API to update the whole repo.

I was already thinking about using sidekiq. But thought something like [crono] to be more appropiate so we don't have to add redis to the stack.

I was thinking that the synchronization should be indempotent and should only pull new stuff, not delete stuff is not there anymore, maybe we can mark stuff as deleted but I was not thinking that far.

[crono]: https://github.com/plashchynski/crono

mssola:
@lsamayoa I agree that is not that complex, but as you can see there are multiple options for doing this: its own process, sidekiq, crono, etc. So, in order to get this right, I'd prefer to investigate more on the different possibilities that we have and see which is the cleanest :)

mssola:
Just as a heads up, the RegistryClient was basically unreliable, but this was fixed with the PR #259. Moreover, all we wanted from the Catalog API has been merged in the PR #260 (by using crono, which is really cool, thanks @lsamayoa for the suggestion ;) ).

Therefore, since the Catalog API is already implemented and this PR needs to be rebased and re-worked anyways, I suggest to close the PR for now.

PR #244, opened 2015-08-01
setting machine fqdn as env variable

geoaxis:
this can help in easy creation of an AWS opsworks rails app.

### DIFF
--- a/deploy/before_migrate.rb
+++ b/deploy/before_migrate.rb
@@ -1,0 +1,7 @@
+rails_env = new_resource.environment["RAILS_ENV"]
+Chef::Log.info("Precompiling assets for RAILS_ENV=#{rails_env}...")
+execute "rake assets:precompile" do
+  cwd release_path
+  command "bundle exec rake assets:precompile"
+  environment "RAILS_ENV" => rails_env
+end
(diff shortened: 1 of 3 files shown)


deploy/before_migrate.rb
```diff
@@ -0,0 +1,9 @@
+rails_env = new_resource.environment["RAILS_ENV"]
+secret_key = new_resource.environment["SECRET_KEY_BASE"]
+Chef::Log.info("Precompiling assets for RAILS_ENV=#{rails_env}...")
```
flavio: This is going to fail if Chef is not in place. Can't we simply use Rails' logger?

flavio:
Can you also fix Rubocop's issue? It's just a matter of running `rubocop -a`

jordimassaguerpla:
I just added the database configuration to the appliance as an overlay file, so that we could merge this once the issues flavio mentions are fixed:

https://build.opensuse.org/package/rdiff/Virtualization:containers:Portus/PortusAppliance?linkrev=base&rev=55

deploy/before_migrate.rb
flavio: I just realized this file is considered only by Chef at deployment time. So ignore my comment.

flavio:
I'm going to take care of rubocop. Thanks for your submission.

PR #256, opened 2015-08-06
Remove containers when configuring the DB

mauromorales:
Hi guys, here are the changes we were doing on my machine in case you want to merge them into master.

Thanks again for the help!


mssola:
Thanks! :)

Issue #265, opened 2015-08-13
dashboard: show the repositories in which the current user has access to


mssola:
I'm thinking about re-doing the "Starred repos" container on the dashboard so it has tabs. One tab would be for repositories in which the current user has access to, and the other tab would be "starred repos". By tabs I mean something similar to what's done in Github with the list of repos.

flavio:
I'm already on it :)

PR #267, opened 2015-08-13
Show more details on the dashboard about repos

flavio:
It is now possible to show all the repositories an user has access to,
the personal repositories and the starred ones.

This fixes issue https://github.com/SUSE/Portus/issues/265

**NOTE WELL:** almost ready to be merged, we have to figure out why some tests are failing. Apparently it's a bug that has always been there, we simply never realized it.

### All the repositories
all

### Starred repositories
starred

### Personal repositories
personal

### DIFF
--- a/app/assets/javascripts/dashboard.coffee
+++ b/app/assets/javascripts/dashboard.coffee
@@ -1,0 +1,12 @@
+$(document).on "page:change", ->
+  $('#starred a').on 'click', (event) ->
+    e.preventDefault()
+    $(this).tab('show')
+
+  $('#all a').on 'click', (event) ->
+    e.preventDefault()
+    $(this).tab('show')
+
+  $('#personal a').on 'click', (event) ->
+    e.preventDefault()
+    $(this).tab('show')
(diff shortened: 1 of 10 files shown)


mssola:
The code LGTM, just fix the failing tests. Some notes on the UI though:
- There's too much padding on top of "Repositories".
- Tabs should have no separation with the contents of the container. That is, right now a selected tab looks like this: "Tab", then a grey line, then the contents. Instead, don't leave any separation between the tabs and the contents of it.
- There shouldn't be a line on top of the first listed repo.

flavio:
I fixed the broken tests, that turned to be caused by a nice corner case (see last commit).

flavio:
BTW, I have no idea why travis CI is unhappy with this PR, all is passing.

@mssola: I remember you had a similar issue in the past, can you help me?

mssola:
@flavio just run the test suite on your machine and open up the coverage/index.html file. In there you should be able to see where do you have uncovered code.

flavio:
I didn't think about code coverage, sorry!

PR #280, opened 2015-08-20
Remove explicit provides

jordimassaguerpla:
We are not using the provides for tracking the bundled gems, so we better don't have it because it creates noise.

I did it on a way that if we decide to get it back, we can revert the latest commit and we will have what we will probably need, that is "Provides: bundled(rubygem-foo) = XXX"


flavio:
Why aren't we using it? Don't we rely on that to have security mails for the gems required by the portus rpm?

jordimassaguerpla:
We don't rely on that. It was a proposal.

PR #281, opened 2015-08-20
Review production configuration

jordimassaguerpla:
This is required for creating an RPM that can be supported.

In order to simplify the configuration when installing from an RPM, I've moved the configuration to:

* environment variables for Portus configuration for production
* scripts for generating registry, ssl, apache and db configuration
* the environment variables for Portus will be set in the apache configuration

This way, the Portus configuration for production can be reused when deploying with other ways other than the RPM. However, if you deploy with the RPM, you can set up those variables in the apache configuration.

Also, external dependencies like registry, ssl, db, apache, can be easily configured by running some scripts. These scripts will provide default configurations, thus, the user can decide not to run them. For example, if the user has an external registry, or wants to use something else other than apache2, or apache2 is running on another fronted, or has his own ssl keys (which is recommended)...

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -20,7 +20,15 @@ test:
 
 production:
   <<: *default
-  host:     <%= ENV["PRODUCTION_HOST"] %>
-  username: <%= ENV["PRODUCTION_USERNAME"] %>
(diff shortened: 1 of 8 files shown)


flavio:
LGTM, merge once all the tests are passing

PR #282, opened 2015-08-20
Support auth protocol introduced by docker 1.8

flavio:
**NOTE WELL:** This is some preliminary work to fix issue 276. I'm about to leave for my vacation. @mssola please take a look at the code and vet it. I can fix the broken tests once I'm back on the 1st of September (unless you really want to take care of that once you are back from your vacation). 

## The issue

The authorization protocol changed with Docker 1.8:
  * client does pull or push or whatever
  * daemon does ping to registry i.e. https://registry.ip/v2/
  * registry returns 401 along with realm and service, but not scope
  * daemon asks for a token from the auth server, with
    service=<registry>&scope=repository:namespace/image:push,pull

In other words, the daemon always asks for push,pull, even if you are
just doing a pull. The auth server is supposed to respond in the
following fashion:

  * If unauthenticated access is not allowed, return a 401 requiring
user to authentictae
  * If unauthenticated access is allowed to that repo, return a web
token

When user tries the token path against the auth server with credentials:

  * If invalid credentials, return 401
  * If valid credentials, always return a 200 with a JWT that has the
    maximum credentials allowed this user on this repository in this service
    that is a subset of the scope provided.

The daemon will always ask for push,pull, and - as long as I am validly
authenticated - the auth server should always return 200 with a valid
Web token. The Web token will list the max I am allowed.

  * If I am not allowed push or pull, then return a token with no access
  * If I am allowed pull but not push, then return a token with pull
    access only
  * If I am allowed pull and push, then return a token with push and
    pull access

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ PR [#212](https://github.com/SUSE/Portus/pull/212).
 - Added the appliance tests. See PR [#208](https://github.com/SUSE/Portus/pull/208).
 - Star/Unstar repositories. See PR [#230](https://github.com/SUSE/Portus/pull/230).
 - Now users can be disabled. See PR [#240](https://github.com/SUSE/Portus/pull/240).
+- Fixed the authentication process for Docker 1.8. See PR
+[#282](https://github.com/SUSE/Portus/pull/282).
 - And some minor fixes here and there.
 
 ## 1.0.1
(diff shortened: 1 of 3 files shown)


mssola:
Ok, I've pushed a fix for a regression in the first commit. With this, Portus should be able to handle the changes from docker 1.8. I haven't added more tests for it, I'll do it later today.

mssola:
And this should be it. Rubocop is not passing because I pushed yesterday some stricter rules in rubocop. We can merge this anyways and I'll fix the style on master directly.

LGTM. @flavio what do you think ?

flavio:
Thanks @mssola !

PR #286, opened 2015-08-25
check secrets and ssl in production

jordimassaguerpla:
If secrets have not been set or ssl, render an error page. This will
prevent running unsafely in production but also provides with a
running application.

The message needs to be rewritten. I'll do that on a separate commit.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,7 @@
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
+  before_action :check_requirements
   before_action :authenticate_user!
   protect_from_forgery with: :exception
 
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -19,6 +20,25 @@ def after_sign_out_path_for(_resource)
 
   protected
 
+  def check_requirements
```
mssola: Instead of `fix_secrets = false` and then the if, do:
```ruby
fix_secrets = true if Rails.application.secrets.secret_key_base == "CHANGE_ME"
```
The same for `fix_ssl`. If the variable is not defined it will be nil, therefore evaluating to false anyways ;) Moreover, rubocop is complaining about the final if statement. Instead of doing that, you should:
```ruby
return unless fix_secrets || fix_ssl
```
And then proceed with the code.

Moreover, I'd appreciate if you could add documentation to this method :)

jordimassaguerpla:
I just rebased according to mssola comments

mssola:
There are still some failures, but it looks like it is a commit that I pushed directly before :/ Thanks :+1:

PR #297, opened 2015-08-31
Check secrets and ssl in production 2

jordimassaguerpla:
This PR adds additional checks and a nicer ui for error handling.

If you want to see how it would look in production, set config.consider_all_requests_local to false in  config/environments/development.rb.

Here a couple of screenshots:

if database is not properly configured

http://ibin.co/2E3H61mok295

if secrets are not properly configured

http://ibin.co/2E3HNCRxMigl


app/assets/javascripts/errors.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
mssola: Remove this file. It's not used.

app/assets/stylesheets/errors.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the errors controller here.
```
mssola: Remove this file. It's not used.

app/helpers/errors_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ErrorsHelper
```
mssola: Remove this file

config/routes.rb
```diff
@@ -1,5 +1,5 @@
 Rails.application.routes.draw do
-
+  resources :errors, only: [:show]
```
mssola: Instead of doing this, maybe do this:
```ruby
resources :errors, only: [:show], constraints: { id: /\d{3}/ }
```
Note that now the `status` parameter has been renamed to id (to follow Rails conventions on the show parameter).

app/controllers/application_controller.rb
```diff
@@ -22,21 +22,22 @@ def after_sign_out_path_for(_resource)
...
-
-    text = "Please review the following configurations"
-    text += "<ul>"
-    text += "<li>ssl</li>" if fix_ssl
-    text += "<li>secrets</li>" if fix_secrets
-    text += "</ul>"
-    render text: text, status: 500
+    # select any model to check database connection
+    # if database is not configured it will raise an exception
+    User.first
```
mssola: There are better ways to check that you have a connection to the DB I think. This looks just to hackish...

app/controllers/application_controller.rb
```diff
@@ -22,21 +22,22 @@ def after_sign_out_path_for(_resource)
...
+    fix= fixes
```
mssola: space before operator

app/controllers/application_controller.rb
```diff
@@ -22,21 +22,22 @@ def after_sign_out_path_for(_resource)
...
+    return unless fix.has_value?(true)
+    redirect_to "/errors/500?fix_ssl=#{fix[:ssl]}&fix_secret_key_base=#{fix[:secret_key_base]}&fix_secret_machine_fqdn=#{fix[:secret_machine_fqdn]}&fix_secret_encryption_private_key_path=#{fix[:secret_encryption_private_key_path]}&fix_secret_portus_password=#{fix[:secret_portus_password]}", status: 500
   end
 
   def fixes
-    fix_secrets = true if Rails.application.secrets.secret_key_base == "CHANGE_ME"
-    fix_ssl     = true if Rails.env.production? && !request.ssl?
-    [fix_secrets, fix_ssl]
+    fix = {}
```
mssola: use `tap`, as it's done here: https://github.com/SUSE/Portus/blob/master/app/models/jwt_token.rb#L12

app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,14 @@
+class ErrorsController < ApplicationController
+  skip_before_filter :check_requirements
+  skip_before_filter :authenticate_user!
+  def show
```
mssola: Leave an empty line before the method definition

app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+    @fix_database = env["action_dispatch.exception"].class.name.starts_with? "ActiveRecord"
```
mssola: On the same note as before, I'm sure there's a better/cleaner way to do this.

app/views/errors/show.html.erb
```diff
@@ -0,0 +1,84 @@
+<html>
```
mssola: Couldn't you use the main layout ?

config/environments/production.rb
```diff
@@ -80,7 +80,11 @@
   # Run pending migrations
   unless ENV["SKIP_MIGRATION"]
     config.after_initialize do
-      ActiveRecord::Migrator.migrate(Rails.root.join("db/migrate"), nil)
+      begin
+        ActiveRecord::Migrator.migrate(Rails.root.join("db/migrate"), nil)
+      rescue
+        $stderr.puts "Error running migration! Please review database configuration"
```
mssola: Use the Rails logger

app/controllers/application_controller.rb
```diff
@@ -22,21 +22,22 @@ def after_sign_out_path_for(_resource)
...
-    text += "<ul>"
-    text += "<li>ssl</li>" if fix_ssl
-    text += "<li>secrets</li>" if fix_secrets
-    text += "</ul>"
-    render text: text, status: 500
+    # select any model to check database connection
+    # if database is not configured it will raise an exception
+    User.first
+    fix= fixes
+    return unless fix.has_value?(true)
```
mssola:
```
return if fix.empty?
```

mssola:
Could you write a test to check that in a faulty configuration you would get redirected to the errors page ?

config/environments/production.rb
jordimassaguerpla: Rails logger has not yet been initialized here and it is nil.

app/views/errors/show.html.erb
jordimassaguerpla: the main layout has calls to the database thus it will create an infinite loop if the error is the database connection.

jordimassaguerpla:
@mssola : I've done the fixes you asked me. Please could you take a second look?
I am writing the tests and I'll add them in short.

config/routes.rb
```diff
@@ -42,5 +42,6 @@
       put "toggle_admin", on: :member
     end
   end
+  match "(errors)/:status", to: "errors#show", constraints: { status: /\d{3}/ }, via: :all
```
mssola: And now you can remove this match no ?

jordimassaguerpla: no i can't. This match is for routing the (errors) to the errors controller. By (errors) I mean 500, 400 errors, like when an exception is raised.

config/routes.rb
```diff
@@ -1,5 +1,5 @@
 Rails.application.routes.draw do
-
+  resources :errors, only: [:show], constraints: { id: /\d{3}/ }
```
mssola: So if that match is correct, I was wrong in telling you to add this constraint. You can safely remove this `constraints` key, sorry.

config/routes.rb
flavio: Does that include also the errors raised in other circumstances (like the 404 error trying to view a non existing team)?

jordimassaguerpla: Yes it does. Should we handle it differently? I don't know this case.

flavio: No, I just think it's great.

jordimassaguerpla:
I've fixed the tests and rebased the branch. Thanks for your comments so far. Could you take another look?

app/views/errors/show.html.erb
```diff
@@ -0,0 +1,85 @@
...
+  <title>Portus error page</title>
+  <%= stylesheet_link_tag 'application', media: 'all' %>
+  <%= javascript_include_tag 'application'%>
+  <%= csrf_meta_tags %>
+</head>
+<body>
+  <center><%= image_tag "layout/portus-error.png" %></center>
+  <% if @fix.value?(true) %>
+  <p>
+  <center><h1> Ops... something went wrong...</h1></center>
```
flavio: You can remove the space before "Ops"

spec/controllers/errors_controller_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+describe ErrorsController do
+  describe "GET #show" do
+
+    before :all do
+      secrets = Rails.application.secrets
+      @secret_key_base = secrets.secret_key_base
+      @secret_machine_fqdn = secrets.machine_fqdn
+      @secret_encryption_private_key_path = secrets.encryption_private_key_path
+      @secret_portus_password = secrets.portus_password
+    end
```
flavio: You can get rid of this block because you have the same code inside of `before :each`

spec/controllers/errors_controller_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      secrets.portus_password = @secret_portus_password
+    end
+
+    after :all do
+      secrets = Rails.application.secrets
+      secrets.secret_key_base = @secret_key_base
+      secrets.machine_fqdn = @secret_machine_fqdn
+      secrets.encryption_private_key_path = @secret_encryption_private_key_path
+      secrets.portus_password = @secret_portus_password
+    end
```
flavio: Do we really need that? Is that because otherwise the other tests are screwed up?

jordimassaguerpla: exactly

spec/controllers/errors_controller_spec.rb
jordimassaguerpla: It is not the same. Note the "@". This is to get a backup of the values and restore them before each test example and after all.

spec/features/errors_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    around :each do |example|
+      Rails.application.config.consider_all_requests_local = false
+      Rails.application.config.action_dispatch.show_exceptions = true
+      Rails.env = ActiveSupport::StringInquirer.new("production")
+      example.run
+      Rails.application.config.consider_all_requests_local = true
+      Rails.application.config.action_dispatch.show_exceptions = false
+      Rails.env = ActiveSupport::StringInquirer.new("test")
+    end
```
flavio: I think it would be safer to rewrite the code in that way:

```ruby
around :each do |example|
  begin
    Rails.application.config.consider_all_requests_local = false
    Rails.application.config.action_dispatch.show_exceptions = true
    Rails.env = ActiveSupport::StringInquirer.new("production")
    example.run
  ensure
    Rails.application.config.consider_all_requests_local = true
    Rails.application.config.action_dispatch.show_exceptions = false
    Rails.env = ActiveSupport::StringInquirer.new("test")
  end
end
```

spec/controllers/errors_controller_spec.rb
flavio: I see, sorry

flavio:
Please fix Rubocop's complains (`rubocop -a`)

flavio:
LGTM, waiting for @mssola's feedback

jordimassaguerpla:
@flavio : thanks. I've reviewed according to your changes. Let's wait for @mssola feedback.

PR #299, opened 2015-08-31
Review packaging scripts

jordimassaguerpla:
Review scripts so that they can be used for configuration after installing the rpm


packaging/suse/scripts/configure_database.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+if [[ $(id -u) -ne 0 ]] ; then echo "Please run as root" ; exit 1 ; fi
```
mssola: Don't write if statements as one-liners

mssola:
LGTM. Just fix the conflicts

PR #314, opened 2015-09-08
set admin user from a rake task and disable first-user is admin

### LINKED ISSUE #313: set admin user from a rake task and disable first-user is admin
For security reason, I would propose that we have an option to disable the "first-user is admin" feature and instead add a rake task to set the admin user.

jordimassaguerpla:
fixes #313

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -39,3 +39,12 @@ ldap:
   guess_email:
     enabled: false
     attr: ""
+
+# Set first_user_admin to true if you want that the first user that signs up
+# to be an admin.
+#
+# Set to false otherwise. Then you will need to run
+#   rake portus:set_admin
+# in order to set the admin user
+first_user_admin:
+  enabled: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Upcoming Version
 
+- set admin user from a rake task and disable first-user is admin. See PR [#314]
+  (https://github.com/SUSE/Portus/pull/314)
 - Introduced LDAP support. See PR [#301](https://github.com/SUSE/Portus/pull/301).
 - Users will not be able to create namespaces without a Registry currently
 existing.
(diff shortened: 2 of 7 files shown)


jordimassaguerpla:
I've set false as default because it is the safest option, but it breaks backward compatibility ... is it ok?

config/config.yml
```diff
@@ -39,3 +39,14 @@ ldap:
   guess_email:
     enabled: false
     attr: ""
+
+# Set first_user_admin to true if you want that the first user that signs up
+# becomes the admin.
```
flavio: `Set first_user_admin to true if you want that the first user that signs up to be an admin.`

flavio:
yeah, I'm a bit sceptical about breaking the compatibility. I wouldn't do it

mssola:
Same here, default to true. Moreover, the config file is wrong, it should have something set always, so comment out the code. It should be like this:

```yaml
# Set to true if you want that the first user that signs up to be an admin. If
# this is false, then you will need to perform the following command to set the admin user:
#  $ rake portus:set_admin
first_user_admin:
  enabled: true
```

Then, when checking about this in the Ruby code, yo do:

```ruby
APP_CONFIG["first_user_admin"].enabled?
```

mssola:
Oh, and by the way, could you update the Changelog ?

jordimassaguerpla:
does it look good now :)

config/config.yml
```diff
@@ -39,3 +39,14 @@ ldap:
...
     attr: ""
+
+# Set first_user_admin to true if you want that the first user that signs up
+# to be an admin.
+#
+# Set to false otherwise. Then you will need to run
+#   rake portus:set_admin
+# in order to set the admin user
+#
+
```
mssola: empty line not needed

config/config.yml
```diff
@@ -39,3 +39,14 @@ ldap:
...
     enabled: false
     attr: ""
+
+# Set first_user_admin to true if you want that the first user that signs up
+# to be an admin.
+#
+# Set to false otherwise. Then you will need to run
+#   rake portus:set_admin
+# in order to set the admin user
+#
```
mssola: not needed

mssola:
LGTM

flavio:
+1

PR #317, opened 2015-09-08
[compose-setup]: Use the current DOCKER_HOST value if already exists

phspagiari:
The script is forcing the `DOCKER_HOST` value as the hostname of the current machine, but some people use a custom `DOCKER_HOST`.

This small fix will keep the `DOCKER_HOST` value if already exists or replace with the `hostname -f` if not.

### DIFF
--- a/compose-setup.sh
+++ b/compose-setup.sh
@@ -44,7 +44,7 @@ clean() {
 }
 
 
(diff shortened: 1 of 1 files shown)


flavio:
Cool, thanks!

PR #319, opened 2015-09-09
add crono systemd scripts

### LINKED ISSUE #261: Create systemd unit file for portus-cronos
We recently started to use `cronos` to handle some background jobs. We have to ensure the cronos service is started automatically inside of the appliance.

I think the cleanest solution is to create a systemd unit file for that.

Note well: cronos must be started in foreground mode otherwise systemd will have troubles handling it.

@jordimassaguerpla: I can help you with that if you want

jordimassaguerpla:
fixes #261

### DIFF
--- a/packaging/suse/conf/portus_crono.service
+++ b/packaging/suse/conf/portus_crono.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=A time-based background job scheduler daemon
+After=network.target
+
+[Service]
+ExecStart=/usr/sbin/portus_crono
+ExecReload=/usr/bin/kill -s SIGHUP $MAINPID
+Restart=on-abort
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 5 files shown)


packaging/suse/conf/portus_crono.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=A time-based background job scheduler daemon
+After=apache2.service
```
flavio: I don't think it required apache2 to be running. @mssola does the code executed by crono talk with Portus via its API? I think it goes straight to the DB. If that's the case we would need to have mariadb up and running, not apache.

flavio:
LGTM except for the comment I made. We can ignore Travis' results, they are not related with these changes.

packaging/suse/conf/portus_crono.service
jordimassaguerpla: mariadb can be in another server thus I would instead propose to have network running.

mssola: Apache is indeed not needed. We only need the DB.

PR #320, opened 2015-09-09
[packaging] we do not need to make explicit the required libraries

jordimassaguerpla:
rpmbuild will guess it.


flavio:
I agree, these are not needed.

PR #328, opened 2015-09-15
[packaging] move apache conf to /etc/apache/vhosts.d/portus.conf.sample

jordimassaguerpla:
This way this is owned by the package, so we can update it on future
releases, but the user can still override settings in the final file


mssola:
LGTM

PR #329, opened 2015-09-15
ldap: fixed LDAP tests

mssola:
There were some issues with tests requiring LDAP to be enabled. To fix that,
LDAP will be enabled always for the test suite. However, the test suite mocks
away the Portus::LDAP class so the `authenticate!` method fails gracefully for
regular tests. Tests that require the legit version of the `authenticate!`
method have to call the following inside the `before(:each)` block:

```ruby
allow_any_instance_of(Portus::LDAP).to receive(:authenticate!).and_call_original
```


flavio:
LGTM, awesome

PR #332, opened 2015-09-15
More specific error-messages for the implementation of ldap

### LINKED ISSUE #310: Be more specific with LDAP errors
An LDAP connection can fail either because the given credentials are not properly set, or because there's something wrong with the given LDAP configuration. We should be more clear on this regard by:

- Properly logging the given error.
- Alert the user with different messages for each case (right now it's only `:invalid_login`).

jloehel:
Fixes #310

### DIFF
--- a/config/locales/devise.en.yml
+++ b/config/locales/devise.en.yml
@@ -17,7 +17,10 @@ en:
       unauthenticated: "You need to sign in or sign up before continuing."
       unconfirmed: "You have to confirm your email address before continuing."
       user:
-        invalid_login: "Invalid %{authentication_keys} or password."
+        ldap_failed: "The LDAP-support is not enabled."
+        ldap_bind_failed: "Invalid %{authentication_keys} or password."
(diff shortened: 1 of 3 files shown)


config/locales/devise.en.yml
```diff
@@ -17,7 +17,10 @@ en:
       unauthenticated: "You need to sign in or sign up before continuing."
       unconfirmed: "You have to confirm your email address before continuing."
       user:
-        invalid_login: "Invalid %{authentication_keys} or password."
+        ldap_failed: "The LDAP-support is not enabled."
+        ldap_bind_failed: "Invalid %{authentication_keys} or password."
+        invalid_username: "Invalid username. The generation of the personal namespace failed."
```
mssola: This message is misleading since this is not necessarily the case. I'd say something like "User could not be registered. Invalid username.".

mssola:
LGTM. Just fix that small issue :)

PR #334, opened 2015-09-15
[packaging] split create_database

jordimassaguerpla:
we need the create_database.sh script because you call it in the
appliance first-boot


flavio:
LGTM

PR #335, opened 2015-09-15
add init boot script for portus

jordimassaguerpla:
This script has been used in the appliance. However, we need this same
script for setting up some test scenarios with default vaules


mssola:
LGTM

Issue #250, opened 2015-08-03
Don't autocomplete on the signup form


mssola:
Right now the signup form has auto-completion enabled for all the input elements. This should not happen.

PR #337, opened 2015-09-16
Added the autocomplete attribute to the signup-form and turned it off.

jloehel:
Fixes: #250


flavio:
LGTM

PR #344, opened 2015-09-17
Add the 'timezone' requirement to the Portus rpm

flavio:
On a really minimal system (e.g. a Docker image) the timezone package
is usually missing, that would cause some rubygem to fail.


jordimassaguerpla:
lgtm

PR #346, opened 2015-09-17
Set log level for production env

flavio:
This is required to get rid of a deprecation warning:

> DEPRECATION WARNING: You did not specify a `log_level` in
> `production.rb`. Currently, the default value for `log_level` is
> `:info` for the production environment and `:debug` in all other
> environments. In Rails 5 the default value will be unified to `:debug`
> across all environments. To preserve the current setting, add the
> following line to your `production.rb`:


jordimassaguerpla:
lgtm

PR #349, opened 2015-09-18
Appliance hostname

flavio:
Improve/fix how we handle hostname when installing Portus' RPM


flavio:
Code fixed as discussed with @jordimassaguerpla

PR #350, opened 2015-09-18
Make it possible to disable ssl checks on production

flavio:
The default is to enforce ssl usage on production, however in certain circumstances (e.g.: out internal integration test suite) it's actually more convenient to not use ssl.


mssola:
LGTM. Update the wiki https://github.com/SUSE/Portus/wiki/Configuring-Portus#list-of-configuration-options.

flavio:
@mssola the wiki has been updated, thanks

PR #369, opened 2015-09-28
Fix broken install instructions

bencevans:
When `docker-compose up` was run, it hit the following error:

    Couldn't find env file: [redacted]/Portus/docker/environment


bencevans:
Just noticed this has been done also in https://github.com/SUSE/Portus/commit/b82330b56c36c67988ab9b8fa77ba1fb15e8dba5

PR #372, opened 2015-09-29
portusctl: add exec subcommand

flavio:
This is handy for starting arbitrary scripts via bunndler exec (i.e. portusctl exec rails c production)


mssola:
LGTM, just fix rubocop :)

Issue #201, opened 2015-07-06
Add additional info for namespaces and teams pages


mssola:
Like description, github links, etc.

PR #383, opened 2015-10-02
Added additional information for teams and namespaces pages

jloehel:
Fixes: #201

### DIFF
--- a/app/controllers/namespaces_controller.rb
+++ b/app/controllers/namespaces_controller.rb
@@ -1,6 +1,8 @@
 class NamespacesController < ApplicationController
+  include ChangeDescription
+
   respond_to :html, :js
-  before_action :set_namespace, only: [:toggle_public, :show]
+  before_action :set_namespace, only: [:toggle_public, :show, :update]
   before_action :check_team, only: [:create]
 
   after_action :verify_authorized, except: :index
@@ -44,6 +46,12 @@ class NamespacesController < ApplicationController
         format.js { respond_with @namespace.errors, status: :unprocessable_entity }
       end
     end
+  end
+
+  # PATCH/PUT /namespace/1
+  # PATCH/PUT /namespace/1.json
+  def update
+    change_description(@namespace, :namespace)
   end
 
   # PATCH/PUT /namespace/1/toggle_public
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -1,5 +1,7 @@
 class TeamsController < ApplicationController
-  before_action :set_team, only: [:show]
+  include ChangeDescription
+
+  before_action :set_team, only: [:show, :update]
   after_action :verify_policy_scoped, only: :index
   respond_to :js, :html
 
@@ -31,6 +33,12 @@ class TeamsController < ApplicationController
     end
   end
 
+  # PATCH/PUT /teams/1
+  # PATCH/PUT /teams/1.json
+  def update
+    change_description(@team, :team)
+  end
+
   private
 
   def set_team
@@ -38,6 +46,6 @@ class TeamsController < ApplicationController
   end
 
   def team_params
-    params.require(:team).permit(:name)
+    params.require(:team).permit(:name, :description)
   end
 end
--- a/app/policies/namespace_policy.rb
+++ b/app/policies/namespace_policy.rb
@@ -33,6 +33,10 @@ class NamespacePolicy
     push?
   end
 
+  def update?
+    push?
+  end
+
   def toggle_public?
     !namespace.global? && (user.admin? || namespace.team.owners.exists?(user.id))
   end
--- a/app/policies/team_policy.rb
+++ b/app/policies/team_policy.rb
@@ -11,6 +11,11 @@ class TeamPolicy
     user.admin? || @team.users.exists?(user.id)
   end
 
+  def owner?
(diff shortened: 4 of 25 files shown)


app/controllers/teams_controller.rb
```diff
@@ -31,6 +31,24 @@ def create
...
+    @team.update(team_params)
+
+    @team.create_activity :change_team_description,
+                          owner: current_user,
+                          recipient: @team,
+                          parameters: {
+                            old_description: old_description,
+                            new_description: team_params['description']}
+
+    render js: "window.location = '/teams/#{@team.id}'"
```
flavio: Do we really need to force a reload of the entire page, do you think it would be too hard/an overkill to update the different parts of the page?

flavio: I just noticed there's a team_updates.js file, doesn't that make this directive wrong?

flavio:
It would be nice to see some screenshots. Also please fix the rubuto issues.

app/controllers/teams_controller.rb
jloehel: No it's not really needed. You are right. I will move the reload to app/views/teams/update.js.erb and limit the reloading to the specific div-tag.

app/controllers/namespaces_controller.rb
```diff
@@ -46,6 +46,24 @@ def create
...
+    @namespace.update(namespace_params)
+
+    @namespace.create_activity :change_namespace_description,
+                               owner:             current_user,
+                               recipient:         @namespace,
+                               parameters:        {
+                                 old_description: old_description,
+                                 new_description: namespace_params["description"] }
+
+    render js: "window.location = '/namespaces/#{@namespace.id}'"
```
mssola: As @flavio pointed out, don't reload the whole page. Moreover, the activity can get *crazy* since descriptions are just unlimited text. Maybe this can be hidden when showing the view somehow ?

app/controllers/teams_controller.rb
```diff
@@ -31,6 +31,24 @@ def create
     end
   end
 
+  # PATCH/PUT /teams/1
+  # PATCH/PUT /teams/1.json
+  def update
```
mssola: This is *identical* to the one from the namespaces_controller. Maybe we could merge functionality through a concern.

app/controllers/teams_controller.rb
```diff
@@ -38,6 +56,6 @@ def set_team
   end
 
   def team_params
-    params.require(:team).permit(:name)
+    params.require(:team).permit(:name, :description)
```
mssola: Revert this, since it looks like for update you just want to modify the description, not the name.

app/policies/namespace_policy.rb
```diff
@@ -33,6 +33,10 @@ def create?
     push?
   end
 
+  def update?
```
mssola: why ?

app/policies/team_policy.rb
```diff
@@ -11,6 +11,12 @@ def member?
     user.admin? || @team.users.exists?(user.id)
   end
 
+  def isOwner?
```
mssola: ```isOwner?``` -> ```owner?```

spec/controllers/teams_controller_spec.rb
```diff
@@ -104,6 +104,25 @@
     end
   end
 
+  describe "PATCH #update" do
+    it "does not allow to change the description by viewers and contributers" do
+      disallowed_roles = Array["viewer", "contributer"]
```
mssola: Drop the "Array" thing

mssola:
Snapshots please :)

app/controllers/namespaces_controller.rb
jloehel: Shall I limit the length of the description? The view of the activity shows not the value of the description. Only if it has been edited or deleted. Only the csv contains this information. So, I guess I can remove the parameters?

jloehel:
Some screenshots:

screenshot - 10022015 - 05 43 13 pm
screenshot - 10022015 - 05 43 35 pm
screenshot - 10022015 - 05 44 35 pm

flavio:
There are a couple of things that I don't like:

## 1st picture
  * I would change "Information about testing" to "Testing group".
  * I would drop the "description" header and just put the description straight into the "Testing group" box
  * I would move the "edit" icon straight into the line where "Testing group" is written
  * I would rename "Member of testing" to "Members"
  * I would rename "Namespaces owned by testing" to "Namespaces owned"

## 2nd picture
  * Everything looks fine once the changes proposed to slide #1 will be in place.

## 3rd picture
  * Everything is good.

mssola:
I totally agree with @flavio. Note that:

> - I would rename "Member of testing" to "Members"
> - I would rename "Namespaces owned by testing" to "Namespaces owned"

These two are unrelated to this change, but maybe it's a good idea to change this while we are at it in this PR :)

jloehel:
Additional screenshots:

screenshot - 10072015 - 03 22 28 pm
screenshot - 10072015 - 03 23 04 pm
screenshot - 10072015 - 03 28 08 pm

flavio:
That looks definitely better!

mssola:
LGTM

mssola:
@flavio do you agree on mergin this ?

PR #385, opened 2015-10-02
portusctl: show example of Registry's configuration

flavio:
Show a working example of Registry's configuration when the Registry is
not running on the same host of Portus.

### DIFF
--- a/packaging/suse/portusctl/lib/template_writer.rb
+++ b/packaging/suse/portusctl/lib/template_writer.rb
@@ -3,14 +3,25 @@
 class TemplateWriter
   # Process the given template and writes it to the final destination
   def self.process(template_name, output, context)
-    template = File.join(
-      File.expand_path("../../templates", __FILE__),
-        template_name)
-
-    conf  = ERB.new(File.read(template), nil, "<>").result(context)
+    t = render(template_name, context)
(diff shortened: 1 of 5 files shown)


packaging/suse/portusctl/lib/template_writer.rb
```diff
@@ -3,14 +3,23 @@
...
-    conf  = ERB.new(File.read(template), nil, "<>").result(context)
+    t = render(template_name, context)
 
     File.open(output, "w") do |file|
-      file.write(conf)
+      file.write(t)
     end
   end
+
+  def self.render(template_name, context)
```
mssola: Could you add documentation on this method ? :)

packaging/suse/portusctl/lib/template_writer.rb
```diff
@@ -3,14 +3,23 @@
...
+    ERB.new(load_template(template_name), nil, "<>").result(context)
+  end
+
+  private
+
+  def self.load_template(template_name)
```
mssola: same here

mssola:
LGTM

mssola:
Rubocop is not liking this PR. Could you fix that ? :)

PR #387, opened 2015-10-04
Replace Font Awesome symbol with SVG.

### LINKED ISSUE #312: Use something else for the user icon when gravatar is disabled
Right now we are using the one as given by font-awesome. But it's ugly and it doesn't align with the text that is besides it.

brycereitano:
Closes #312 

screen shot 2015-10-04 at 11 29 06

Signed-off-by: Bryce Reitano <bryce@rocketmade.com>


brycereitano:
Test runner was using some cached changes, fixing now.

brycereitano:
Looks like there is a deprecated method that needs to be updated, causing the build to fail. Other than that, my build environment seems to be in place to avoid multiple builds in the future.

mssola:
LGTM. The rubocop failure seems unrelated to this change.

PR #388, opened 2015-10-05
Add table icons for namespaces and teams.

### LINKED ISSUE #36: Add support for for Roles and type of Namespace icons
In the current namespaces tab we should display next to each namespace two additional icons
  - a **Role icon**
  - a **Namespace type** icon

A Role icon  represents the role owned by the user on the namespace (owner | collaborator | viewer).

A Namespace type icon can be of two different types, personal or team. 
As a personal namespace is practically speaking a namespace owned by a team composed by only        one person, the distinction should be done on the number of members the team is composed of.

brycereitano:
Closes #36

Additionally, I fixed the deprecation that RuboCop was complaining about. 

Many of the icons are debatable, but I felt these fit the bill pretty well.

### Screenshots

portus_1
portus_2
portus_3

Signed-off-by: Bryce Reitano <bryce@rocketmade.com>

### DIFF
--- a/app/helpers/namespaces_helper.rb
+++ b/app/helpers/namespaces_helper.rb
@@ -1,0 +1,5 @@
+module NamespacesHelper
+  def can_manage_namespace?(namespace)
+    current_user.admin? || namespace.team.owners.exists?(current_user.id)
+  end
+end
--- a/app/helpers/teams_helper.rb
+++ b/app/helpers/teams_helper.rb
@@ -12,4 +12,31 @@ module TeamsHelper
       "-"
     end
   end
+
+  # Render the namespace scope icon.
+  def team_scope_icon(team)
+    if team.team_users.count > 1
+      icon = "fa-users"
+      title = "Team"
+    else
+      icon = "fa-user"
+      title = "Personal"
+    end
+
+    content_tag :i, "", class: "fa #{icon} fa-lg", title: title
+  end
+
+  # Render the team user role icon.
+  def team_user_role_icon(team_user)
+    icon = case team_user.role
+           when "owner"
+             "fa-key"
+           when "contributor"
+             "fa-exchange"
+           when "viewer"
+             "fa-eye"
+           end
+
+    content_tag :i, "", class: "fa #{icon} fa-lg", title: team_user.role.titleize
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ and [#294](https://github.com/SUSE/Portus/pull/294).
 - Now users can be enabled/disabled. See PR [#240](https://github.com/SUSE/Portus/pull/240).
 - Fixed the authentication process for Docker 1.8. See PR
 [#282](https://github.com/SUSE/Portus/pull/282).
+- Added icons to the following tables: teams and members. See PR
+[#388](https://github.com/SUSE/Portus/pull/388).
 - And some minor fixes here and there.
 
 ## 1.0.1
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -14,17 +14,9 @@
 @import 'images';
 @import 'dashboard';
 @import 'panels';
+@import 'tables';
 @import 'forms';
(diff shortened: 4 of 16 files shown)


flavio:
I, first of all thanks for your pull request!

I talked with @mssola earlier. We don't like the icons used for the namespace. We think it's probably better to avoid icons for namespaces.

The icons for the teams a good. The same applies to the icon of the "viewer" and "contributor" roles. The icon for the "admin" role is not obvious at first glance, but we cannot think about anything else.

Let's wait for @eotchi's opinion before merging the code.

app/helpers/namespaces_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  # Render the namespace scope icon.
+  def namespace_scope_icon(namespace)
+    icon = "fa-square"
+    title = "Personal"
+
+    if namespace.team.team_users.count > 1
+      icon = "fa-sitemap"
+      title = "Shared"
+    end
```
mssola: I'd prefer:

```ruby
if namespace.team.team_users.count > 1
  icon = "fa-sitemap"
  title = "Shared"
else
  icon = "fa-square"
  title = "Personal"
end
```

app/helpers/teams_helper.rb
```diff
@@ -12,4 +12,32 @@ def role_within_team(team)
       "-"
     end
   end
+
+  # Render the namespace scope icon.
+  def team_scope_icon(team)
```
mssola: Same as for the `namespace_scope_icon` method

app/helpers/teams_helper.rb
```diff
@@ -12,4 +12,32 @@ def role_within_team(team)
...
+      icon = "fa-users"
+      title = "Team"
+    end
+
+    content_tag :i, "", class: "fa #{icon} fa-lg", title: title
+  end
+
+  # Render the team user role icon.
+  def team_user_role_icon(team_user)
+    icon = "fa-times-circle"
```
mssola: Put this inside of the `else` block inside of the `case` statement below.

brycereitano: Would you prefer? The second seems weird to me, because variable scope seems implicit. 

```ruby
    icon = case team_user.role
           when "owner"
             "fa-bolt"
           when "contributor"
             "fa-exchange"
           when "viewer"
             "fa-eye"
           else
             "fa-times-circle"
           end
```
or
```ruby
    case team_user.role
    when "owner"
      icon = "fa-bolt"
    when "contributor"
      icon = "fa-exchange"
    when "viewer"
      icon = "fa-eye"
    else
      icon = "fa-times-circle"
    end
```

brycereitano:
@flavio Are we all agreed to remove the namespace icons?

app/helpers/teams_helper.rb
mssola: Yes, let's use the first one :)

mssola:
@brycereitano at least me and @flavio agreed on it. But I'd like to hear from @eotchi first :)

eotchi:
I agree too - as long as the icons for the namespaces are not representative enough I'd rather stay without them.

mssola:
LGTM, could you squash these two commits into a single one ?

brycereitano:
@mssola @flavio One of the designers that I co-work with has suggested these for the roles.
Owner: [fa-key](http://fortawesome.github.io/Font-Awesome/icon/key/)
Contributor: [fa-pencil](https://fortawesome.github.io/Font-Awesome/icon/pencil) or [fa-exchange](https://fortawesome.github.io/Font-Awesome/icon/exchange)
Viewer: [fa-eye](https://fortawesome.github.io/Font-Awesome/icon/eye)

Owner: I think switching from the bolt to the key would be more descriptive to the user.
Contributor: I'm for either exchange or pencil.
Viewer: Is already the eye.

mssola:
@brycereitano I agree, the `fa-key` seems to fit better. For contributor, I'd go for the `fa-exchange`.

flavio:
I agree with Miquel

mssola:
Thanks! :)

Issue #231, opened 2015-07-23
Add policy checks to repositories_controller


flavio:
We should make sure the current_user is actually entitled to see a certain repository. Also the list of repositories to show should be limited to what the user can actually access.

Issue #225, opened 2015-07-10
Hide links in the activities stream regarding actions that an unauthorized user cannot see.


mssola:
Imagine that the `user1` is an admin and then it performs the following actions:

1. Makes his personal namespace public.
2. Then he makes the namespace private again.

These actions will appear in the stream of `user2`, a regular user that cannot access the personal namespace of `user1` after it has been set to private. However, these actions are presented as links, so `user2` can be tempted to click on them, and therefore get redirected to an "Access Denied" page.

This should be corrected, so the messages do not include links for unauthorized users.

PR #410, opened 2015-10-09
Added policy checks to the repositories_controller and modified the activities policy.

jloehel:
Fixes: #231
Fixes: #225

### DIFF
--- a/app/policies/repository_policy.rb
+++ b/app/policies/repository_policy.rb
@@ -1,4 +1,17 @@
 class RepositoryPolicy
+  attr_reader :user, :repository
+
+  def initialize(user, repository)
+    @user = user
+    @repository = repository
+  end
+
+  def show?
+    @user.admin? ||
+      @repository.namespace.public? ||
+      @repository.namespace.team.users.exists?(user.id)
+  end
+
   class Scope
     attr_reader :user, :scope
 
@@ -8,12 +21,18 @@ class RepositoryPolicy
     end
 
     def resolve
-      @scope
-        .joins(namespace: { team: :users })
-        .where("namespaces.public = :namespace_public OR " \
-               "users.id = :user_id",
-               namespace_public: true, user_id: @user.id)
-        .distinct
+      if user.admin?
+        @scope.all
+      else
+        # Show repositories only if the repository is public or
+        # the repository belongs to the current_user
+        @scope
+          .joins(namespace: { team: :users })
+          .where("namespaces.public = :namespace_public OR " \
+                 "users.id = :user_id",
+                  namespace_public: true, user_id: @user.id)
+          .distinct
+      end
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/policies/repository_policy.rb
```diff
@@ -1,4 +1,15 @@
 class RepositoryPolicy
+  attr_reader :user, :repository
```
mssola: Do you really need this ?

spec/controllers/repositories_controller_spec.rb
```diff
@@ -12,7 +12,10 @@
   describe "GET #index" do
 
     it "assigns all repositories as @repositories" do
-      repository = create(:repository)
+      team = create(:team)
+      TeamUser.create(team: team, user: user, role: TeamUser.roles["owner"])
+      namespace = create(:namespace, team: team)
+      repository = create(:repository, namespace: namespace)
       get :index, {}, valid_session
```
flavio: Could you please also create a repository that is not visible by the logged in user?

spec/controllers/repositories_controller_spec.rb
```diff
@@ -22,7 +25,10 @@
   describe "GET #show" do
 
     it "assigns the requested repository as @repository" do
-      repository = create(:repository)
+      team = create(:team)
+      TeamUser.create(team: team, user: user, role: TeamUser.roles["owner"])
+      namespace = create(:namespace, team: team)
+      repository = create(:repository, namespace: namespace)
```
flavio: I think this part can be refactored because it's common to the two tests.

I would define a `let(:visible_repository) { create(:repository, namespace: namespace) }` and all the other objects involved at the top of the spec.

app/policies/repository_policy.rb
jloehel: Otherwise the variable is not visible for the test, or not?

mssola: t you're not using them, are you? If you remove this file, tests are not passing?

jloehel: Do you mean the whole file or the line 2? If I remove the line, the test will fail with the following error message:

NameError:
       undefined local variable or method `user' for #<RepositoryPolicy:0x00000008f9b660>

mssola: Ok, ignore this then :)

mssola:
LGTM

spec/controllers/repositories_controller_spec.rb
```diff
@@ -12,9 +16,8 @@
   describe "GET #index" do
 
     it "assigns all repositories as @repositories" do
-      repository = create(:repository)
       get :index, {}, valid_session
-      expect(assigns(:repositories)).to eq([repository])
+      expect(assigns(:repositories)).to eq([visible_repository])
```
flavio: There's a hidden problem here: `let` is lazy evaluated, meaning `invisible_repository` might not even exist at this point in time. Please switch to `let!`.

[This](https://www.relishapp.com/rspec/rspec-core/v/2-6/docs/helper-methods/let-and-let) section of Rspec's docs explains the behaviour pretty well

mssola:
I'll merge this and fix that `let` statement myself, since we need this bug fix before the second stable release.

Issue #412, opened 2015-10-09
portusctl: Probably  setup --local-registry sets wrong value for rootcertbundle:


dukas:
Opensuse 13.2
Portus-20151008124708-1.1.x86_64
Documentation used: [Installing-Portus](https://github.com/SUSE/Portus/wiki/Installing-Portus)
```bash
portusctl setup --local-registry 
```
generated config.yml
```yaml
version: 0.1
loglevel: info
storage:
  filesystem:
    rootdirectory: /var/lib/docker-registry
  delete:
    enabled: true
http:
  addr: :5000
  tls:
    certificate: /etc/registry/ssl.crt/portus.crt
    key: /srv/Portus/config/server.key

auth:
  token:
    realm: https://portus.domain.dom/v2/token
    service: portus.domain.dom:5000
    issuer: portus.domain.dom
    rootcertbundle: /etc/registry/ssl.crt/portus.crt

notifications:
  endpoints:
    - name: portus
      url: https://portus.domain.dom/v2/webhooks/events
      timeout: 500ms
      threshold: 5
      backoff: 1s
```
causes error, starting registry:
```bash
panic: unable to configure authorization (token): unable to parse token auth root certificate: asn1: structure error: tags don't match (16 vs {class:0 tag:2 length:257 isCompound:false}) {optional:false explicit:false application:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} tbsCertificate @4

goroutine 1 [running]:
github.com/docker/distribution/registry/handlers.NewApp(0x7f8a89cc2778, 0xc820238450, 0xc820192d50, 0x3, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/abuild/rpmbuild/BUILD/docker-distribution-2.1.1/.gopath/src/github.com/docker/distribution/registry/handlers/app.go:188 +0x1df9
main.main()
        /home/abuild/rpmbuild/BUILD/docker-distribution-2.1.1/cmd/registry/main.go:72 +0x6cf

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
        /usr/lib64/go/src/runtime/asm_amd64.s:1696 +0x1

goroutine 34 [syscall]:
os/signal.loop()
        /usr/lib64/go/src/os/signal/signal_unix.go:22 +0x18
created by os/signal.init.1
        /usr/lib64/go/src/os/signal/signal_unix.go:28 +0x37

[...]

PR #414, opened 2015-10-09
portusctl: use right certificate

flavio:
Fixes issue #412

Signed-off-by: Flavio Castelli <fcastelli@suse.com>


flavio:
@jordimassaguerpla this is the issue we discussed on Thu, as you can see the fix is pretty simple.

PR #428, opened 2015-10-12
Changes Dockerfile for faster iteration builds

### LINKED ISSUE #427: Change Dockerfile build order to speedup iterations
In order to get faster iterations, the Dockerfile should be altered in a way that Docker can cache most of its layers during rebuilds. Static layers should be moved up and leave down the one that changes more often.

I will submit a PR with the proposed changes.

gbisheimer:
Fixes #427

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,17 @@
 FROM rails:4.2.2
 MAINTAINER Flavio Castelli <fcastelli@suse.com>
 
+ENV COMPOSE=1
+EXPOSE 3000
+
 RUN mkdir /portus
 WORKDIR /portus
-ADD . /portus
-RUN bundle install --retry=3
 
 # install supervisord, this is required to run cronos
 RUN apt-get update && apt-get install -y supervisor
 RUN ln -s /portus/docker/crono-supervisord.conf /etc/supervisor/conf.d/crono.conf
 
-ENV COMPOSE=1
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -12,4 +12,6 @@ RUN ln -s /portus/docker/crono-supervisord.conf /etc/supervisor/conf.d/crono.con
 
 ENV COMPOSE=1
 
+ADD . /portus
+
 EXPOSE 3000
```
flavio: This can be moved up

Dockerfile
```diff
@@ -12,4 +12,6 @@ RUN ln -s /portus/docker/crono-supervisord.conf /etc/supervisor/conf.d/crono.con
 
 ENV COMPOSE=1
 
+ADD . /portus
```
flavio: Right now the docker-compose setup is for development purposes: the git checkout is always mounted into the running container. Hence we can even drop this directive.

gbisheimer: That's correct for development, but I'm planning to use docker containers to deploy to production as well. So I need to have the code inside the container.
The only thing that differs is the volume mount and the environment variables (RAILS_ENV)

flavio:
Ok, just move the `EXPOSE` directive upper. Then I'll merge the PR

flavio:
Superseded by #524 524

PR #460, opened 2015-10-19
Add check for array of email in LDAP

GrantStreetGroup:
This update handles when the defined LDAP attr has an array of emails not just a single string.

### DIFF
--- a/lib/portus/ldap.rb
+++ b/lib/portus/ldap.rb
@@ -203,7 +203,11 @@ module Portus
       if cfg["attr"].empty?
         guess_from_dn(record["dn"])
       else
-        record[cfg["attr"]] || ""
+        email = record[cfg["attr"]] || ""
+
+        # Handle a LDAP record that has more than one entry.
+        email.is_a?(Array) ? email.first : email
+
       end
     end
 
--- a/spec/lib/portus/ldap_spec.rb
+++ b/spec/lib/portus/ldap_spec.rb
@@ -332,6 +332,15 @@ describe Portus::LDAP do
       ]
     end
 
+    let(:multiple_emails) do
+      [
+        {
+          "dn"    => ["ou=users,dc=example,dc=com"],
+          "email" => ["user1@example.com", "user2@example.com"]
+        }
+      ]
+    end
+
     it "returns an empty email if disabled" do
       ge = { "enabled" => false, "attr" => "" }
       APP_CONFIG["ldap"] = { "enabled" => true, "base" => "", "guess_email" => ge }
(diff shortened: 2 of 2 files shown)


lib/portus/ldap.rb
```diff
@@ -203,7 +203,14 @@ def guess_email
       if cfg["attr"].empty?
         guess_from_dn(record["dn"])
       else
-        record[cfg["attr"]] || ""
+        email = record[cfg["attr"]] || ""
+
+        #Handle a LDAP recond that has more than one entry.
```
mssola: recond -> record

lib/portus/ldap.rb
```diff
@@ -203,7 +203,14 @@ def guess_email
...
+        if email.kind_of?(Array)
```
mssola: More rubyist:

```ruby
email.is_a?(Array) ? email.first : email
```

Thus also removing the return statement afterwards.

mssola:
Good catch! Could you also provide tests for it. Take a look at the `spec/lib/ldap_spec.rb` file at the `#guess_email` describe block. It should be as easy as modifying one of those picked emails to an array.

Also, it looks like there are some issues with rubocop, but since we already have some on master, feel free to ignore them.

Last but not least, remember to sign-off your commits.

GrantStreetGroup:
I've a newbie when it come to github etiquette..    Also I'm a ruby newbie as well , so thanks for the comments.

dohnto:
@GrantStreetGroup thank you for this feature, I was not sure if this bug is on our side or not.

spec/lib/portus/ldap_spec.rb
```diff
@@ -332,6 +332,15 @@ def load_configuration_test
       ]
     end
 
+    let(:multiple_emails) do
+      [
+        {
+          "dn"     => ["ou=users,dc=example,dc=com"],
+          "email"  => ["user1@example.com","user2@example.com"]
```
mssola: Rubocop is complaining about not adding space after the comma. So, this should be instead:

```ruby
["user1@example.com", "user2@example.com"]
```

mssola: rubocop is also complaining that you're not leaving the proper alignment for each element. This means that you should leave only one space between the "email" key and the arrow. Then the line above should be aligned to the same level.

mssola:
The are more complaints raised by rubocop, but the others can be ignored, since it's about this class getting bigger and more complex. I'll just update the rubocop config so it ignores this, while keeping in mind that this class can be improved structure-wise.

Anyways, the patch looks good to me :) Just fix those nitpicks. Oh, and by the way, squash all your commits into a single one afterwards, and sign-off this commit. Thanks!

mssola:
@GrantStreetGroup thanks!

PR #467, opened 2015-10-21
fixed: Time.now to Time.zone.now and defined a class method to Regist…

vpereira:
…ry to get always the first registry and avoid future confusion

### DIFF
--- a/app/models/registry.rb
+++ b/app/models/registry.rb
@@ -13,6 +13,14 @@ class Registry < ActiveRecord::Base
   # On create, make sure that all the needed namespaces are in place.
   after_create :create_namespaces!
 
+  # Today the data model supports many registries
+  # however Portus just supports on Registry
+  # therefore to avoid confusion, define just one way
+  # to ask for the registy
+  def self.get
+    Registry.first
+  end
+
   # Returns the global namespace owned by this registry.
   def global_namespace
     Namespace.find_by(registry: self, global: true)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -54,7 +54,7 @@ class User < ActiveRecord::Base
     Namespace.find_or_create_by!(
       team:     team,
       name:     username,
-      registry: Registry.last # TODO: fix once we handle more registries
+      registry: Registry.get # TODO: fix once we handle more registries
     )
   end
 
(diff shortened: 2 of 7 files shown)


app/models/registry.rb
```diff
@@ -13,6 +13,13 @@ class Registry < ActiveRecord::Base
   # On create, make sure that all the needed namespaces are in place.
   after_create :create_namespaces!
 
+  # today the data model supports many registries
```
mssola: today -> Today

app/models/registry.rb
```diff
@@ -13,6 +13,13 @@ class Registry < ActiveRecord::Base
...
+  # however Portus today just supports on Registry
```
mssola: however Portus just supports one Registry

app/models/registry.rb
```diff
@@ -13,6 +13,13 @@ class Registry < ActiveRecord::Base
...
+  # therefore to avoid confusion, define just one way
+  # to ask for the registy
+  def self.get
+    Registry.first
+  end
```
mssola: Leave an empty line after this

app/models/registry.rb
```diff
@@ -13,6 +13,13 @@ class Registry < ActiveRecord::Base
   # On create, make sure that all the needed namespaces are in place.
   after_create :create_namespaces!
 
+  # today the data model supports many registries
+  # however Portus today just supports on Registry
+  # therefore to avoid confusion, define just one way
+  # to ask for the registy
```
mssola: Add the following note: "TODO: fix once we handle more registries"

app/models/user.rb
```diff
@@ -54,7 +54,7 @@ def create_personal_namespace!
     Namespace.find_or_create_by!(
       team:     team,
       name:     username,
-      registry: Registry.last # TODO: fix once we handle more registries
+      registry: Registry.get
```
mssola: Please, don't remove the comment. Even if it's in the documentation of the `Registry.get` method, this might be useful when refactoring into having multiple registries.

mssola:
/me can be picky sometimes :) Just fix these small nitpicks and I'll merge it.

app/models/registry.rb
vpereira: after my method or after the comment?

app/models/user.rb
vpereira: do you want to add the same comment to all places where you were using Registry.last/first?

app/models/registry.rb
mssola: after the method

app/models/user.rb
mssola: that would be even better. Thanks :)

PR #475, opened 2015-10-22
Add sleep time before configuring db first time

depay:
Portus will offen fail connecting database, if operating immediately
after mariadb container starting. So it's better adding some seconds
for waiting.

Signed-off-by: depay <depay19@163.com>

### DIFF
--- a/compose-setup.sh
+++ b/compose-setup.sh
@@ -7,22 +7,25 @@ setup_database() {
 
   TIMEOUT=90
   COUNT=0
-  printf "Portus: configuring database..."
-  docker-compose run --rm web rake db:create db:migrate db:seed &> /dev/null
+  RETRY=1
 
-  while [ $? -ne 0 ]; do
-    printf " [FAIL]\n"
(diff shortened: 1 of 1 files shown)


mssola:
It's known that Portus will fail at connecting to the database right away, that's why we use that loop afterwards: if the connection fails, sleep for some time and try again. This loop will take 90 seconds in the worst case scenario, which most probably means that something wrong occurred on the DB side.

But maybe I'm missing something.

depay:
i know that, just giving a sleep before first try...something change like try->sleep->try to sleep->try->sleep, for the first try always fails..

mssola:
If this is the case, maybe you can refactor that loop so first it sleeps and then it tries to connect. Therefore, remove the docker-compose command previous to the loop, and the loop will: sleep, try the connect, and then, if it fails, print the error message and try again.

depay:
good idea, i've updated the commit.

mssola:
Thanks!

PR #477, opened 2015-10-22
Ui minor changes and responsiveness of the website improved. 

cyntss:
Plus: on android (iphone not tested but should have a similar behaviour), adding the website to the home will give an effect of having installed a native app removing the URL bar on top and also has an app icon.


mssola:
LGTM :+1: Some notes:

- Maybe the placeholder of the search element should say "Search repository" (instead of "repo").
- The size of the icon on notifications is incosistent. See the uploaded images.

logos1
logos2

mssola:
Don't worry about the test failures, I'll fix them after we merge this ;)

cyntss:
@mssola fixed ;)

PR #485, opened 2015-10-25
Fix bug --- fixes not passed as a argument

farmerworking:


mssola:
@farmerworking Thanks

PR #491, opened 2015-10-26
Gh pages documentation

cyntss:
Based on Mmsola's work I made the new documentation page.

Someone needs to check if the images included in the documents form the wiki still work. It didnt work locally for me but it didn't make much sense to me since the src of images points to the absolute url in github.

Unfortunately, I cannot invest more time today in Portus because Im already in Manager's sprint, but I needed to finish with this before giving closure to the Portus Sprint.

Something else, remember to use Gulp if you make changes in the .less, js or new images so it takes care of compressing and uglifying things for us.


flavio:
@cyntss thanks for you help, we really appreciate it.

I didn't have time to look through the whole PR. However I noticed you didn't write how to use `gulp` inside of the README.md file. I think it would be really useful to have this last piece of information documented.

mssola:
Do not target the `master` branch ! Rather, merge onto the `gh-pages` branch. Moreover, there are some pages that should not be in here, because they are for developers only. Let me try this locally :)

mssola:
Generally speaking it looks really good :+1:, but there are tons of stuff to be done on top of this (typos, bad pages, pages that shouldn't be there, etc.). I'll do that myself. From the design point of view, I'd really appreciate if you could change the following:

Please, use the same top bar in all the layouts. Regardless if you are in the main page, or in any of the posts, I'd like to see the same top navigation bar. This gives a better consistency between pages, and it gives a uniform navigation structure. However, I really like this new section on the left side. So, my proposal: keep the top navigation bar for all the pages, and keep the left side section with just the list of posts.

Besides this main pain point, I can do the rest :)

However, I don't want this branch to be merged into `gh-pages` directly, since this branch is deployed automatically, and I don't want to expose a WIP thing to the public. Therefore, fix what I said, and I'll pick your branch from there. Once it's complete, I'll submit the PR myself.

mssola:
@flavio note that the notice on Gulp has been added to the README. Everything else LGTM. Awesome job :clap:. Now I'm taking over this branch, polishing the contents and hopefully in a couple of days I'll have something presentable :)

flavio:
Thanks a lot to both of you! :heart:

PR #496, opened 2015-10-27
review gem versions to match those in SLE-12

jordimassaguerpla:
In order to have Portus supported, we need to use rubygems that are part
of SLE-12

I run

bundle exec rspec spec

and all tests are green

If we accept this PR, despite the fact that we have to downgrade certain gems, we would only have to package these 2 rubygems:

rubygem-i18n ~> 0.7
rubygem-mime-types ~> 1.16


flavio:
I'm fine with it. I expect all these gems, despite being older, not be affected by security issues; otherwise they would not be part of SLE12. Am I right?

jordimassaguerpla:
@flavio you are right

jordimassaguerpla:
@mssola : any objection?

mssola:
LGTM

PR #526, opened 2015-10-30
refa: use rake db:setup

Salzig:
```
rake db:setup # Create the database, load the schema, and initialize with the seed data (use db:reset to also drop the database first)
```


mssola:
Thanks!

Issue #453, opened 2015-10-19
Provide default descriptions for global and personal namespaces


flavio:
These namespaces are created automatically by Portus, we should provide some meaningful description in there.

PR #529, opened 2015-11-02
Added default descriptions for global and personal namespaces

jloehel:
Fixes: #453

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -37,3 +37,18 @@ section.sign-up {
     color: $white
   }
 }
+
+.edit_team, .edit_namespace {
+  textarea {
+    max-height: 600px;
+    min-height: initial;
+    max-width: 100%;
+  }
+}
+
+#new-namespace-form, #new-team-form {
+  textarea {
+    max-height: 600px;
+    min-height: initial;
+  }
+}
--- a/app/models/registry.rb
+++ b/app/models/registry.rb
@@ -131,11 +131,12 @@ class Registry < ActiveRecord::Base
       owners: User.where(admin: true),
       hidden: true)
     Namespace.create!(
-      name:     "portus_global_namespace_#{count}",
-      registry: self,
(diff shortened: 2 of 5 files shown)


mssola:
The design of the form in the namespaces page is broken with your patch applied :/

app/assets/stylesheets/forms.scss
```diff
@@ -37,3 +37,16 @@ section.sign-up {
     color: $white
   }
 }
+
+.edit_team, .edit_namespace {
+  textarea {
+    max-height: 600px;
+    max-width: 50%;
```
mssola: Why 50% ? I'd expect the maximum to be the container itself, not 50% of it.

app/models/registry.rb
```diff
@@ -130,12 +130,12 @@ def create_namespaces!
...
-      registry: self,
-      public:   true,
-      global:   true,
-      team:     team)
+    Namespace.create!(name:        "portus_global_namespace_#{count}",
+                      registry:    self,
+                      public:      true,
+                      global:      true,
+                      description: "The global namespace for the registry, #{Registry.name}.",
+                      team:        team)
```
mssola: Why all the extra spacing? Can't you do it in the same way as you did in the `app/models/user.rb` file?

app/models/registry.rb
```diff
@@ -130,12 +130,12 @@ def create_namespaces!
...
-      name:     "portus_global_namespace_#{count}",
-      registry: self,
-      public:   true,
-      global:   true,
-      team:     team)
+    Namespace.create!(name:        "portus_global_namespace_#{count}",
+                      registry:    self,
+                      public:      true,
+                      global:      true,
+                      description: "The global namespace for the registry, #{Registry.name}.",
```
mssola: Remove the comma.

mssola:
Now everything looks fine. However, while we are at it, could you add a min-height rule for these textareas ? Preferably, the min-height of them should be the same as the initial height.

PR #536, opened 2015-11-05
Added the possibility to rename the team.

jloehel:
Screenshot:
screenshot - 11052015 - 03 00 37 pm


mssola:
Cool :clap: Some comments:

Take a look at the generated activity:

lala2

This is not *that* useful, isn't it? It would be better to say something like: "mssola renamed the team from asd to new-name". Basically:

- It's weird to say that you've edited a name: you rename ("changed the name" could work too).
- You should provide what was the previous name, and what's the current name. The link should be the new name. Note that this can be done through activity parameters.

Second of all, the `ChangeName` and the `ChangeDescription` concerns can be merged. They're almost identical. What I would do is:

1. Create a concern named ChangeNameDescription.
2. This concern would have a public method: `change_name_description(object, symbol)`.
3. Depending on the changed column, it would go to change either the name or description. Note also that this is only important for the created activity. You could update both the name *and* the description when performing the `update` (if the user hasn't changed the name, well, it will update to the same thing :P).

On the implementation of this method, I wouldn't be shy with helper protected methods. That is, take advantage of the similarities between both concerns. Note that you can make this work even for models that do not have an editable name (e.g. Namespace). To be more certain, I'd write a new test like this:

1. Check that it works for a mock model with just an editable description.
2. Check that it works for a mock model with just an editable description.

You don't have to be too exhaustive on this test, but it would be a nice addition.

On the form itself, I'd make the `name` input "required". This is because a team always has to have a name. Note that this would also work even if you just modify the description.

app/views/teams/show.html.slim
```diff
@@ -1,42 +1,56 @@
...
-      = form_for @team, remote: true, html: {role: 'form'} do |f|
-        = f.text_area(:description, class: 'form-control', placeholder: html_escape(@team.description))
-        br
+.panel-group
+  .collapse id="update_team_#{@team.id}"
+    = form_for @team, remote: true, html: {role: 'form'} do |f|
+      .panel.panel-default
+        .panel-heading
+          .input-group
+            = f.text_field(:name, class: 'form-control', placeholer: html_escape(@team.name), required: true)
```
mssola: Add `tabindex` to 1.

app/views/teams/show.html.slim
```diff
@@ -1,42 +1,56 @@
...
+            .input-group-btn
+              button.btn.btn-link.btn-xs.btn-edit-role[
+                value="#{@team.id}"
+                type="button"
+                class="button_edit_team"]
+                i.fa.fa-close
+                | Edit team
+        .panel-body
+          = f.text_area(:description, class: 'form-control', placeholder: html_escape(@team.description))
```
mssola: Add `tabindex` to 2. It's a bit of a nitpicking from my side (sorry :P), but this is one of those small additions that will make it friendlier for guys that make a heavy usage of the keyboard like me :)

app/controllers/concerns/change_name_description.rb
```diff
@@ -0,0 +1,34 @@
+module ChangeNameDescription
```
mssola: Add documentation please.

app/controllers/concerns/change_name_description.rb
```diff
@@ -0,0 +1,34 @@
...
+  extend ActiveSupport::Concern
+
+  def change_name_description(object, symbol)
```
mssola: Add documentation, stating the type, purpose of the given arguments.

app/controllers/concerns/change_name_description.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    return if old_name == new_name["name"] || new_name.empty?
+    object.update(new_name)
+    create_activity(object, symbol, old_name, new_name)
+  end
+
+  private
+
+  def create_activity(object, symbol, old, new)
```
mssola: Documentation, while stating what are these parameters.

app/controllers/concerns/change_name_description.rb
```diff
@@ -0,0 +1,34 @@
...
+    ctx, new = new.first
+    object.create_activity "change_#{symbol}_#{ctx}".to_sym,
+                                  owner:      current_user,
+                                  recipient:  object,
+                                  parameters: {
```
mssola: Can't the values of this hash be put on the same line ?

spec/controllers/teams_controller_spec.rb
```diff
@@ -142,15 +144,31 @@
...
-      expect(team_description_activity.parameters[:old_description]).to eq(old_description)
-      expect(team_description_activity.parameters[:new_description]).to eq("new description")
+      expect(team_description_activity.parameters[:old]).to eq(old_description)
+      expect(team_description_activity.parameters[:new]).to eq("new description")
+    end
+
+    it "editing of the team name" do
+      old_name = team.name
+      expect do
+        patch :update, id: team.id, team: { name:        "new name",
```
mssola: This looks kind of weird. What I'd do:

```ruby
team = { name: "new name", description: team.description }
patch :update, id: team.id, team: team, format: "js"
```

Yes, they're two lines too, but they are more readable I think.

app/views/teams/show.html.slim
jloehel: No problem. Better than too loose. I glad about every review. :-)

mssola:
Just fix these small issues and it will be good to go :)

mssola:
Could you rebase your commit? Right now it has conflicts.

PR #539, opened 2015-11-07
refa: improve create_and_update!-Repository readability

Salzig:
found the implementation a bit hard to read.


mssola:
Agreed, now it looks much better. Thanks :)

Issue #283, opened 2015-08-23
Is it possible to restrict account creation?


deitch:
Is it possible to restrict account creation, so that not everyone can set up an account? As I understand it now, anyone who can access the Portus web UI can set up their own account and then push/pull images. Is it possible to prevent that?

vitorarins:
I am running the Portus app inside a Local Network, so that is not a problem right now for me. But I am thinking of deploying Portus in our Cloud servers, and I will need to set up a VPN access only to the server in order to circumvent that problem.
It would be nice if there was an e-mail configuration, so that every user that tries to set up an account has to be approved by the admin or something like that.

deitch:
I think a request-approval process is as good as an admin-creation process. Either works.

mssola:
Hi @deitch and @vitorarins , this is some of a duplicate of the issue #179. As I said there, I'd like to bring LDAP support first before doing anything on this regard. Therefore, after LDAP support is implemented, I will take a look at this.

vitorarins:
Hi @mssola , sorry I missed that issue before commenting on this one. In my particular case I wouldn't use LDAP here in the company, but surely is a great feature to have. Thank you for the great app BTW.

mssola:
Fixed, see the issue #179.

Issue #179, opened 2015-06-24
Disable registration/users


DirkHeinke:
It would be nice to have an option to disable the registration for users and/or to set new users to default inactive. This will give the admin control about who uses the registry.

mssola:
This might make sense for large organizations, but for small organizations this can be a bit of a burden. So I feel like it would be a good idea if it's well implemented.

I'd make it a configurable option in the `config.yml` file: `verify_users` or something like that, and it would be false by default. Big organizations could set it to true on the installation process. As of disabling registration or default inactive, I'd go for a simpler approach. Let's imagine that `verify_users` is set to true, then:

1. On the "Sign up" page, instead of registering the user, it would redirect him to a page telling the user "Your petition is being processed by the administrators", or something like that.
2. The admins would receive an email, notifying them about the petition. In this email there would be a link to the "Pending users".
3. On the "Pending users" page, there would be a list of users that have filled the "Sign up" form and that the admins haven't accepted/rejected yet.
4. On accept: the user gets an email and gets registered.
5. (Maybe) On rejected: the user is added to a list of people that have tried to register, so this user doesn't try it again.

This feature wouldn't make it on the first release most probably, but a bit of discussion won't hurt :) So, @flavio, what do you think about this?

flavio:
I agree with you. Let's see what @eotchi thinks about it.

rtrauntvein:
As a suggestion, this could be essentially achieved with LDAP support #150 -

If LDAP is configured, essentially make the Portus user store read-only. Permissions and access is then based off of LDAP group membership. The jabber chat server project openfire handles LDAP and registration toggle in this way.

The large organizations out there that would want to disable registration, are most likely going to have LDAP in place.  Then you don't need to really re-engineer the sign up process or have a approvals queue.

mssola:
In my first comment I said that this configurable option would only be available through the `config.yml` file. This is too strict, maybe. I'd say to make the `config.yml` option available but that the admins could have an interface to change this anyways. This would mean that admins that are really sure about this feature can set it up when installing Portus, but they could still change it afterwards without restarting the application.

mssola:
@rtrauntvein fair point. This is why I want to think more thoroughly on LDAP support before implementing this feature. Thanks!

kenni-shin:
fast solution can be delete line
`.text-center = link_to 'or, Create a new account', new_user_registration_url`
in file /srv/Portus/app/views/devise/sessions/new.html.slim

[...]

PR #543, opened 2015-11-08
feat: simple user creation form

Salzig:
feat: simple user creation form

first steps for #283 and #179

### DIFF
--- a/app/views/admin/users/new.html.slim
+++ b/app/views/admin/users/new.html.slim
@@ -1,0 +1,26 @@
+.panel.panel-default
+  .panel-heading
+    h5 New User
+  .panel-body
+    = form_for [:admin, @user], html: {class: 'form-horizontal', role: 'form'} do |f|
+      .form-group
+        = f.label :username, {class: 'control-label col-md-2'}
+        .col-md-7
+          = f.text_field(:username, class: 'form-control', required: true, autofocus: true)
+      .form-group
+        = f.label :email, {class: 'control-label col-md-2'}
+        .col-md-7
+          = f.email_field(:email, class: 'form-control', required: true)
+      .form-group
+        = f.label :password, {class: 'control-label col-md-2'}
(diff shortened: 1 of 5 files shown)


app/views/admin/users/new.html.slim
```diff
@@ -0,0 +1,35 @@
...
+
+    = form_for [:admin, @user], html: {class: 'form-horizontal', role: 'form'} do |f|
+      .form-group
+        = f.label :username, {class: 'control-label col-md-2'}
+        .col-md-7
+          = f.text_field(:username, class: 'form-control', required: true, autofocus: true)
+      .form-group
+        = f.label :email, {class: 'control-label col-md-2'}
+        .col-md-7
+          = f.text_field(:email, class: 'form-control', required: true)
```
mssola: use `f.email_field` instead.

mssola:
Awesome :) !

I have one issue though. When the admin makes a mistake, it gets displayed like this:

form

In order to get it properly, you have to change in the line 20 of the `app/controllers/admin/users_controller.rb` file:

```ruby
render "new"
```

to:

```ruby
edirect_to new_admin_user_path, alert: @user.errors.full_messages
```

Then it will look like this:

form2

Besides that, everything looks ok to me. I'll update the related issue with a checklist of the steps to get this feature implemented. Thanks for stepping in!

app/views/admin/users/new.html.slim
Salzig: good catch, will change that later.

Salzig:
bildschirmfoto 2015-11-09 um 23 46 19

mssola:
LGTM. Could you squash it into a single commit?

Salzig:
Sure, will do that later.

Salzig:
done.

mssola:
@Salzig thanks!

Issue #451, opened 2015-10-19
Allow usage of markdown for team/namespace description


flavio:
It would be nice to have this feature.

lsamayoa:
https://richonrails.com/articles/rendering-markdown-with-redcarpet

mssola:
This issue cannot be marked as closed, because I think that this should also apply to the description of namespaces.

PR #546, opened 2015-11-09
Added the usage of markdown for the team and namespace description

jloehel:
Fixes: #451

### DIFF
--- a/app/views/namespaces/show.html.slim
+++ b/app/views/namespaces/show.html.slim
@@ -26,10 +26,10 @@
         p
           'No description has been posted yet.
       - else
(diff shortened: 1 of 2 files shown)


app/views/namespaces/show.html.slim
```diff
@@ -26,10 +26,10 @@
         p
           'No description has been posted yet.
       - else
-        '#{@namespace.description}
+        = markdown(@team.description)
```
mssola: `@namespace`, no?

jloehel: Of course, I will change it to @namespace. Sorry.

Issue #39, opened 2015-05-06
Use  typeahead  when adding users to a team


eotchi:
Improve the usability of adding a user to a team. 
Typeahead or similar (*) should be introduced so to suggest a list of valid users matching the current input of the user.

(*) https://twitter.github.io/typeahead.js/examples/

flavio:
On issue #103 @mssola proposed the usage of [chosen](http://harvesthq.github.io/chosen/).

What do you (@mssola) think about typeahead?

mssola:
@flavio I have no preference between any of the proposed solutions. `Chosen` was the one that @cyntss suggested me. So... I don't know :D

flavio:
I'll let you and @cyntss figure out what to use. I have no preference either.

brycereitano:
I was thinking about his feature, but should a non-admin be able to see a list of users? The type-ahead would expose that.

flavio:
Yes that's true, but I don't see that as a big security issue. Do you? Sites like github are doing the same thing; also organizations have already shared addressbooks containing this information...

brycereitano:
I do not, but I thought I would bring it up.

Would you mind if I pick up this issue?

flavio:
Sure, go ahead! :)

mssola:
Note that this also should apply to team names in the "Create namespace" form inside of the namespaces page.

PR #547, opened 2015-11-09
Using typeahead when adding users to a team

jloehel:
Fixes: #39

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,3 +1,8 @@
+// Typeahead
+/*
+*= require bootstrap-typeahead-rails
+*/
+
 // Vendored fonts and styles.
 @import "font-awesome";
 @import "pacifico";
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,7 +9,7 @@ class ApplicationController < ActionController::Base
   include Pundit
   rescue_from Pundit::NotAuthorizedError, with: :deny_access
 
-  respond_to :html
+  respond_to :html, :json
 
   # Two things can happen when signing in.
   #   1. The current user has no email: this happens on LDAP registration. In
@@ -83,6 +83,9 @@ class ApplicationController < ActionController::Base
   # Render the 401 page.
   def deny_access
     @status = 401
-    render template: "errors/401", status: 401, layout: "errors"
+    respond_to do |format|
+      format.html { render template: "errors/401", status: @status, layout: "errors" }
+      format.all { render nothing: true, status: @status }
+    end
   end
 end
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -1,7 +1,7 @@
 class TeamsController < ApplicationController
   include ChangeDescription
 
-  before_action :set_team, only: [:show, :update]
+  before_action :set_team, only: [:show, :update, :typeahead]
   after_action :verify_policy_scoped, only: :index
   respond_to :js, :html
 
@@ -37,6 +37,17 @@ class TeamsController < ApplicationController
   # PATCH/PUT /teams/1.json
   def update
     change_description(@team, :team)
+  end
+
+  # GET /teams/1/typeahead/%QUERY
+  def typeahead
+    authorize @team
+    @query = params[:query]
+    matches = User.search_from_query(@team.member_ids, "#{@query}%").pluck(:username)
+    matches = matches.map { |user| { username: user } }
+    respond_to do |format|
+      format.json { render json: matches.to_json }
+    end
   end
 
   private
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -107,6 +107,11 @@ class User < ActiveRecord::Base
     "Sorry, this account has been disabled."
   end
 
+  # Returns all users who match the query.
+  def self.search_from_query(members, query)
+    enabled.where.not(id: members).where(arel_table[:username].matches(query))
+  end
+
   protected
 
   # Returns whether the given user can be disabled or not. The following rules
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,10 @@
 Rails.application.routes.draw do
   resources :errors, only: [:show]
-  resources :teams, only: [:index, :show, :create, :update]
+  resources :teams, only: [:index, :show, :create, :update] do
+    member do
+      get "typeahead/:query" => "teams#typeahead", :defaults => { format: "json" }
+    end
+  end
   resources :team_users, only: [:create, :destroy, :update]
   resources :namespaces, only: [:create, :index, :show, :update] do
     put "toggle_public", on: :member
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ gem "crono"
 gem "net-ldap"
 gem "redcarpet"
 gem "font-awesome-rails"
+gem "bootstrap-typeahead-rails"
 
 # This is already a Rails dependency, but we use it to run portusctl
 gem "thor"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -56,6 +56,8 @@ GEM
     bootstrap-sass (3.3.5)
       autoprefixer-rails (>= 5.0.0.1)
       sass (>= 3.2.19)
+    bootstrap-typeahead-rails (0.10.5.1)
+      railties (>= 3.0)
     builder (3.2.2)
     byebug (5.0.0)
       columnize (= 0.9.0)
@@ -319,6 +321,7 @@ DEPENDENCIES
   awesome_print
   base32
   bootstrap-sass (~> 3.3.4)
+  bootstrap-typeahead-rails
   byebug
   capybara
   codeclimate-test-reporter
@@ -365,6 +368,3 @@ DEPENDENCIES
   webmock
   wirb
   wirble
-
-BUNDLED WITH
-   1.10.5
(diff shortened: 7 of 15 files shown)


app/controllers/teams_controller.rb
```diff
@@ -39,12 +40,28 @@ def update
     change_description(@team, :team)
   end
 
+  def typeahead
+    authorize @team
+    search = []
+    valid_users = User.enabled.where.not(id: @team.team_users.pluck(:user_id))
+    valid_users.each do |user|
+      search.push(username: user.username) if user.username.start_with?(@query)
```
Salzig: You could do this in database. 
```
team_user_ids = @team.team_users.pluck(:user_id)
User.enabled.where.not(id: team_user_ids).where(User.arel_table[:username].matches("#@query%")).pluck(:username)
```

config/routes.rb
```diff
@@ -2,6 +2,7 @@
   resources :errors, only: [:show]
   resources :teams, only: [:index, :show, :create, :update]
   resources :team_users, only: [:create, :destroy, :update]
+  get "teams/typeahead/:id/:query" => "teams#typeahead", :defaults => { format: "json" }
```
Salzig: Wouldn't it be better to handle this as member action?
```
resources :teams, only: [:index, :show, :create, :update]
  member do
    get :typeahead
  end
end
```
@flavio @mssola : any opinion?

app/controllers/teams_controller.rb
mssola: I agree with @Salzig

config/routes.rb
mssola: I agree with @Salzig

app/assets/stylesheets/application.scss
```diff
@@ -21,4 +25,3 @@
 @import 'tables';
 @import 'forms';
 @import 'types';
-@import 'responsive-utilities';
```
mssola: why ?

app/controllers/teams_controller.rb
```diff
@@ -1,7 +1,8 @@
 class TeamsController < ApplicationController
   include ChangeDescription
 
-  before_action :set_team, only: [:show, :update]
+  before_action :set_team, only: [:show, :update, :typeahead]
+  before_action :set_query, only: [:typeahead]
```
mssola: Since this is only used for the `typeahead` method, I think that you don't really need to do this in a filter.

app/controllers/teams_controller.rb
```diff
@@ -39,12 +40,28 @@ def update
     change_description(@team, :team)
   end
 
+  def typeahead
```
mssola: Documentation

app/controllers/teams_controller.rb
```diff
@@ -39,12 +40,28 @@ def update
...
+    authorize @team
+    search = []
+    valid_users = User.enabled.where.not(id: @team.team_users.pluck(:user_id))
```
mssola: Could you move this logic into the User model ? That is, I think that the User model should have a `search_from_query` method (or one with a better name :P) with this logic inside of it. Then, provide unit tests for this specific method. I'm saying this because I'm not really comfortable with queries that are not trivial in a controller.

mssola:
This branch has conflicts because I've merged #534

app/controllers/teams_controller.rb
```diff
@@ -39,6 +39,17 @@ def update
     change_description(@team, :team)
   end
 
+  # GET /teams/1/typeahead/%QUERY
+  def typeahead
+    authorize @team
+    @query = params[:query]
+    matches = User.search_from_query(@team.member_ids,
+                                     "#{@query}%").map { |user| { username: user } }
```
Salzig: IMHO: its a code smell to fetch and build objects where only a single attribute is needed.

jloehel: Bootstrap Typeahead for Rails (https://github.com/Nerian/bootstrap-typeahead-rails) uses typeahead 0.10.5.1 and for this version it is necessary to build objects. Should I switch to a newer version of typeahead? (0.11.1)

Salzig: no, i don't mean those hashes. I was thinking about those ActiveRecord Objects that get build right before you map them down to hashes. Given there are 100 results by `search_from_query` you're building 100 ActiveRecord Objects containing all information stored for those results, just to map them down to username. A huge change would be to pluck the username, so ActiveRecord is only fetching the username from database. 
```
matches = User.search_from_query(@team.member_ids,"#{@query}%").pluck(:username).map { |username| { username: username } }
```

app/models/user.rb
```diff
@@ -107,6 +107,11 @@ def inactive_message
     "Sorry, this account has been disabled."
   end
 
+  # Returns all users who match the query.
+  def self.search_from_query(members, query)
+    User.enabled.where.not(id: members).where(arel_table[:username].matches(query)).pluck(:username)
```
Salzig: btw, `search_from_query` plucks the username, wouldn't have guessed that by the name. Would be cool if it would return the ActiveRelation Object (just remove `.pluck(:username)` in the end)

app/controllers/teams_controller.rb
Salzig: sry, hadn't looked at the implementation of `search_from_query`, and hadn't expect it to return just matching usernames.

app/models/user.rb
mssola: I agree. Remove the `.pluck(:username)` here.

Salzig: maybe we could further improve on this. Extract scopes, refactor to scope.

1.
```
def self.search_from_query(members, query)
  User.enabled.where.not(id: members).where(arel_table[:username].matches(query))
end
```

2.
```
def self.search_from_query(members, query)
  enabled.where.not(id: members).where(arel_table[:username].matches(query))
end
```

3.
```
scope :search_from_query, -> (members, query) {
  enabled.where.not(id: members).where(arel_table[:username].matches(query))
}
```

4.
```
scope :name_matches, ->(query) { where(arel_table[:username].matches(query)) }
scope :search_from_query, -> (members, query) { enabled.where.not(id: members).name_matches(query) }
```

mssola: I'd honestly go for option 2. I don't like options 3 and 4 because I think that scopes should only be created if we plan to use them in multiple places.

Salzig: Option 2 and 3 are exactly the same. A Scope is just a function returning a ActiveRelation Object, it's just a different syntax and it's less obvious if you write it in the 2. way.

mssola: I agree, but I still see scopes as "handy one-liners for models" to put it in a way. I guess it's a matter of different tastes :) I don't have any strong opinions between 2 vs 3, but if I had to vote for any of these, I'd vote the second option.

Salzig: then we should at least go for 2. 3-4 are maybe a bit opinionated.

mssola: By second option I meant Option 2 from your list, so we agree :)

app/views/errors/401.json.slim
```diff
@@ -0,0 +1 @@
+'#{@status}
```
mssola: remove this file

app/controllers/application_controller.rb
```diff
@@ -83,6 +83,7 @@ def protected_controllers?(*controllers)
   # Render the 401 page.
   def deny_access
     @status = 401
+
     render template: "errors/401", status: 401, layout: "errors"
```
mssola: Use `respond_to`. For html we would use this same render, but for json:

```ruby
render nothing: true, status: status, :content_type => 'application/json'
```

app/models/user.rb
```diff
@@ -107,6 +107,11 @@ def inactive_message
     "Sorry, this account has been disabled."
   end
 
+  # Returns all users who match the query.
+  def self.search_from_query(members, query)
+    User.enabled.where.not(id: members).where(arel_table[:username].matches(query))
```
mssola: `User` is not needed.

app/views/errors/401.json.slim
jloehel: Should I add the error message directly to the the application_controller.rb?
```ruby
def deny_access
  @status = 401
  respond_to do |format|
    format.html { render template: "errors/401", status: 401, layout: "errors" }
    format.js { render js: @status }
    format.json { render json: @status }
  end
end
```

mssola: by above I meant below :P See my comment below about `render nothing: true`

mssola:
The only problem that I got, and sorry for not noticing this before, is that one side-effect of this is that you've changed the size of the input element for the name. Could you change that?

Besides that, it looks good to me.

mssola:
LGTM

flavio:
Good job @jloehel! LGMT

Salzig:
beside comment in routes.rb:9 LGTM. 

Well done!

config/routes.rb
```diff
@@ -1,7 +1,12 @@
 Rails.application.routes.draw do
   resources :errors, only: [:show]
-  resources :teams, only: [:index, :show, :create, :update]
+  resources :teams, only: [:index, :show, :create, :update] do
+    member do
+      get "typeahead/:query" => "teams#typeahead", :defaults => { format: "json" }
+    end
+  end
   resources :team_users, only: [:create, :destroy, :update]
+  # get "teams/typeahead/:id/:query" => "teams#typeahead", :defaults => { format: "json" }
```
mssola: @Salzig is right. Please remove this comment.

mssola:
Good job :clap:

flavio:
@mssola or @jloehel: could you also update the changelog?

mssola:
@flavio let me do it

PR #549, opened 2015-11-10
review gem versions to match the ones available in SUSE Enterprise Linux

jordimassaguerpla:
Server

fix bnc#953771: Release of Portus build requirements to SLE-12 codestream

I had to fix some slim issues that appeared when downgrading to 2.0.2,
as not using "," but " " to separate values in an array, and use
"if-elsif-else" clauses instead of "case-when"

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 
 gem "rails", "~> 4.2.2"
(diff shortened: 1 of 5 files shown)


flavio:
LGTM

ping @mssola

mssola:
LGTM

PR #552, opened 2015-11-11
Fix bug --- deal with upload not finished repo

farmerworking:
for these kind of repo, tags will be nil

### DIFF
--- a/app/jobs/catalog_job.rb
+++ b/app/jobs/catalog_job.rb
@@ -28,8 +28,12 @@ class CatalogJob < ActiveJob::Base
     # In this loop we will create/update all the repos from the catalog.
     # Created/updated repos will be removed from the "repos" array.
     catalog.each do |r|
-      repository = Repository.create_or_update!(r)
-      dangling_repos.delete repository.id
+      if r["tags"].blank?
+        Rails.logger.debug "skip upload not finished repo #{r["name"]}"
+      else
+        repository = Repository.create_or_update!(r)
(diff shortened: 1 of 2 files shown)


app/jobs/catalog_job.rb
```diff
@@ -28,8 +28,12 @@ def update_registry!(catalog)
     # In this loop we will create/update all the repos from the catalog.
     # Created/updated repos will be removed from the "repos" array.
     catalog.each do |r|
-      cou   = Repository.create_or_update!(r)
-      dangling_repos = dangling_repos.delete_if { |re| re == cou.id }
+      if r["tags"].nil?
```
mssola: I'd prefer if you could use `blank?`, this way we also check for an empty array.

app/jobs/catalog_job.rb
```diff
@@ -28,8 +28,12 @@ def update_registry!(catalog)
...
+        Rails.logger.debug "skip upload not finished repo #{r['name']}"
```
flavio: Can you change that to `Rails.logger.debug "skip upload not finished repo #{r["name"]}"` otherwise the code styling checks will keep failing.

mssola:
Travis is complaining that:

- You should add a test for this specific case. Just extend one of the tests in `spec/jobs/catalog_job_spec.rb`.
- rubocop is failing.

mssola:
Ok fix that rubocop complain and squash your commits into a single one. I'll merge it then ;)

Issue #560, opened 2015-11-12
Portus can't reach docker registry with "compose-setup.sh" demo-deployment


cduch:
I've followed the instructions to create a Portus demo-environment with the compose-setup.sh script in a freshly installed OpenSUSE 42.1 (with docker and docker-compose installed).

The compose-setup.sh script told me to:
```
Open http://172.17.42.1:3000 with your browser and perform the following steps:
  1 - Create an admin account
  2 - Add a new registry: choose a custom name, enter 172.17.42.1:5000 as hostname
```
I've created the account in step one.

I'm stuck at step 2 with the error:
```
Error: connection refused. The given registry is not available! You can skip this check by clicking on the "Skip remote checks" checkbox.
```
I skipped the check as advised and I was able to do a docker login with my credentials from step 1 on the cmd line. Even docker push works with the registry. But when I take a look in the Portus UI, the are no activities at all. 

The output of docker-compose ps:
```
      Name                     Command               State                       Ports                      
-----------------------------------------------------------------------------------------------------------
portus_crono_1      /usr/bin/supervisord             Up      3000/tcp                                       
portus_db_1         /docker-entrypoint.sh mysqld     Up      3306/tcp                                       
portus_registry_1   /registry/entry.sh               Up      0.0.0.0:5000->5000/tcp, 0.0.0.0:5001->5001/tcp 
portus_web_1        puma -b tcp://0.0.0.0:3000 ...   Up      0.0.0.0:3000->3000/tcp  
```
So I'm kinda stucked now.

mssola:
Sadly this is a known issue. See #441. I want to fix this situation as soon as possible, but I haven't had the time yet :( Even if this is a duplicate of #441, I'll keep this open.

cyphar:
@cduch Can you please post the version of `docker-compose` you are using? It should be noted that there's a bug in `1.5.1` which causes our configuration setup to not work. This upstream bug was fixed in `1.5.2`, and might solve your problem.

cyphar:
@cduch This should be fixed by now. Make sure:

- To use the current master or v2, since there are some fixes regarding how the DOCKER_HOST is set.
- Cleanup any previous generated files: `docker/environment` and `docker/registry/config.yml`.
- Make sure that you are using docker-compose 1.5.2. Compose 1.5.0 and 1.5.1 are known to contain a bug in which the `env_file` directive was not properly parsed (and therefore the DOCKER_HOST was not properly set in the Portus container).
[...]

PR #564, opened 2015-11-12
refa: distinguish timeout and connection refused

Salzig:
could be helpful for issues like #560


mssola:
Travis is complaining that there are some missing tests. Just go to the `spec/models/registry_spec.rb` file and add a test case in the "reachable" describe block. As you can see, you only have to add your new case in that array, and then it should be fine.

Besides that, it looks good to me. Thanks!

Salzig:
Haha, that's happening when you haven't had the patience to wait for the tests, create a pull request and go to bed anyways. 
Will fix Travis later.

Issue #553, opened 2015-11-11
Updating an already indexed image is not reflected in Portus


boyand:
Currently if you change an already uploaded image ( with tag "latest" for example ) and push the new version, the Portus UI does not update the timestamp to the last push. This coupled with https://github.com/SUSE/Portus/issues/512 makes it very hard to figure out if an image has been updated.

mssola:
I agree.

JonathonReinhart:
:+1: In addition to the timestamp, there is no entry in the "Recent activities" view.

Example: I just pushed a new version as both `:0.0.5` and as `:latest`, but "Recent activities" only shows me pushing the new tag, `:0.0.5`.

iitns:
+1

iitns:
+1

Though the timestamp it is not updated to the last push, pushed image is updated in distribution.

NickHu:
+1
Still happens on latest master (4b2f3bf).
I'm a little confused at what https://github.com/SUSE/Portus/blob/4b2f3bf63866cf4b6498f58a222c186001f1a44f/app/models/repository.rb#L143-L146 is supposed to do - the comment would imply that it addresses this, but I think I may be misunderstanding it - perhaps someone could clarify?

mssola:
@NickHu so, the thing here is that digests will get updated properly. The only thing that this is not properly updated in the `updated_at` attribute (in fact, we kind of never cared about the updated_at attribute :sweat_smile:). Moreover, a new activity should be created for that.

PR #566, opened 2015-11-12
feat: track tag recreation

Salzig:
possible fix for #553

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -50,7 +50,7 @@ class Repository < ActiveRecord::Base
 
   # Add the repository with the given `repo` name and the given `tag`. The
   # actor is guessed from the given `event`.
-  def self.add_repo(event, namespace, repo, tag)
+  def self.add_repo(event, namespace, repo, tag_name)
     actor = User.find_from_event(event)
     return if actor.nil?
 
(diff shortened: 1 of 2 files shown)


Salzig:
Will fix travis later.

mssola:
Thanks for taking care of this :)

I have one small suggestion though. If I'm not mistaken, with this we will create a new "create" activity for this tag. But this is a bit of a lie, isn't it :) Instead, in this case the activity to be created should be something like "user updated ...". What do you think?

Salzig:
I think we could find arguments for both variants (destroy&update vs. recreate/updated). I thought it would be good to destroy the old tag, and create a new one would be less effort to implement and support in the future, as can use everything as it is right now for the sake of another database record. Alternative way would be to create a new kind of activity to distinguish between pushed and "repushed/updated" tags. 
Do you prefer the later one?

flavio:
On Nov 14, 2015 8:30 PM, "Ben Rexin" <notifications@github.com> wrote:
>
> Alternative way would be to create a new kind of activity to distinguish
between pushed and "repushed/updated" tags.
> Do you prefer the later one?

That sounds better to me.

Salzig:
I've to rethink this.

mssola:
@Salzig I'd also go the later one. Thanks for taking care of this :)

JonathonReinhart:
One possible argument for keeping this as it was originally suggested:  The "Recent activities" view currently only says "<Username> pushed <tagname>"... which is true. It doesn't say that he pushed *a new* tag, only that one was pushed at all.

Issue #558, opened 2015-11-12
Use typeahead to autocomplete teams when creating a namespace


mssola:
That is, in the form that can be found in the namespaces#index page.

PR #567, opened 2015-11-13
Using typeahead to autocomplete teams when creating a namespace

jloehel:
Fixes: #558

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -44,7 +44,7 @@ class TeamsController < ApplicationController
     authorize @team
     @query = params[:query]
     matches = User.search_from_query(@team.member_ids, "#{@query}%").pluck(:username)
-    matches = matches.map { |user| { username: user } }
+    matches = matches.map { |user| { name: user } }
     respond_to do |format|
       format.json { render json: matches.to_json }
     end
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -1,6 +1,13 @@
 class Team < ActiveRecord::Base
   include PublicActivity::Common
 
+  # Returns all the teams that are not special. By special team we mean:
+  #   - It's the global namespace (see: Registry#create_namespaces!).
+  #   - It's a personal namespace (see: User#create_personal_namespace!).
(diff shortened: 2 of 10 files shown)


app/models/team_user.rb
```diff
@@ -25,6 +25,11 @@ def create_activity!(type, owner, parameters = nil)
     team.create_activity type, owner: owner, recipient: user, parameters: params
   end
 
+  # Returns all team IDs which are manageable by one user
+  def self.get_valid_team_ids(id)
+    where(user_id: id, role: 2).pluck(:team_id)
```
mssola: What's role 2 ? Is it owner ? If so, then rewrite it using the `owner` scope:

```ruby
owner.where(user_id: id).pluck(:team_id)
```

app/models/team.rb
```diff
@@ -26,4 +26,9 @@ def self.all_non_special
   def member_ids
     team_users.pluck(:user_id)
   end
+
+  # Returns all teams whose name match the query
+  def self.search_from_query(valid_teams, query)
+    where(id: valid_teams).where(hidden: 0).where(arel_table[:name].matches(query))
```
mssola: Use the `all_non_special` method for the `where(hidden: 0)` part. Moreover, could you transform the `all_non_special` method into a scope ? (while preserving all its documentation of course).

app/controllers/teams_controller.rb
```diff
@@ -44,7 +44,7 @@ def typeahead
     authorize @team
     @query = params[:query]
     matches = User.search_from_query(@team.member_ids, "#{@query}%").pluck(:username)
-    matches = matches.map { |user| { username: user } }
+    matches = matches.map { |user| { name: user } }
```
mssola: How doesn't this break the form for auto-completing user names in the "Add member" form ?

jloehel: I rewrote the javascript in teams.coffee. Now I use one function for both forms.

mssola: Ah ok, didn't notice this change :)

Issue #204, opened 2015-07-06
Add the ability to add comments on repositories


mssola:

jloehel:
Who is allowed to write/update/delete comments? Are there any specific requirements for the view?

mssola:
In my opinion, anyone who has access to the repo (so even a viewer), should be allowed to write a comment.

flavio:
Yeah, I agree with @mssola

PR #583, opened 2015-11-19
Added the the ability to add comments on repositories

jloehel:
### Screenshots:
screenshot - 11262015 - 03 24 35 pm
screenshot - 11242015 - 05 23 29 pm
screenshot - 11242015 - 05 22 50 pm

Fixes: #204

### DIFF
--- a/app/assets/javascripts/comments.coffee
+++ b/app/assets/javascripts/comments.coffee
@@ -1,0 +1,8 @@
+$(document).on "page:change", ->
+
+  # Shows and hides the comment form
+  $('#write_comment_repository_btn').unbind('click').on 'click', (event) ->
+      $('#write_comment_form').toggle 400, "swing", ->
+        if $('#write_comment_form').is(':visible')
+          $('#comment_body').focus()
+          layout_resizer()
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -27,4 +27,5 @@
 @import 'tables';
 @import 'forms';
 @import 'types';
+@import 'comments';
 @import 'responsive-utilities';
--- a/app/assets/stylesheets/comments.scss
+++ b/app/assets/stylesheets/comments.scss
@@ -1,0 +1,15 @@
+#write_comment_repository_btn, .destroy_comments_btn {
+  padding-top: 0px;
+  padding-bottom: 0px;
+  border: 0px;
+}
+
+.comment-thumbnail {
+  padding: 10px 0px;
+  img {
+    border-radius: 100%;
+    border: 4px solid $gray-light;
+    width: 40px;
+  }
+}
+
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,0 +1,33 @@
+class CommentsController < ApplicationController
+  before_action :set_repository
+  respond_to :js, :html
+
+  # POST /repositories/1/comments
+  # POST /repositories/1/comments.json
+  def create
+    @comment = @repository.comments.new(params.require(:comment).permit(:body))
+    @comment.author = current_user
+    authorize @comment
+
+    if @comment.save
+      respond_with(@comment)
+    else
+      respond_with @comment.errors, status: :unprocessable_entity
+    end
+  end
+
+  # DELETE /repositories/1/comments/1
+  # DELETE /repositories/1/comments/1.json
+  def destroy
+    @comment = @repository.comments.find(params[:id])
+    authorize @comment
+    @comment.destroy
+    respond_with @comment
+  end
+
+  private
+
+  def set_repository
+    @repository = Repository.find(params[:repository_id])
+  end
+end
--- a/app/controllers/repositories_controller.rb
+++ b/app/controllers/repositories_controller.rb
@@ -13,6 +13,7 @@ class RepositoriesController < ApplicationController
   def show
     authorize @repository
     @tags = @repository.tags.order("created_at DESC")
+    @repository_comments = @repository.comments.all
     respond_with(@repository)
   end
 
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -1,0 +1,12 @@
+class Comment < ActiveRecord::Base
+  include PublicActivity::Common
+  belongs_to :repository
+  belongs_to :author, class_name: "User", foreign_key: "user_id"
+
+  validates :body, presence: true
+
+  # Returns true if the user is the author of the comment
+  def author?(user)
+    user_id == user.id
+  end
+end
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -5,6 +5,7 @@ class Repository < ActiveRecord::Base
   belongs_to :namespace
   has_many :tags, dependent: :delete_all
   has_many :stars, dependent: :delete_all
+  has_many :comments, dependent: :delete_all
 
   # We don't validate the format because we get that from the registry, and
   # it's guaranteed to be well-formatted there.
--- a/app/views/comments/create.js.erb
+++ b/app/views/comments/create.js.erb
@@ -1,0 +1,15 @@
+<% if @comment.errors.any? %>
+  $('#alert p').html("<%= escape_javascript(@comment.errors.full_messages.join('<br/>')) %>");
+<% else %>
+  <% if @repository.comments.count == 1 %>
+    $(".comment_string").text(" Comment");
+    $(".container-full").text("")
+  <% elsif @repository.comments.count == 2 %>
+    $(".comment_string").text(" Comments");
+  <% end %>
+  $(".number_of_comments").text("<%= escape_javascript(@repository.comments.count.to_s) %>");
+  $("<%= escape_javascript(render @comment) %>").appendTo(".container-full");
+  $('#alert p').html("New comment posted");
+  $('#write_comment_form').fadeOut();
+<% end %>
+$('#alert').fadeIn();
--- a/app/views/comments/destroy.js.erb
+++ b/app/views/comments/destroy.js.erb
@@ -1,0 +1,14 @@
+<% if @comment.errors.any? %>
+  $('#alert p').html("<%= escape_javascript(@comment.errors.full_messages.join('<br/>')) %>");
+<% else %>
+  <% if @repository.comments.count == 0 %>
+    $(".comment_string").text(" Comments");
+    $(".container-full").text("Nobody has left a comment yet.");
+  <% elsif @repository.comments.count == 1 %>
+    $(".comment_string").text(" Comment");
+  <% end %>
+  $("#comment_<%= escape_javascript(@comment.id.to_s) %>").remove();
+  $(".number_of_comments").text("<%= escape_javascript(@repository.comments.count.to_s) %>");
+  $('#alert p').html("Comment successfully deleted");
+<% end %>
+$('#alert').fadeIn();
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,6 +13,7 @@ Rails.application.routes.draw do
 
   resources :repositories, only: [:index, :show] do
     post :toggle_star, on: :member
+    resources :comments, only: [:create, :destroy]
   end
 
   devise_for :users, controllers: { registrations: "auth/registrations",
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151112003047) do
+ActiveRecord::Schema.define(version: 20151124150353) do
 
   create_table "activities", force: :cascade do |t|
     t.integer  "trackable_id",   limit: 4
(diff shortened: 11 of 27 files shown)


mssola:
Good hob :clap:. Now, I'm afraid that I'm going to nitpick a bit:

- I don't like that when there are no comments, it's just a "grey" area. You can see that in your first screenshot actually. The background should be white, and there should be a text saying something on the lines of "Nobody has left a comment yet".
- I don't quite like the icon for adding a new comment. Maybe @cyntss can help us with this. Anyways, for now you can just ignore this.
- The "Write a Comment" link is actually not aligned with the "$count Comments" title, when it should. This is due the `btn` class adding a bit of extra padding. I'm not saying that you should remove this class, but maybe you could polish this. See what I'm saying in the following image:

comment

Now, I'm not so sure about the `title`. Honestly I would've preferred simple comments: just the text, that's it. This would make the footer of the comment unnecessary, so you could put that information on the header of the comment. Also, in the footer you put something like:

    mssola commented on busybox 6 minutes ago

It's redundant to say that I commented on `busybox`. I'm on the `busybox` repo already, I know where the comment is. Moreover, since it's easy to do it, I'd also add the profile picture of the user. So, in my opinion, a comment should look like:

- A small header (with little to no padding), no footer.
- The header contains the profile picture, and something like "mssola commented 6 minutes ago". "mssola" should be bold and it should be a link. Just like Github.

Another think I'm not sure about is the icon being used for the activity. I'd prefer any of the 6 icons named `comment` in font awesome :)

Functionality-wise, the rest looks good to me. Now, let's dive into the code.

app/assets/javascripts/comments.coffee
```diff
@@ -0,0 +1,14 @@
+$(document).on "page:change", ->
+
+  # Shows and hides the comment form
+  $('#write_comment_repository_btn').unbind('click').on 'click', (event) ->
+      $('#write_comment_form').toggle 400, "swing", ->
+        if $('#write_comment_form').is(':visible')
+          $('#write_comment_repository_btn i').addClass("fa-minus-circle")
+          $('#write_comment_repository_btn i').removeClass("fa-plus-circle")
```
mssola: This doesn't work. Mainly because it has the `fa-user-plus` class. Therefore, all this adding/removing of classes just doesn't work for me. Note that this would fix one of my complaints in my previous comment,

app/assets/stylesheets/comments.scss
```diff
@@ -0,0 +1,3 @@
+#comments {
```
mssola: I don't know if this would be even needed anymore if we agree on the header/footer thing I was saying in my previous comment :)

app/helpers/comments_helper.rb
```diff
@@ -0,0 +1,2 @@
+module CommentsHelper
```
mssola: Remove this

app/models/comment.rb
```diff
@@ -0,0 +1,13 @@
+class Comment < ActiveRecord::Base
+  include PublicActivity::Common
+  belongs_to :repository
+  belongs_to :author, class_name: "User", foreign_key: "user_id"
+
+  validates :title, presence: true
+  validates :body, presence: true
+
+  # Returns true if the user is the author of the comment
+  def author?(user)
```
mssola: Doesn't Rails offer a helper method because you declared `:author` as a belongs_to relationship ?

app/models/repository.rb
```diff
@@ -21,6 +22,11 @@ class Repository < ActiveRecord::Base
     # options :namespace_name, type: :fulltext
   end
 
+  # Returns the number of Comments
+  def number_of_comments
```
mssola: `comments.all.count` is the same as `comments.count`. So, remove this method altogether.

app/policies/public_activity/activity_policy.rb
```diff
@@ -37,7 +37,9 @@ def resolve
                "(team_users.user_id = ? OR namespaces.public = ?)",
                "Repository", user.id, true)
 
-      team_activities.union_all(namespace_activities).union_all(repository_activities).distinct
+      team_activities.union_all(namespace_activities)
```
mssola: If this was not an issue for rubocop, why ?

spec/policies/comment_policy_spec.rb
```diff
@@ -0,0 +1,4 @@
+require "rails_helper"
```
mssola: Remove this.

spec/models/comment_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+describe Comment do
+
```
mssola: Remove empty line

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      it "does not allow invalid parameters" do
+        sign_in owner
+        repository_id = invisible_repository.id
+        post :create, repository_id: repository_id, comment: invalid_attributes, format: :js
+        expect(assigns(:comment)).to be_a_new(Comment)
+        expect(response.status).to eq 422
+      end
+    end
+  end
+  describe "DELETE #destroy" do
```
mssola: Add an empty line before the `describe` block.

db/schema.rb
```diff
@@ -31,6 +31,18 @@
   add_index "activities", ["recipient_id", "recipient_type"], name: "index_activities_on_recipient_id_and_recipient_type", using: :btree
   add_index "activities", ["trackable_id", "trackable_type"], name: "index_activities_on_trackable_id_and_trackable_type", using: :btree
 
+  create_table "comments", force: :cascade do |t|
+    t.string   "title",         limit: 255
```
mssola: As I said in my big comment, I don't like comments having a `title`. I'd remove this.

app/views/comments/create.js.erb
```diff
@@ -0,0 +1,17 @@
...
+<% else %>
+  <% if @repository.number_of_comments == 1 %>
+    $(".comment_string").text(" Comment");
+  <% elsif @repository.number_of_comments == 2 %>
+    $(".comment_string").text(" Comments");
+  <% end %>
+  $(".number_of_comments").text("<%= escape_javascript(@repository.number_of_comments.to_s) %>");
+  $("<%= escape_javascript(render @comment) %>").appendTo(".container-full");
+  $('#alert p').html("New comment posted");
+  $('#alert').fadeIn();
```
mssola: Note that the other branch of this if-else statement is also calling `$('#alert').fadeIn();`. Maybe you can just call this after the if-else statement, so it's valid for both branches.

app/controllers/repositories_controller.rb
```diff
@@ -1,5 +1,5 @@
 class RepositoriesController < ApplicationController
-  before_action :set_repository, only: [:show, :toggle_star]
+  before_action :set_repository, only: [:show, :toggle_star, :number_of_comments]
```
mssola: which controller is defining this `:number_of_comments` filter. Because I can't see it.

app/controllers/repositories_controller.rb
```diff
@@ -13,6 +13,7 @@ def index
   def show
     authorize @repository
     @tags = @repository.tags.order("created_at DESC")
+    @repository_comments = @repository.comments.page(params[:comments_page]).per(5)
```
mssola: I'm actually not sure about paginating comments. Neither Github nor Gitlab, nor Docker Hub does it. I don't expect a *ton* of comments to be created honestly. I think that this can be simplified. If you agree, then get rid off the pagination part on the view of this method.

config/routes.rb
```diff
@@ -13,7 +13,9 @@
 
   resources :repositories, only: [:index, :show] do
     post :toggle_star, on: :member
+    resources :comments, only: [:create, :destroy]
   end
+  match "destroy_comment/:id" => "repositories#show", :via => :delete
```
mssola: I suppose that this is a leftover, right ?

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  # POST /repositories/1/comments.json
+  def create
+    @comment = @repository.comments.new(comment_params)
+    @comment.author = current_user
+    authorize @comment
+
+    if @comment.save
+      @repository.create_activity :create_comment,
+                                  owner:      current_user,
+                                  parameters: {
```
mssola: Instead of doing this, just save: `comment_id`, `comment_create`, `repo_id` and `repo_name`. These are the attributes that are useful whenever the comment or the repo has been removed.

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  # DELETE /repositories/1/comments/1
+  # DELETE /repositories/1/comments/1.json
+  def destroy
+    @comment = @repository.comments.find(params[:id])
+    authorize @comment
+
+    @repository.create_activity :destroy_comment,
+                                owner:      current_user,
+                                parameters: {
```
mssola: Same as with the `create` method.

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    respond_with @comment
+  end
+
+  private
+
+  def set_repository
+    @repository = Repository.find(params[:repository_id])
+  end
+
+  def comment_params
```
mssola: If you remove the `:title` attribute, this method won't be necessary anymore I think.

app/views/comments/destroy.js.erb
```diff
@@ -0,0 +1,14 @@
...
+<% else %>
+  <% if @repository.number_of_comments == 0 %>
+    $(".comment_string").text(" Comments");
+  <% elsif @repository.number_of_comments == 1 %>
+    $(".comment_string").text(" Comment");
+  <% end %>
+  $("#comment_<%= escape_javascript(@comment.id.to_s) %>").remove();
+  $(".number_of_comments").text("<%= escape_javascript(@repository.number_of_comments.to_s) %>");
+  $('#alert p').html("Comment successfully deleted");
+  $('#alert').fadeIn();
```
mssola: The same as for the `create.js.erb` file

app/controllers/comments_controller.rb
mssola: Maybe the body is needed, so it's saved on the CSV. That being said, the title is not needed imho

mssola:
Besides all my comments, in the activities you still have the problem that they are going to crash if the repo containing the added/removed comment gets removed. The way to go is:

- Check whether the resource (trackable) actually exists.
  - If it does: use it.
  - Otherwise use what's been stored in the `parameters` attribute.

app/controllers/comments_controller.rb
jloehel: Why do we need the repo_id and the repo_name in parameters? The id and name of the repo are reachable via activity.trackable.

app/controllers/repositories_controller.rb
jloehel: Yes, it's a leftover. I will remove it.

app/models/comment.rb
jloehel: No, tried it without this method and it doesn't work. Is it necessary require or include this helper method?

app/controllers/comments_controller.rb
mssola: not if the activity.trackable is no longer there ;) That is, even if a repo gets removed (by, for example, the catalog job), the activity is still reachable.

app/models/comment.rb
mssola: Ignore this then

app/policies/public_activity/activity_policy.rb
jloehel: I added a policy for comment-activities and after that the line was too long. Later I forgot to restore the orginally line.

app/controllers/comments_controller.rb
jloehel: I thought only tags are removable, but now it makes sense.

flavio: Do we really need to track comments? I fear they would just clutter the audit log making harder to find the relevant messages.

app/controllers/comments_controller.rb
flavio: Same consideration here.

flavio:
Good job @jloehel :+1: 

Also thanks to @mssola for the code review!

app/controllers/comments_controller.rb
mssola: I don't really have a strong opinion here. Also, we haven't defined really which kinds of actions deserve an activity :)

So, even if @jloehel is going to kill me for saying this :D, let's do as @flavio says. This will simplify this PR a *lot* :)

app/models/comment.rb
```diff
@@ -0,0 +1,12 @@
+class Comment < ActiveRecord::Base
+  include PublicActivity::Common
+  belongs_to :repository
+  belongs_to :author, class_name: "User", foreign_key: "user_id"
+
+  validates :body, presence: true
+
+  # Returns true if the user is the author of the comment
+  def author?(user)
+    author.id == user.id
```
Salzig: i would say it is better to compare using the `user_id` field, instead of building the author object which is maybe not needed.

mssola: I agree

app/assets/stylesheets/application.scss
```diff
@@ -27,4 +27,5 @@
 @import 'tables';
 @import 'forms';
 @import 'types';
+@import 'comments';
```
mssola: Remove this since the imported file is empty. And remove the ` app/assets/stylesheets/comments.scss` file as well.

spec/policies/comment_policy_spec.rb
```diff
@@ -0,0 +1,2 @@
+describe CommentPolicy do
```
mssola: Remove this

PR #588, opened 2015-11-20
latest tag first

farmerworking:


mssola:
I'd personally expect the list to be sorted by name, but I don't really have a strong opinion here.

@flavio what do you think ?

flavio:
I can see why the "latest" tag should be placed on top of the page.

However I would propose something like that:
  * latest tag is shown on top
  * remaining tags are sorted by name

What do you think?

mssola:
@flavio: I was confused by that too, but the proposal is put the newest tags on top, not that the special label "latest" should be put on top :)

I agree with your proposal. Makes sense. @farmerworking what do you think ?

farmerworking:
What I said may be a little confusing.

`latest` means the most recently created tag, not the latest tag itself.

The most recently created tag is most likely to be used so that I would like to see it on top otherwise once a repository has too many tags, the tag I would like to use will be very low in the page or in the last page if pagination is used

mssola:
Thinking about it, I think that this is a good default. Note that the `latest` tag is usually the latest pushed :) Therefore, this should give the "right" result most of the times. Moreover, I'd also add the possibility to order by any column on the frontend side. This shouldn't be too hard, and it would already fix any other situations.

@flavio what do you think ?

flavio:
Yeah, I'm fine with it now. LGTM

Issue #458, opened 2015-10-19
portusctl: use sha256 to generate certificates


flavio:
Right now it's using sha1, which is not secure enough.

jloehel:
Is it ok, if I modify the bash-script gensslcert temporarily and add the option "default_md  = sha256"?

flavio:
I'm not an expert of the tool we use to generate the certs, but if that works why not? :)

mssola:
Closing because of #1464

PR #593, opened 2015-11-23
Portusctl uses now sha256 to generate certificates

jloehel:
Fixes: #458

### DIFF
--- a/packaging/suse/portusctl/lib/configurator.rb
+++ b/packaging/suse/portusctl/lib/configurator.rb
@@ -50,9 +50,10 @@ EOM
         "-e", @options["ssl-email"],
         "-c", @options["ssl-country"],
         "-l", @options["ssl-city"],
-        "-s", @options["ssl-state"]
+        "-s", @options["ssl-state"],
+        "-x", @options["digest"]
       ]
-      Runner.exec("gensslcert", args)
+      Runner.exec("/srv/Portus/packaging/suse/portusctl/bin/gensslcert", args)
     end
 
     FileUtils.chown("wwwrun", "www", "/etc/apache2/ssl.key")
--- a/packaging/suse/portusctl/bin/gensslcert
+++ b/packaging/suse/portusctl/bin/gensslcert
@@ -1,0 +1,208 @@
+#!/bin/bash
+# Peter Poeml <apache@suse.de>
+#
+# Script to generate ssl keys for mod_ssl, without requiring user input
+# most of it is copied from mkcert.sh of the mod_ssl distribution
+#
+# XXX This is just a hack, it won't be able to do anything you want!
+#
+
+function usage
+{
+	cat <<-EOF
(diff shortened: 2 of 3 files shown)


mssola:
Maybe you should update the [Portus.spec.in](https://github.com/SUSE/Portus/blob/master/packaging/suse/Portus.spec.in) file so the RPM takes care of installing this `gensslcert` script.

/ping @jordimassaguerpla

flavio:
> Maybe you should update the Portus.spec.in file so the RPM takes care of installing this gensslcert script.

I don't think that's a good idea. The certificate is generated by portusctl AFAIK

mssola:
@flavio yes, but this commit is actually replacing the `gensslcert` script being used by `portusctl` with its own version. I was saying that the rpm should handle this. What are you saying, that there should be a way to work around this without bringing our own version of `gensslcert` ?

jloehel:
Should I call openssl directly via ruby and get rid of the bash script?

flavio:
I would prefer to keep invoking `gensslcert` because by doing that we will get security fixes. For example a couple of months ago there was a security issue that was fixed and detected inside of this script.

jloehel:
So instead of providing my own script I inject the option into the existing one and do a rollback afterwards?

jordimassaguerpla:
Sorry I lack a bit of context here. Why we are doing this?

packaging/suse/portusctl/lib/configurator.rb
```diff
@@ -50,9 +50,10 @@ def ssl
         "-e", @options["ssl-email"],
         "-c", @options["ssl-country"],
         "-l", @options["ssl-city"],
-        "-s", @options["ssl-state"]
+        "-s", @options["ssl-state"],
+        "-x", @options["digest"]
       ]
-      Runner.exec("gensslcert", args)
+      Runner.exec("/srv/Portus/packaging/suse/portusctl/bin/gensslcert", args)
```
Salzig: has this path to be absolute?

mssola: @Salzig It's `portusctl`, which is the utility installed with the RPM. In this scenario it's guaranteed that Portus is installed in  `/srv/Portus`.

@jloehel  In order to be cleaner, you should be using the `PORTUS_ROOT` constant as defined [here](https://github.com/SUSE/Portus/blob/master/packaging/suse/portusctl/lib/constants.rb).

flavio: Wow, I didn't realize we were going to vendor the genssl script. I don't like this approach because it makes things harder to maintain. If the script does not offer this possibility we should fix the script inside of the SUSE package.

mssola: Agreed, I'd feel more comfortable if this is fixed upstream.

jloehel:
I will fix this issue in upstream.

Issue #590, opened 2015-11-23
Allow sorting of tags depending on the client side


mssola:
Since PR #588, tags are sorted by the `created_at` attribute. I think that this could be improved by allowing the user to sort depending on another column by clicking on the column itself. Note that this would also allow the user to change whether to sort on ascending/descending order.

For now, the only columns allowed for this kind of sorting should be the name and the creation time. The other don't make too much sense imho.

PR #599, opened 2015-11-25
The columns of the tag-table are now sortable.

jloehel:
Screenshot:
screenshot - 11252015 - 06 07 39 pm

Fixes: #590 


mssola:
Good work :clap: Some notes:

- There is a gem for this `datatables` plugin. Use this instead of copying everything into the vendor directory.
- You should select just a couple of columns to sortable, not all of them.

flavio:
Sortable columns?! :joy:

I didn't look at the code yet, but it's awesome. We will have to extend that to the other tables of course.

mssola:
> I didn't look at the code yet, but it's awesome. We will have to extend that to the other tables of course.

Note that this was a reaction to the PR #588. But yes, we should generalize this to every column. I'll open up a new issue about this. For now, I'd prefer to keep this PR fixing the issue #588.

mssola:
@jloehel by the way, your commit message is wrong! You're fixing the issue #590, not the PR #588 :)

jloehel:
This pull-request will be replaced by #607.

PR #604, opened 2015-11-27
order by updated_at

farmerworking:
Since latest tag is often overrided which should be the first in the list

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -51,6 +51,7 @@ class Repository < ActiveRecord::Base
 
   # Add the repository with the given `repo` name and the given `tag`. The
   # actor is guessed from the given `event`.
+  # nil is returned when actor can't be found otherwise repository is returned
   def self.add_repo(event, namespace, repo, tag)
     actor = User.find_from_event(event)
     return if actor.nil?
@@ -60,13 +61,18 @@ class Repository < ActiveRecord::Base
     repository = Repository.find_by(namespace: namespace, name: repo)
     if repository.nil?
       repository = Repository.create(namespace: namespace, name: repo)
-    elsif repository.tags.exists?(name: tag)
-      return
     end
 
-    digest = event.try(:[], "target").try(:[], "digest")
-    tag = repository.tags.create(name: tag, author: actor, digest: digest)
-    repository.create_activity(:push, owner: actor, recipient: tag)
+    if repository.tags.exists?(name: tag)
+      tag = repository.tags.find_by(name: tag)
+      tag.touch
+      repository.create_activity(:override, owner: actor, recipient: tag)
+    else
+      digest = event.try(:[], "target").try(:[], "digest")
+      tag = repository.tags.create(name: tag, author: actor, digest: digest)
+      repository.create_activity(:push, owner: actor, recipient: tag)
+    end
+
     repository
   end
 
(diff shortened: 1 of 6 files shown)


flavio:
Sounds reasonable to me.

mssola:
If I'm not mistaken, tags are never updated, but created. Therefore in this case tags will have the same value for `created_at` and `updated_at`. Am I missing something ?

flavio:
What happens when an existing tag is overwritten?

mssola:
That doesn't happen I think: https://github.com/SUSE/Portus/blob/master/app/models/repository.rb#L58

flavio:
As usual @mssola is right :)

@farmerworking : sorting by `updated_at` won't help you unless you change the piece of the code pointed by @mssola

mssola:
Note that this was tried in the PR #566. So it's a bit tricky to get it right. Moreover, I don't know if @Salzig is already working on this :)

Salzig:
@mssola nope, i'm not working on it.

farmerworking:
How about this modification?

app/models/repository.rb
```diff
@@ -61,12 +61,14 @@ def self.add_repo(event, namespace, repo, tag)
     if repository.nil?
       repository = Repository.create(namespace: namespace, name: repo)
     elsif repository.tags.exists?(name: tag)
-      return
+      repository.tags.where(name: tag).each { |tag| tag.touch }
```
mssola: This would be better:
```ruby
return unless repository.tags.find_by(name: "latest").try(:touch)
```

This takes into account:

1. The method returns nil on failure.
2. The name of a tag inside of a repo is unique.

mssola:
Generally speaking, yes, this is the way to go. However note:

1. You have to update tests. They are failing with your latest commit.
2. You should add more tests for this. For example:
  - Check that the update_at is really touched.
  - Check that the new activity is created.
3. You should create views for the new activity that you've just introduced (both html and csv).

mssola:
Also, sign off your commit. Moreover, the commit message should reflect what's being implemented. With your patch the most important change is that we are tracking updates on tags. Write this on your commit message please.

app/models/repository.rb
Salzig: return isn't needed here. maybe even an improvement, because now this method returns always a repository

mssola: @Salzig well, not always. Note that when the actor could not be found, it also returns nil. I think that we should return nil when no repo has been added.

Salzig: good point. :+1:

mssola: @farmerworking to clarify this, could you add what's the returned object on the documentation of this method ?

farmerworking:
I just add all necessary things but I can't figure out why the check failed

cyphar:
@farmerworking If you press "Details" next to the failed Travis build, you can find out what the issue was. Namely, the tests failed with this:

```
The command "bundle exec rspec spec" exited with 2.
$ bundle exec rubocop -V
warning: parser/current is loading parser/ruby21, which recognizes
warning: 2.1.6-compliant syntax, but you are running 2.1.2.
0.32.0 (using Parser 2.2.2.5, running on ruby 2.1.2 x86_64-linux)
```

Salzig:
Tests are failing cause of degradation in test coverage

mssola:
LGTM

@farmerworking as @Salzig mentioned, just fix the code coverage error.

mssola:
@farmerworking this PR needs to be rebased. Could you do that ?

mssola:
Closing, but feel free to re-open once you rebased your changes :wink:

PR #605, opened 2015-11-27
All textareas have now a max/min width/height.

jloehel:
Fixes: #603

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -38,19 +38,10 @@ section.sign-up {
   }
 }
 
-.edit_team, .edit_namespace {
-  textarea {
-    max-height: 600px;
-    min-height: initial;
-    max-width: 100%;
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/forms.scss
```diff
@@ -38,18 +38,12 @@ section.sign-up {
...
-.edit_team, .edit_namespace {
-  textarea {
-    max-height: 600px;
-    min-height: initial;
-    max-width: 100%;
-  }
-}
-
-#new-namespace-form, #new-team-form {
+.edit_team, .edit_namespace, #new-namespace-form, #new-team-form, #new-comment-form {
```
mssola: Instead of doing this, why don't you create a class named like `.fixed-textarea`, and apply this class to every textarea in which we want to put these limits (in our case to every one of them). This way this would work for whenever we add a new form and we forget to add it here.

Issue #600, opened 2015-11-26
Apply the "datatables" plugin to all the tables in Portus


mssola:
The PR #599 introduced the "datatables" plugin, that allows us to make tables sortable on a specific column. As mentioned in the pull request, we should extend this to every table in Portus.

mssola:
Closing since this is no longer relevant.

PR #607, opened 2015-11-27
The columns of all the tables are now sortable.

jloehel:
### Screenshots
screenshot - 12042015 - 03 24 36 pm
screenshot - 12042015 - 03 23 50 pm

Fixes: #600
Fixes: #590

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem "redcarpet"
 gem "font-awesome-rails"
 gem "bootstrap-typeahead-rails"
 gem "rails_stdout_logging", group: [:development, :staging, :production]
+gem "jquery-datatables-rails", "~> 3.3.0"
 
 # Used to store application tokens. This is already a Rails depedency. However
 # better safe than sorry...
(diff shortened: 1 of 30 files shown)


flavio:
Wow, that looks really magical: no change inside of the controllers. Are we sure this sorting is working as expected also when pagination is involved? I fear this sorts only the visible results...

jloehel:
The sorting happens only on the client side but I can add a remote source if you want.

flavio:
I want to have a "traditional" behaviour: if the results are spread over different pages, sorting should take into consideration all the items.

I fear sorting on the client side takes into consideration only what is known by the client: the results of the page which is currently displayed. If that's the case we should implement sorting on the server side (where that IMHO should not happen).

Am I wrong?

mssola:
I agree. For the `repositories#show` page this is totally fine. That being said, it would feel really weird that for sorting a table we need to hit the server honestly.

cyphar:
Well, if we wanted to do it all client-side you'd need to do pagination on the client and send them all of the rows of the table. Which isn't that bad unless you have thousands of rows.

PR #613, opened 2015-12-04
set RUBYLIB when running bundle in portusctl

jordimassaguerpla:
We have found that under certain circumstances, portusctl fails to run
bundler.

After debugging it, we found out that on our test machine RUBYLIB was
set as

/usr/lib64/ruby/gems/2.1.0/gems/bundler-1.10.6/lib/

which was causing bundler not being able to run, since this is a
different version than the one we have in /srv/Portus/vendor/....

My guess is that this is because of rubygem-bundler being installed on
that system.

This commit sets that environment variable to our bundler directory.

fix issue#612


mssola:
LGTM

Issue #614, opened 2015-12-04
Portus shows wrong amount of teams for admin user


Evalle:
I created 3 teams in Portus, and admin is a member of each of them 

![after3](https://cloud.githubusercontent.com/assets/2839811/11590258/3f7d6612-9a8f-11e5-9598-b10eb4ea02f0.jpeg)

but on users page I can see that admin is member of 5 teams

![after2](https://cloud.githubusercontent.com/assets/2839811/11590295/7941ac14-9a8f-11e5-8f51-6f78bc6e1302.jpeg)

mssola:
Note that Portus shows the proper amount of teams, the only thing that to of them are "virtual"/"hidden". Therefore, this is not a bug but an enhancement, since it's not obvious from a user's perspective.

PR #615, opened 2015-12-04
The proper number of teams for admin/users

jloehel:
Virtual/hidden teams are no longer counted for the "number of teams"-column under admin/users.

Fixes: #614

### DIFF
--- a/app/views/admin/users/index.html.slim
+++ b/app/views/admin/users/index.html.slim
@@ -40,7 +40,7 @@
                       i.fa.fa-lg[class="fa-toggle-#{user.admin? ? 'on': 'off'}"]
 
               td= user.teams.reduce(0){ |sum, t| sum += t.namespaces.count}
-              td= user.teams.count
+              td= user.teams.all_non_special.count
               td.enabled-btn
                 - if current_user.id == user.id && @admin_count == 1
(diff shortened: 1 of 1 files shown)


app/views/admin/users/index.html.slim
```diff
@@ -40,7 +40,7 @@
                       i.fa.fa-lg[class="fa-toggle-#{user.admin? ? 'on': 'off'}"]
 
               td= user.teams.reduce(0){ |sum, t| sum += t.namespaces.count}
-              td= user.teams.count
+              td= user.teams.not_hidden.count
```
mssola: Use the `all_non_special` scope as defined by the `Team` model.

app/models/team.rb
```diff
@@ -26,6 +26,11 @@ def member_ids
     team_users.pluck(:user_id)
   end
 
+  #Returns only visible teams
+  def self.not_hidden
```
mssola: Not needed, see the `all_non_special` scope defined above.

flavio:
That was fast @jloehel :) I think the code is fine.

@mssola I think we can merge it and cherry-pick the code into the v2 branch

mssola:
@flavio done

PR #618, opened 2015-12-05
docker-compose: fix Portus when using compose

cyphar:
Fix DOCKER_HOST not being populated properly when using docker-compose,
resulting in users not being able to to use Portus with the compose
setup.

Signed-off-by: Aleksa Sarai <asarai@suse.com>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,7 +28,6 @@ registry:
     - ./docker/registry:/registry
     - /registry_data
   entrypoint: /registry/entry.sh
-  env_file: docker/environment
   links:
     - web
   ports:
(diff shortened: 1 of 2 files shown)


cyphar:
**NOTE**: this might actually be an upstream bug, because `docker-compose` doesn't seem to want to inherit an environment from `env_file` if the service is extending another service.

cyphar:
Sorry, this fix was actually fixing an upstream bug in `docker-compose==1.5.1`. The upstream bug was fixed in `1.5.2`, and thus this change doesn't fix any of the other issues with the compose setup.

PR #622, opened 2015-12-08
Introduce automated repositories

flavio:
Automated repositories are built from a Dockerfile stored inside of a
remote repository (e.g. a Git repo).

Signed-off-by: Flavio Castelli <fcastelli@suse.com>

### DIFF
--- a/app/jobs/catalog_job.rb
+++ b/app/jobs/catalog_job.rb
@@ -38,7 +38,8 @@ class CatalogJob < ActiveJob::Base
 
     # At this point, the remaining items in the "repos" array are repos that
     # exist in the DB but not in the catalog. Remove all of them.
+    # Automated repositories are never removed.
     Tag.where(repository_id: dangling_repos).find_each(&:delete_and_update!)
-    Repository.where(id: dangling_repos).destroy_all
+    Repository.where(id: dangling_repos, source_url: "").destroy_all
   end
 end
--- a/app/controllers/repositories_controller.rb
+++ b/app/controllers/repositories_controller.rb
@@ -17,6 +17,49 @@ class RepositoriesController < ApplicationController
     respond_with(@repository)
   end
 
+  # GET /repositories/new
+  def new
+    @repository = Repository.new
+    @repository.namespace = Namespace.find(params[:namespace_id]) if params[:namespace_id]
+    authorize(@repository) if @repository.namespace
+
+    scope = NamespacePolicy::Scope.new(current_user, Namespace)
+    scope.include_personal_namespace = true
+    @namespaces = scope.resolve
+    logger.debug(@namespaces.count)
+  end
+
+  # POST /repositories
+  # POST /repositories.json
+  def create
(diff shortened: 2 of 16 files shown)


flavio:
I'm a bit in a rush now. It's a public holiday in Italy and I gotta go to have fun up in the mountains with the rest of my family. I'll attach the screenshots this evening (unless you are faster with the review).

mssola:
Ok, so before looking into the code too much I have some questions:

- I thought that we wouldn't have the concept of "automated build repositories" :D What I thought is that we would provide a link in the repositories#show page, in which we would allow the user to provide the info about how to build that repo/tag from a Dockerfile. That being said, I'm not against the "automated build repositories", but I'd also like to see this other method of setting this up.
- From the previous point, there's no link to the repositories#edit page. Or at least I couldn't find it :) If you want to keep the idea that only repos that have been created with the "automated build repositories" form can be edited, then you should filter this also in the controller. Otherwise, you could just provide a link to the edit page for any repo, and then my first point would just disappear :)
- How do you handle the Catalog job ? Note that the catalog job removes dangling repos (repos with no tags in the registry, and this will happen when you create an "automated build repository" always). You could update the code so the catalog job takes into account repos with the "source_url" column set.

flavio:
> I thought that we wouldn't have the concept of "automated build repositories" :D What I thought is that we would provide a link in the repositories#show page, in which we would allow the user to provide the info about how to build that repo/tag from a Dockerfile. That being said, I'm not against the "automated build repositories", but I'd also like to see this other method of setting this up too.
> From the previous point, there's no link to the repositories#edit page. Or at least I couldn't find it :) If you want to keep the idea that only repos that have been created with the "automated build repositories" form can be edited, then you should filter this also in the controller. Otherwise, you could just provide a link to the edit page for any repo, and then my first point would just disappear :)

I forgot to add some UI files to the PR. Everything is solved with them as we discussed on IRC.

> How do you handle the Catalog job ? Note that the catalog job removes dangling repos (repos with no tags in the registry, and this will happen when you create an "automated build repository" always). You could update the code so the catalog job takes into account repos with the "source_url" column set.

I didn't, thanks for pointing that out. I updated the PR accordingly.

app/jobs/catalog_job.rb
```diff
@@ -38,7 +38,8 @@ def update_registry!(catalog)
 
     # At this point, the remaining items in the "repos" array are repos that
     # exist in the DB but not in the catalog. Remove all of them.
+    # Automated repositories are never removed.
     Tag.where(repository_id: dangling_repos).find_each(&:delete_and_update!)
-    Repository.where(id: dangling_repos).destroy_all
+    Repository.where("id IN (?) AND source_url = ?", dangling_repos, "").destroy_all
```
mssola:
```ruby
Repository.where(id: dangling_repos, source_url: "").destroy_all
```

mssola:
Besides that comment, LGTM

flavio:
Fixed, merging.

PR #623, opened 2015-12-09
Move create namespace link

flavio:
Move the "create namespace" link

The link has been moved inside of the "Namespaces you have access to" container as opposed to the "Special namespaces" one.

It makes more sense to have the link in there because the new namespaces that can be created are not going to be special ones.

## Before
before

## After

after


mssola:
LGTM. Just squash your commits and feel free to merge it

PR #625, opened 2015-12-09
Introduce application tokens

flavio:
Right now the Docker client stores the credentials in plain
format on the host. This is really bad from a security point of view,
especially for users using LDAP to authenticated.

This commit introduces the concept of "application tokens". Each user
can have at most 5 application tokens. The tokens are created in a
random secure way by Portus and are stored inside of its database after
being hashed via bcrypt.

The application tokens can be used by all the programs authenticating
against a Docker registry protected by Portus (e.g.: the docker cli
client). They cannot be used to log into Portus' web interface.

The application tokens can be revoked at any time by using a new UI.

### Token management UI
new-token

Notice a new token has just been created

### Token removal

tag-remove

After the token has been removed:
tag-removed

### Token limit reached

This user has already 5 tokens, hence he cannot add new ones:
token-disabled

### DIFF
--- a/app/controllers/api/v2/tokens_controller.rb
+++ b/app/controllers/api/v2/tokens_controller.rb
@@ -2,6 +2,21 @@
 # use in order to perform operation into the registry. This is the last step in
 # the authentication process for Portus' point of view.
 class Api::V2::TokensController < Api::BaseController
+  before_action :attempt_authentication_against_application_tokens
+
+  # Try to perform authentication using the application tokens. The password
+  # provided via HTTP basic auth is going to be checked against the application
+  # tokens a user might have created.
+  # If the user has a valid application token then the other forms of
+  # authentication (Portus' database, LDAP) are going to be skipped.
+  def attempt_authentication_against_application_tokens
+    user = authenticate_with_http_basic do |username, password|
+      user = User.find_by(username: username)
+      user if user && user.application_token_valid?(password)
+    end
+    sign_in(user, store: true) if user
+  end
+
   # Returns the token that the docker client should use in order to perform
   # operation into the private registry.
   def show
--- a/app/controllers/application_tokens_controller.rb
+++ b/app/controllers/application_tokens_controller.rb
@@ -1,0 +1,39 @@
+# ApplicationTokensController manages the creation/removal of application tokens
+class ApplicationTokensController < ApplicationController
+  respond_to :js
+
+  # POST /application_tokens
+  def create
+    @plain_token = Devise.friendly_token
+
+    @application_token = ApplicationToken.new(create_params)
+    @application_token.user = current_user
+    @application_token.token_salt = BCrypt::Engine.generate_salt
+    @application_token.token_hash = BCrypt::Engine.hash_secret(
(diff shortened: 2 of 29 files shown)


flavio:
You can start reviewing the code, everything works fine and has been tested. However I want to write some tests using capybara.

app/controllers/api/v2/tokens_controller.rb
```diff
@@ -2,6 +2,16 @@
 # use in order to perform operation into the registry. This is the last step in
 # the authentication process for Portus' point of view.
 class Api::V2::TokensController < Api::BaseController
+  before_action :check_token
+
+  def check_token
```
mssola: Please, add some documentation

app/controllers/application_tokens_controller.rb
```diff
@@ -0,0 +1,43 @@
+# ApplicationTokensController manages the creation/removal of application tokens
+class ApplicationTokensController < ApplicationController
+  respond_to :js
+
+  ## GET /application_tokens/new
```
mssola: I guess that this is a leftover ;)

app/controllers/application_tokens_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      respond_with @application_token
+    else
+      respond_with @application_token.errors, status: :unprocessable_entity
+    end
+  end
+
+  # DELETE /application_token/1
+  def destroy
+    @application_token = ApplicationToken.find(params[:id])
+    # @team_user.create_activity!(:remove_member, current_user)
```
mssola: Remove this

app/models/application_token.rb
```diff
@@ -0,0 +1,16 @@
+class ApplicationToken < ActiveRecord::Base
+  belongs_to :user
+
+  validates :application,
```
mssola: why not in the same line ?

app/models/user.rb
```diff
@@ -5,6 +5,8 @@ class User < ActiveRecord::Base
   USERNAME_CHARS  = "a-z0-9"
   USERNAME_FORMAT = /\A[#{USERNAME_CHARS}]{4,30}\Z/
 
+  APPLICATION_TOKENS_MAX = 5
```
mssola: why not in the `ApplicationToken` model ? It looks like it would be more suitable there no ?

mssola:
LGTM. Just fix all my nitpicking :P

app/models/user.rb
flavio: I added there because it's a setting that applies on a per user basis

app/controllers/application_tokens_controller.rb
flavio: Good catch, I forgot about it. I'm going to track activities with a another commit inside of this PR

flavio:
Now application token events are tracked too:

token-activities

flavio:
Everything is ready to be reviewed, I'll squash all the commits into a single one before doing the merge.

mssola:
LGTM. Once tests are passing, just squash your commits and merge it.

PR #636, opened 2015-12-16
compose: support Docker 1.9 overlay networking

cyphar:
Rather than using the (quite outdated) links system, use the new overlay
network driver added in Docker 1.9. While this support is currently (as
of Docker compose 1.5.2) experimental, it works pretty well.

Signed-off-by: Aleksa Sarai <asarai@suse.com>

The two setups (linksv1 and overlay) are completely independant,
and doesn't require any configuration generation.

### DIFF
--- a/compose-setup.sh
+++ b/compose-setup.sh
@@ -2,8 +2,16 @@
 
 set -e
 
+compose() {
+  if [ "$OVERLAY" -eq 1 ]; then
+    docker-compose --x-networking --x-network-driver overlay -f ./docker-compose.overlay.yml $@
+  else
+    docker-compose $@
+  fi
+}
+
 check_version() {
-	VER=$(docker-compose version --short)
+	VER=$(compose version --short)
 	if [ "$VER" = "1.5.1" ]; then
 		# Known to have bugs when generating configs.
 cat >&2 <<EOF
@@ -41,7 +49,7 @@ setup_database() {
     COUNT=$((COUNT+5))
 
     printf "Portus: configuring database..."
-    docker-compose run --rm web rake db:setup &> /dev/null
+    compose run --rm web rake db:setup &> /dev/null
 
     RETRY=$?
     if [ $RETRY -ne 0 ]; then
@@ -66,13 +74,14 @@ clean() {
     done
   fi
 
-  docker-compose kill
-  docker-compose rm -fv
+  compose kill
+  compose rm -fv
 }
 
 usage() {
-  echo "Usage: $0 [-f]"
+  echo "Usage: $0 [-f] [-o]"
   echo "  -f force removal of data"
+  echo "  -o use overlay networking (requires Docker >=1.9)"
 }
 
 # Force the current directory to be named "portus". It's known that other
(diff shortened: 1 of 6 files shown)


mssola:
If I'm not mistaken, in docker-compose 1.5 you can specify the name of each node. This would be cool in order to get rid of [this](https://github.com/SUSE/Portus/pull/636/files#diff-37d44ef9e52354aa0009f787d039f01cR86). That is, this way we wouldn't force users to name Portus' directory Portus :)

Besides that, I think that this would be ready to be merged if we allowed this script to accept an `--overlay` or `--experimental` flag. This way, we would default to the current behavior, but if the user really needs the experimental version, he/she can get it through a flag. This wouldn't complicate the implementation because it could be handled inside of the `compose` function.

cyphar:
@mssola Some of the configs depend on the hostnames provided by the overlay network (which are different when you're not using links), we'd have to generate the configs.

mssola:
Agh, that's right... Ok, let's forget about this for now then.

cyphar:
PTAL. :whale:

compose-setup.sh
```diff
@@ -2,8 +2,16 @@
 
 set -e
 
+compose() {
+  if [ "$OVERLAY" -eq 1 ]; then
+    docker-compose --x-networking --x-network-driver overlay -f ./docker-compose.yml.overlay $@
```
mssola: docker-compose.overlay.yml, no ?

mssola:
Since docker-compose.overlay.yml and docker-compose.yml are almost identical, maybe docker-compose.yml can inherit from docker-compose.overlay.yml and add the links ?

compose-setup.sh
cyphar: :smile_cat:

docker/registry/entry.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 
 if [ ! -e /registry/config.yml ]; then
-  sed -e "s|DOCKER_HOST|${DOCKER_HOST}|g" /registry/config.yml.template > /registry/config.yml
+  sed -e "{
+    s|DOCKER_HOST|${DOCKER_HOST}|g;
+    s|PORTUS_DB_HOST|${PORTUS_DB_HOST}|g;
```
mssola: The config.yml file is not using the `PORTUS_DB_HOST` variable. You can remove this line.

mssola:
Besides that, LGTM. I'd like to hear from @flavio before merging this though ;)

mssola:
I don't know if this makes sense (I'm no expert on docker networking :P), but I think that it would be convenient that the caller of this script can decide which network driver to use. For example, in order to have an "overlay" network, one has to configure it (e.g. with a consul node and a swarm cluster). This can be tedious for people who are just experimenting with this. For this reason, maybe users can provide an environment variable named `NETWORK_DRIVER`, and then select an alternative driver that they have configured (e.g. `bridge`).

The default value would still be `overlay`. Moreover, it would be better to try that the overlay network works before doing anything at all. If it doesn't, then we have the chance to show users a fancy message about it.

cyphar:
Closed in favor of #649.

PR #649, opened 2015-12-28
Compose cleanup

flavio:
This PR incorporates the changes made by @cyphar to support the new overlay network model (see PR https://github.com/SUSE/Portus/pull/636) plus some more fixes.

The major change is the introduction of the `-e EXTERNA_IP` flag. This flag is mandatory and is used to provide the IP/FQDN of the machine that is going to expose both the registry and the portus services. The code no longer relies on guessing this information by looking at DOCKER_HOST. In some cases this automatic guess fails leading to tons of issues.

The code works fine both when using the traditional "link" feature and the new "overlay" network. I would like someone to test the PR on a computer using something like boot2docker.


mssola:
Could you take [this comment](https://github.com/SUSE/Portus/pull/636#issuecomment-165802931) into account ?

mssola:
Moreover, could you move all the docker-compose* files into the `compose` directory ?

mssola:
While we are at it, we can address the issue #654 in this PR.

mssola:
Another issue, while we are at it, is that currently this statement fails for me:

    docker-compose run --rm web rake db:setup &> /dev/null

Maybe it's because of schema not being perfectly aligned with the DB (by dealing with multiple development branches at the same time this can happen :P). It's better to call `db:migrate:reset` and then, if that was successful, call `db:seed`.

flavio:
I made all the changes @mssola pointed out. I also changed our compose file to reference a specific release of mariadb, not only for issue #654, but because it's a good practice.

The biggest change introduced by this commit are:
   * removal of the overlay network feature
   * introduction of a disclaimer about the purposes of this deployment method

I removed all the overlay network stuff because this deployment method is designed only for development and testing; hence everything is on the same computer. The overlay network is handy only when doing a multi host deployment, which is something we want to address with a different compose setup targeted at production.

Moreover some docker-compose features (system volumes) do not work on a multi-host deployment.

compose-setup.sh
```diff
@@ -113,7 +127,21 @@ while getopts "fh" opt; do
   esac
 done
 
+cat <<EOM
+
+###########
+# WARNING #
+###########
+
+This deployment method is intended for testing/development purposes.
```
mssola: I wonder if it would be a good idea to state that we will provide a production-ready version of this on the near future also in this message.

mssola:
Maybe we could pass the following arguments to the compose executable `-p portus`. This way we would also fix another known issue (see [this](https://github.com/SUSE/Portus/blob/master/compose-setup.sh#L81). This is at least valid on docker-compose 1.5

flavio:
> Maybe we could pass the following arguments to the compose executable -p portus.

As discussed on IRC let's keep things simple. We won't add this flag.

mssola:
Besides my comment on the message, it looks good to me.

flavio:
@cyphar what do you think about that?

Issue #504, opened 2015-10-28
Hiding the registry behind Portus (Apache2) ?


kooltux:
Hi all,
First, congrats for your work on Portus !

In short: is there a procedure to proxy registry API calls through Portus and avoid to expose the registry port ?

In details:

I managed to have Portus running with apache2 and the docker registry inside an opensuse container.

What I did roughly:
* docker pull opensuse:latest, then run it (this is a minimal opensuse 13.2)
Then, inside the container:
* add repo http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_13.2
* install docker-distribution-registry-2.1.1-11.2.x86_64
* add repo http://download.opensuse.org/repositories/Virtualization:/containers:/Portus/openSUSE_13.2
* install Portus-20151027090209-1.1.x86_64
* finally run 'portusctl setup' with appropriate arguments. 

The result: I have a working Portus on port 443, and the associated registry on port 5000 (this is the default config generated by portusctl). I'm then able to use 'docker login' and 'docker push' as explained in the docs. 

For example, I'm able to run a pull on a public namespace:
```
$ docker pull myhostname.mydomain.com:5000/kooltux/busybox
```

Now, I'd like to go further and avoid to expose the registry port 5000 (that is, hide the registry behind Portus) to see my private registry as a single server running on port 443. The reason is simple: there are lot of places where accessing a service on port 5000 won't be authorized. Also, I'd like to avoid to have 2 different public addresses (and 2 different hostnames meaning 2 SSL certs...) to have both the registry and portus accessible.

So I tried to make my docker requests on Portus instead of the registry, but the following commands fail:
```
$ docker pull myhostname.mydomain.com:443/kooltux/busybox 
```
Curl fails too:
```
$ curl -k https://myhostname.mydomain.com:443/v2/
```
I searched for documentation on this topic but didn't find any. Any suggestion to this make this configuration work is welcome :)

kooltux:
I just found the controllers responsible for /v2/... calls in Portus: they're located in app/controllers/api/v2/.
So the idea to use ProxyPass in Apache is a bad one, as there's a name conflict (we can't decide if v2 refers to Portus API or Registry API). If the route is renamed, then all /v2 calls could be forwarded to the registry... 
On the other hand, with current route /v2/..., all calls to the registry API would land on Portus. So they just need to be passed to the registry by a "catch-all" controller.

thenewguy:
I am looking into this as well. I posted a question to the mailing list.

[...]

PR #687, opened 2016-01-26
Change error message on openSSL's errors

flavio:
Do not try to be smart when validating a registry connection using SSL.
Let's just provide a generic error and log the exception into the logs.

This is going to help with issues like #504


mssola:
LGTM. Merge once tests are passing.

flavio:
I fixed the rubocop issue and merged the code.

PR #688, opened 2016-01-26
Dockerimage: add phantomjs

flavio:
Add the phantomjs to the Portus Docker image. Needless to say this image
is used for testing and development purposes, it's not meant for
production usage. Hence it makes sense to have all the tools actually
needed at development time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,15 @@ WORKDIR /portus
 COPY Gemfile* ./
 RUN bundle install --retry=3
 
+# Install phantomjs, this is required for testing and development purposes
+# There are no official deb packages for it, hence we built it inside of the
+# open build service.
+RUN echo "deb http://download.opensuse.org/repositories/home:/flavio_castelli:/phantomjs/Debian_8.0/ ./" >> /etc/apt/sources.list
+RUN wget http://download.opensuse.org/repositories/home:/flavio_castelli:/phantomjs/Debian_8.0/Release.key && \
+  apt-key add Release.key && \
+  rm Release.key
+RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


mssola:
I guess that this is a needed step to run tests inside of a container, right ? If this is the case, I'd even prefer to create a specific compose file inside of the `docker` just for that. I'm saying this because phantomjs is simply not used in development mode.

flavio:
> I'm saying this because phantomjs is simply not used in development mode.

Aren't you supposed to run the whole test suite before submitting any code change to Portus? :grin:

mssola:
Fair enough.

Dockerfile
```diff
@@ -8,4 +8,15 @@ WORKDIR /portus
...
+# Install phantomjs, this is required for testing and development purposes
+# There are no official deb packages for it, hence we built it inside of the
+# open build service.
+RUN echo "deb http://download.opensuse.org/repositories/home:/flavio_castelli:/phantomjs/Debian_8.0/ ./" >> /etc/apt/sources.list
+RUN wget http://download.opensuse.org/repositories/home:/flavio_castelli:/phantomjs/Debian_8.0/Release.key && \
+  apt-key add Release.key && \
+  rm Release.key
+RUN apt-get update && \
+    apt-get install -y phantomjs && \
+    apt-get clean
```
mssola: Just a small nitpick. I've seen in jfrazelle/dockerfiles that instead of `apt-get clean` she does:

```
	&& rm -rf /var/lib/apt/lists/*
```

I don't know which is best. Moreover, I'd add `--no-install-recommends` to the install command.

flavio:
@mssola thanks for the tips, I applied all of them. Merging

PR #695, opened 2016-01-28
removed duplicate ports item from config template

bibby:
The actual ports configuration is a few lines down.


cyphar:
LGTM

mssola:
@bibby thanks!

PR #698, opened 2016-01-30
Upgrade ruby version to 2.3

kalabiyau:


kalabiyau:
Just wanted to open discussion about having master on top on new and shiny and lock down pieces to be supported via tags

mssola:
I would generally be in favor of this, but if I recall correctly we set the ruby version to what it's available on openSUSE. In Ruby's case:

- openSUSE 13.2: 2.1
- Leap: 2.1
- Tumbleweed: 2.2

That being said, I have no idea if there are any plans to release 2.3 anytime soon. Moreover, you've also upgraded the `json` package. The version that we are using is not working on ruby's head version, so we really needed to upgrade this package anyways. However, I remember that there were some problems in regards to the version.

/ping @flavio @jordimassaguerpla

kalabiyau:
@mssola I think, that being locked to o.s. ruby versions is just some form of vendor lock :) We definitely should support them, but maybe find other ways of doing so. But you are right - I shall speak to @flavio

jordimassaguerpla:
Also SLE-12 has ruby 2.1 .  Portus is supported on SLE-12, thus if we want to release new versions of Portus for SLE, we need to make sure Portus works on ruby 2.1.

Thus, I would say, for me is fine that we run tests on different versions of ruby, as long as we run it for 2.1 and make sure those pass.

Then, this is fine https://github.com/SUSE/Portus/commit/fac84fbc856ff0cf0da916d3e91fb93d84f7d892
but I wouldn't go with this one, thought: https://github.com/SUSE/Portus/commit/4ab2d9c21e9a53874180b4710b3631a7c42c288f

mssola:
Let's list what other distros are doing:

| Distro (version) | ruby version |
|---------------------|-----------------|
| Debian (stable) | 2.1.5 |
| Debian (testing and unstable) | 2.2.4 |
| Ubuntu 14.04 LTS | 1.9.3 |
| Ubuntu 15.04 | 2.1.0 |
| Ubuntu 15.10 | 2.1.5 |
| Ubuntu 16.04 (to be released) | 2.2 |
| Fedora 22 | 2.2.4 |
| Fedora 23 | 2.2.4 |
| Archlinux | 2.3 |

My sources:
- [Debian](https://packages.debian.org/search?keywords=ruby&searchon=names&suite=all&section=all)
- [Ubuntu](http://packages.ubuntu.com/search?keywords=ruby&searchon=names&suite=all&section=all)
- [Fedora](http://pkgs.fedoraproject.org/cgit/rpms/ruby.git/)
- [Archlinux](https://www.archlinux.de/?page=PackageDetails;repo=extra;arch=x86_64;pkgname=ruby)

For what I see, both Debian and Ubuntu are shipping 2.1 for their stable and current releases. Both Debian and Ubuntu are going to release 2.2 in the near future. Fedora is the only one being on 2.2 on a stable relase, and Archlinux is on the bleeding edge as always.

Note that we are requiring `2.1.2` which is a pretty old patch release. We could upgrade at least to `2.1.8`, which was released last September. That being said, in the near future we should be upgrading to 2.2 instead of 2.3 I agree that we should test for 2.3 anyways, but the `.ruby-version` file states which version of ruby we are using always for development purposes, and what are we using to test Portus both in Travis (among other versions) and internally (only this version).

My conclusions:

- We should update the `.ruby-version` file to a patch level release of `2.2` soon.
- We should modify the `.travis.yml` file to include `2.3`, but keeping the `2.2` release that we pick and `2.1`.
- We will update to `2.3` in the future. For that to happen, at least openSUSE, Debian, Ubuntu and Fedora should have `2.3` in the same status as `2.2` today.

@jordimassaguerpla @flavio what do you say ?

flavio:
I agree with @jordimassaguerpla:
> Thus, I would say, for me is fine that we run tests on different versions of ruby, as long as we run it for 2.1 and make sure those pass.

I think there's no problem in testing also this version of Ruby. However we still have to make sure everything works fine with the ruby versions shipped on SLE and openSUSE. This is going to be covered both by Travis and by our internal integration tests.

Also we should stick with gems that are available for the ruby versions shipped with openSUSE and SLE.

Would that work for you?

kalabiyau:
I will leave that decision on you guys as you know better, but I think conversation was fruitful to get different perspectives.

kalabiyau:
Sorry, misclick :)

Conan-Kudo:
@mssola Fedora 24 will have Ruby 2.3. See [this Change page](https://fedoraproject.org/wiki/Changes/Ruby_2.3) for details.

mssola:
@kalabiyau imho this PR can be merged if:

- You rollback your changes on the `.ruby-version` file.
- Add `2.1.8` and `2.2.4` too on the `.travis.yml` file. This way we test all the stable releases on travis :)

@jordimassaguerpla upgrading the `json` package to `1.8.3` is inevitable, otherwise Portus is not installable with ruby 2.3, which is a stable release.

jordimassaguerpla:
@mssola : I agree

h0tbird:
Don't forget Alpine Linux v3.3 with ruby 2.2.4

mssola:
@kalabiyau just update `json` to `1.8.3` so it works on ruby `2.3`. After that, squash your commits into one, and sign it off ;)

kalabiyau:
@mssola please take a look

mssola:
Thanks!

Issue #669, opened 2016-01-18
Show Portus version and last commit ID in WebGUI


h0tbird:
It would be nice to know which version of Portus is running.

mssola:
Thanks to the PR #707, admin users will see the version of Portus on the footer. Closing.

PR #707, opened 2016-02-03
Display the git branch/commit when possible

lonewulf:
Hi, I've seen this requested as a feature, I put together a small patch that should do the trick.

It displays the git branch/commit if the git command and the .git directory are present on the system.

Signed-off-by: Yann Lopez <l@lonewulf.net>

### DIFF
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -44,3 +44,6 @@ html
         'Created with
         i.fa.fa-heart
         'by the SUSE team
+      - if current_user.admin? && Version.value
+          p
+            'Version: #{Version.value}
--- a/config/initializers/version.rb
+++ b/config/initializers/version.rb
@@ -1,0 +1,28 @@
+require 'mkmf'
+  
+# Version module
+# Makes the app version available to the application itself
+# Needs the git executable for all git operations
+module Version
+  git = find_executable('git')
+  BRANCH = git ? `git symbolic-ref HEAD 2>/dev/null | cut -d"/" -f 3  2>/dev/null`.chomp : nil
+  COMMIT = git ? `git log --pretty=format:'%h' -n 1 2>/dev/null`.chomp : nil
+  TAG = git ? `git tag --points-at $(git rev-parse HEAD) 2>/dev/null`.chomp : nil
+  
+  # Version.value returns the app version
+  # Priority: git tag > git branch/commit > VERSION file
+  def self.value
+    if TAG.present?
+      "#{TAG}"
+    elsif COMMIT.present?
+      if BRANCH.present?
+        "#{BRANCH}@#{COMMIT}"
+      else
+        "#{COMMIT}"
+      end
+    else
+      version = Rails.root.join("VERSION")
+      File.read(version).chomp if File.exists?(version)
+    end
+  end
+end


config/initializers/git_info.rb
```diff
@@ -0,0 +1,7 @@
+module Git
+    `which git`
+    if $?.success?
+      BRANCH = `git symbolic-ref HEAD 2>/dev/null | cut -d"/" -f 3`.strip
+      COMMIT = `git log --pretty=format:'%h' -n 1`.strip
+    end
```
mssola: 2 spaces for indentation level ;)

app/views/layouts/application.html.slim
```diff
@@ -39,6 +39,9 @@ html
         = yield
   footer
     .aside-footer
+      p
+        - unless Git::BRANCH.empty?
+          '#{Git::BRANCH}@#{Git::COMMIT}
```
mssola: Maybe it would be better that the `Git` module provided a method named `version` or something, thus moving the logic to the ruby code. Plus, I'd also want to see the commit even if the branch is empty. This can happen for example by performing the following commands:

```
$ git remote add lonewulf https://github.com/lonewulf/Portus
$ git fetch lonewulf
$ git checkout lonewulf/669_git_webgui
```

In this case `BRANCH` is empty but `COMMIT` is not.

config/initializers/git_info.rb
```diff
@@ -0,0 +1,7 @@
+module Git
+    `which git`
+    if $?.success?
```
mssola: I'd prefer that you required the `mkmf` library, and then used the `find_executable` method. It's cleaner.

config/initializers/git_info.rb
```diff
@@ -0,0 +1,7 @@
+module Git
```
mssola: Could you add documentation? It doesn't have to be too thorough :)

mssola:
Nice! This might fix #669. Before doing that, I think that we can add some improvements:

- What if the branch + commit combo is too large ? Instead of putting this information in the left side, I think that it would be better to put it on the footer, since we have a *lot* of horizontal space there :)
- It would be awesome if we could detect that commit X has actually a tag. This would be cool for example in the case that you are using an official release of Portus, since then you'd see something like "Version 2.0.1" or something.
- I think that it would be better to hide this information to regular users. Maybe we can check whether the current user is an admin before displaying this ?

Thanks!

app/views/layouts/application.html.slim
mssola: Actually, on a second thought, maybe you could rename this module to `Version`, and the method could be `self.to_s`. What do you think ?

config/initializers/git_info.rb
jordimassaguerpla: what about displaying also the version in 

https://github.com/SUSE/Portus/blob/master/VERSION

mssola: @jordimassaguerpla I think that we could do that if picking it from git is not working. My reasons:

1. I'd prefer to stick to git, since it would be more reliable than the `VERSION` file.
2. There's one case in which the git command won't work: when using a distro package (e.g. the RPM that we produce for openSUSE; good catch @jordimassaguerpla :P).

So, pick it from git if you can, but if you can't, just stick to the `VERSION` file.

app/views/layouts/application.html.slim
lonewulf: Sounds right to me

config/initializers/git_info.rb
lonewulf: I need to figure out why find_executable crashes on my alpine based container first, but it was my intention in the first place.

config/initializers/git_info.rb
lonewulf: Right !

lonewulf:
Should do the trick :)

app/views/layouts/application.html.slim
```diff
@@ -44,3 +44,7 @@ html
         'Created with
         i.fa.fa-heart
         'by the SUSE team
+      p
+        - if current_user.admin?
+          - if Version.value.present?
```
mssola: Since this function might return nil, you can merge both if's as:

    - if current_user.admin? && Version.value

config/initializers/version.rb
```diff
@@ -0,0 +1,32 @@
+require 'mkmf'
+
+module Version
+
+  # Version module
```
mssola: The documentation of this module should go just above the declaration. Just like we do for example withe the [LDAP](https://github.com/SUSE/Portus/blob/master/lib/portus/ldap.rb#L24) class.

config/initializers/version.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  git = find_executable('git')
+  VERSION = File.read(File.join(Rails.root,'VERSION')).chomp rescue nil
+  BRANCH = git ? `git symbolic-ref HEAD 2>/dev/null | cut -d"/" -f 3  2>/dev/null`.chomp : nil
+  COMMIT = git ? `git log --pretty=format:'%h' -n 1 2>/dev/null`.chomp : nil
+  TAG = git ? `git tag --points-at $(git rev-parse HEAD) 2>/dev/null`.chomp : nil
+  
+  # Version.value returns the app version
+  # Priority: git tag > git branch/commit > VERSION file
+
```
mssola: Remove this empty line

config/initializers/version.rb
```diff
@@ -0,0 +1,32 @@
...
+      if BRANCH.present?
+        "#{BRANCH}@#{COMMIT}"
+      else
+        "#{COMMIT}"
+      end
+    elsif VERSION.present?
+      "#{VERSION}"
+    end
+  end
+
```
mssola: Remove this empty line

mssola:
Nitpicking, sorry :) Functionality-wise, it looks perfect. Thanks!

config/initializers/version.rb
lonewulf: Which line number are you talking about ?

config/initializers/version.rb
lonewulf: Same question :)

config/initializers/version.rb
```diff
@@ -0,0 +1,32 @@
+require 'mkmf'
+
+module Version
+
+  # Version module
+  # Makes the app version available to the application itself
+  # Needs the git executable for all git operations
+
+  git = find_executable('git')
+  VERSION = File.read(File.join(Rails.root,'VERSION')).chomp rescue nil
```
mssola: I don't like that `rescue` honestly. I'd prefer that you moved this logic in the `value` method. So instead of:

```ruby
elsif VERSION.present?
  "#{VERSION}"
end
```

It should be:

```ruby
else
  version = Rails.root.join("VERSION")
  File.read(version).chomp if File.exists?(version)
end
```

It's both safer and cleaner.

lonewulf:
Bah, force pushed a simple fixup commit :(

I think everything should be fine now tough

lonewulf:
Not sure why the test suite failed on this one...

mssola:
@lonewulf don't worry, that failure is not because of your commit. It happens randomly, and we have to fix it anyways.

app/views/layouts/application.html.slim
```diff
@@ -44,3 +44,6 @@ html
         'Created with
         i.fa.fa-heart
         'by the SUSE team
+      p
```
mssola: Could you include the `p` tag inside of the if statement. Otherwise for non-admin users they will get an extra empty paragraph :)

mssola:
Argh, sorry for not noticing that before :(

lonewulf:
That's what code review is for :)

lonewulf:
I'll squash it all if it's fine

mssola:
Perfect! Just squash it all and I'll merge it.

mssola:
@lonewulf thanks!

lonewulf:
@mssola You're welcome

h0tbird:
Thank you! I have just built Portus from master and I see `Version: 2.0.0` in the footer.
I was expecting the last commit ID or at least 2.0.2 which happens to be the last release.
Am I misunderstanding something?

cyphar:
@h0tbird We have a separate branch for 2.0 releases (mainly so we can backport certain patches only into 2.0.x). Unfortunately, this means that `git tag --points-at` doesn't work in this case. It's possible to use `rev-parse` to do this (@jloehel can probably give a few hints with how to do that), but I'm fairly sure that the only way to figure out what tag to use is to do some kind of rebase to figure out what patches were backported to figure out the version.

mssola:
@h0tbird what do you mean by "built" ? :) Are you doing it in a way in which the `.git` directory is destroyed ? Do you have git installed on the machine where Portus resides ?

About `Version: 2.0.0`, I guess that it's picking it from the `VERSION` file, which is the backup solution when git fails. The fact that this file contains `2.0.0` is ... well ... unfortunate :D Let me fix that :)

cyphar:
@mssola If he's building from master, I'm fairly sure you'd get `2.0.0` because that's latest tag on master. Right? Or have we fixed that?

mssola:
@cyphar nope if git is not available. In this case we pick whatever the `VERSION` file says, and right now it says `2.0.0`.

h0tbird:
@mssola You're right I [destroy](https://github.com/h0tbird/docker-portus/blob/18eb19f55455b720c9660b44ae4ac901a6753ea1/Dockerfile#L24-L25) `.git` before I run `bundle install`. I do that to keep the container image smaller.

mssola:
@h0tbird yeah, makes sense. I guess that I'll update that file to `2.1-dev` in master. For your case though, maybe you can update this file with the info you want from git before destroying it ?

h0tbird:
Yes, I guess I can do that. Thank you for the idea!

h0tbird:
Done. Now I see `Version: 98ff345` :)

jordimassaguerpla:
What about having "master" inside the VERSION file?

Writing something like 2.1-dev  means we should know which version is going to be the next one, and we don't know if it will be 2.1 or 3.0 ....

jordimassaguerpla:
In second thought, having 2.1-dev lgtm

PR #735, opened 2016-02-15
Support multiple scope parameters in a token request

databus23:
While testing Portus as an authorisation server for a private docker registry I ran into occasional authentication errors especially when pushing to repositories that don't exist yet.

The error I'm seeing from the docker client is "authentication required" and after re-trying for 1-2 times it always works.
Example:
```
$> docker push 192.168.163.129:5000/busybox
The push refers to a repository [192.168.163.129:5000/busybox]
5f70bf18a086: Preparing
2c84284818d1: Preparing
unauthorized: authentication required
$> docker push 192.168.163.129:5000/busybox
The push refers to a repository [192.168.163.129:5000/busybox]
5f70bf18a086: Pushed
2c84284818d1: Pushed
latest: digest: sha256:9176b113ee7620cb59f5df4c8afa86e990529e3ca328fe15cbc499eeb64bedfd size: 711
5f70bf18a086: Layer already exists
2c84284818d1: Layer already exists
v2: digest: sha256:9176b113ee7620cb59f5df4c8afa86e990529e3ca328fe15cbc499eeb64bedfd size: 711
```
When this authentication errors happens the registry shows the error:
```
registry_1 | time="2016-02-15T14:53:14Z" level=warning msg="error authorizing context: insufficient scope" ...
```
Looking at the Portus rails logs, the problem is that the token request from the docker client contains multiple scope query parameters:
```
Started GET "/v2/token?account=d062284&scope=repository%3Abusybox%3Apush%2Cpull&scope=repository%3Ad062284%2Fbusybox%3Apull&service=192.168.163.129%3A5000"
```
So the docker client (version 1.10.0) for some reason requests `[push,pull]` and `[pull]` actions for the same repository in one token request. The problem here is that the query parsing from rails mangles this and only the last scope parameter gets populated into `params[:scope]`. Unfortunately the last parameter often only contains the `[pull]` action.

Quoting the [official distribution spec] (https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md) having multiple scope query parameters is valid:
>  This query parameter should be specified multiple times if there is more than one scope entry from the WWW-Authenticate header.

So this commit changes the token controller to support multiple scope parameters.

In the process I also had to change the controller to not raise `401 Unauthorised` on authorisation errors which is also explicitly discouraged by the official spec:
> If the client only has a subset of the requested access it **must not be considered an error** as it is not the responsibility of the token server to indicate authorization errors as part of this workflow

This is repeated in the [token specification] (https://github.com/docker/distribution/blob/master/docs/spec/auth/jwt.md):
> The token server should not return errors when the user does not have the requested authorization. Instead, the returned token should indicate whatever of the requested scope the client does have (the intersection of requested and granted access).

So with this change Portus only returns a `401 Unauthorized` on **authentication** errors. I manually checked that this in line with what the official registry does.

### DIFF
--- a/lib/portus/jwt_token.rb
+++ b/lib/portus/jwt_token.rb
@@ -6,12 +6,12 @@ module Portus
   #
   class JwtToken
     # The constructor takes the query parameters as specified in the
-    # specification. The given scope is assumed to have already been processed
+    # specification. The given scopes are assumed to have already been processed
     # by the caller with the authorized scopes.
-    def initialize(account, service, scope)
+    def initialize(account, service, scopes)
       @account = account
       @service = service
-      @scope   = scope
+      @scopes   = scopes
     end
 
     # Returns a hash containing the encoded token, ready to be sent as a JSON
@@ -32,7 +32,7 @@ module Portus
         hash[:nbf]    = issued_at - 5.seconds
         hash[:exp]    = issued_at + expiration_time
         hash[:jti]    = jwt_id
-        hash[:access] = authorized_access if @scope
+        hash[:access] = authorized_access if @scopes
       end
     end
 
@@ -56,11 +56,13 @@ module Portus
 
     # Returns an array with the authorized actions hash.
     def authorized_access
-      [{
-        type:    @scope.resource_type,
-        name:    @scope.resource_name,
-        actions: @scope.actions
-      }]
+      @scopes.map do |scope|
+        {
+          type:    scope.resource_type,
+          name:    scope.resource_name,
+          actions: scope.actions
+        }
+      end
     end
 
     # Returns a (hopefully) unique id for the JWT token.
--- a/app/controllers/api/v2/tokens_controller.rb
+++ b/app/controllers/api/v2/tokens_controller.rb
@@ -20,13 +20,13 @@ class Api::V2::TokensController < Api::BaseController
   # Returns the token that the docker client should use in order to perform
   # operation into the private registry.
   def show
+    authenticate_user! if request.headers["Authorization"]
     registry = Registry.find_by(hostname: params[:service])
-    raise RegistryNotHandled, "Cannot find registry #{params[:service]}" if registry.nil?
 
-    auth_scope = authorize_scopes(registry)
-    authenticate_user! if auth_scope.nil?
+    auth_scopes = []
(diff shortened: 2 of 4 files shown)


mssola:
Wow, thanks ! Conceptually, it makes sense.

> So the docker client (version 1.10.0) for some reason requests [push,pull] and [pull] actions for the same repository in one token request

Yeah, I detected this too when upgrading to docker 1.10. Basically, docker tries to push multiple tags in a single request. This didn't happen before, and so we didn't have this problem :)

I will try this patch as soon as possible, and then I'll take a closer look at the code itself. In the meantime:

- Have you tried this with docker 1.9 ?
- Some tests on the `token_spec.rb` file are failing.

databus23:
@mssola Ups, you are right my inital rubocop commit was a bit hasty and butchered the tests. Fixed now. Thanks!

I tried briefly with docker 1.9 and pushing the same repo with two tags works without problems. You seem to be right about that multiple scopes per request where introduced by docker 1.10. With docker 1.9 I don't see multiple scopes per token request. The changes should be backward compatible anyway. I'm reasonably confident that it will still work with older versions of docker.

lib/portus/jwt_token.rb
```diff
@@ -8,10 +8,10 @@ class JwtToken
     # The constructor takes the query parameters as specified in the
     # specification. The given scope is assumed to have already been processed
```
mssola: Could you update `The given scope is assumed to have already been processed` to  `The given scopes are assumed to have already been processed` ?

mssola:
I've already tested and it works like a charm :clap: 

LGTM. Could you fix my nitpick, squash your commits and sign it off ? Thanks !

databus23:
@mssola squashed and signed.

mssola:
@databus23 sorry to be a bother, but could you keep your commit message inside of 80 columns. Otherwise if I do `git log` from the CLI it's unreadable.

mssola:
@databus23 merged. Thanks for working on this :clap:

I've cherry-picked this to the `v2.0` branch and now I'm performing some tests on it. Once they pass, I'll push it, so it will be available for the `2.0.3` release.

mssola:
This commit is now on the `v2.0` branch.

PR #740, opened 2016-02-16
remove source_url from schema.rb

databus23:
I think this slipped in by accident. There is no migration for this in the master branch. It seems this has is origins in this commit https://github.com/SUSE/Portus/commit/c822bb0d636cc6088e7804bf78bea0f6eec5b6ae which is part of the hackday_13 branch


mssola:
Good catch.

Issue #487, opened 2015-10-26
Activity messages do not fill all the horizontal space available


mssola:
I've got an activity with a message that is slightly over the 400px max-width:

![logos3](https://cloud.githubusercontent.com/assets/767943/10728845/a6e29d04-7be5-11e5-9f18-cd9b93e88caf.png)

I think that messages should use all the horizontal space that is left for them. Note that we cannot just remove max-width, since then it would take the whole "li" space for just the message on very large messages. Also note that current activities follow a responsive design, so just raising the max-width should not be acceptable (I think, but I'm no expert :P).

PR #776, opened 2016-03-06
Activity messages now fill all the horizontal space available

msfernandes:
Fixes: #487 

Large screen:
screenshot-localhost 5000 2016-03-06 01-37-23

Medium screen:
screenshot-localhost 5000 2016-03-06 01-38-13

Small screen:
screenshot-localhost 5000 2016-03-06 01-38-41

When the namespace, for example, is too large, the div with the description exceeds the Recent Activities div:
screenshot-localhost 5000 2016-03-06 01-58-43

So I added `overflow: hidden` for avoid this layout break (see small screenshot). But I think that is better set a max length to these names.

Signed-off-by: Matheus Fernandes <matheus.souza.fernandes@gmail.com>

### DIFF
--- a/app/assets/stylesheets/activities.scss
+++ b/app/assets/stylesheets/activities.scss
@@ -8,6 +8,10 @@
   border-left: 1px solid $gray-light;
   margin-left: 2em;
   margin-top: -15px;
+  .activitie-container {
+    display: flex;
+    overflow: hidden;
+  }
   img {
     border-radius: 100%;
     border: 4px solid $gray-light;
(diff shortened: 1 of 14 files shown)


cyphar:
What about if we wrap it, I'm not sure about truncating messages if they're long. I'd prefer making them taller by wrapping the text.

msfernandes:
`text-overflow` property maybe solve this problem, truncating text. As soon as possible I'll fix it.

msfernandes:
I could not do using html and css, so I used the `truncate` helper from Rails to truncate namespaces, team names, etc, at 40 characters length. But it truncates in large screen too. 

screenshot-localhost 5000 2016-03-07 10-26-26

screenshot-localhost 5000 2016-03-07 10-26-58

This length is ok?

mssola:
Not an expert by any means when it comes to CSS, but wouldn't [this](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/) work ? If that's not the case, maybe truncating into 40 chars is not too bad. We can fix that later.

msfernandes:
Removed rails truncate and added trucante on css!
screenshot-localhost 5000 2016-03-07 13-22-23
screenshot-localhost 5000 2016-03-07 13-27-18
screenshot-localhost 5000 2016-03-07 13-21-40

mssola:
Thanks :clap:

PR #779, opened 2016-03-08
automate portus release

jordimassaguerpla:
This commit adds a "release.sh" script that will release portus in the
build service.

It assumes you will have already tagged and update the VERSION file, as
well as updated the CHANGELOG (see http://port.us.org/docs/how_we_release.html)

It includes a small fix in the make_spec.sh that checks for bundler


mssola:
LGTM :clap:

Issue #586, opened 2015-11-20
The error message is not always located in the user's field of vision.


jloehel:
If one team have a lot of team members and namespaces the user often have to scroll up to see the error message. Just a simple ux note.

PR #781, opened 2016-03-08
Updated alert to always show to user whenever triggered

msfernandes:
Fixes: #586 

If the topbar is on the screen the notification appears below it, else the notification appears on the top.

Large Screens
------

screenshot-localhost 5000 2016-03-08 11-41-53
screenshot-localhost 5000 2016-03-08 11-43-11

Medium Screens
--------

screenshot-localhost 5000 2016-03-08 11-54-17

Small Screens
--------

screenshot-localhost 5000 2016-03-08 11-53-46

Signed-off-by: Matheus Fernandes <matheus.souza.fernandes@gmail.com>

### DIFF
--- a/app/assets/javascripts/alert.coffee
+++ b/app/assets/javascripts/alert.coffee
@@ -1,0 +1,23 @@
+float_alert = exports ? this
+float_alert.refreshFloatAlertPosition = ->
+  box = $('.float-alert')
+  top = box.css('top')
+  if $(this).scrollTop() < 60
+    box.css('top', 72 - $(this).scrollTop() + 'px')
+
+  $(window).scroll ->
+    if $(this).scrollTop() > 60
+      # box.stop().animate { 'top': '12px' }, 70
+      box.css('top', 12 + 'px')
+    else
+      box.css('top', 72 - $(this).scrollTop() + 'px')
+      # box.stop().animate { 'top': top }, 70
+    return
+  return
+
+float_alert.setTimeOutAlertDelay = ->
+  setTimeout ->
+    $(".alert-hide").click()
+  , 4000
+
+float_alert.setTimeOutAlertDelay()
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -16,9 +16,11 @@ class Admin::UsersController < Admin::BaseController
 
     if @user.persisted?
       flash[:notice] = "User created successfully!"
+      flash[:float] = true
       redirect_to admin_users_path
     else
       flash[:alert] = @user.errors.full_messages
+      flash[:float] = true
       render "new"
     end
   end
@@ -34,9 +36,9 @@ class Admin::UsersController < Admin::BaseController
     attr = params.require(:user).permit([:email])
 
     if @user.update_attributes(attr)
-      redirect_to admin_users_path, notice: "User updated successfully"
+      redirect_to admin_users_path, notice: "User updated successfully", float: true
     else
-      redirect_to edit_admin_user_path(@user), alert: @user.errors.full_messages
+      redirect_to edit_admin_user_path(@user), alert: @user.errors.full_messages, float: true
     end
   end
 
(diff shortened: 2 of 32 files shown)


mssola:
This looks really classy, thanks :+1: That being said, there are some places in which we use the `alert` and `alert-info` (or other) classes without them being actually "alerts" :D Take for example the `/admin/registries` page:

screenshot_20160311_121713

In this case I'd prefer the old behavior. Also, I'd also prefer the old behavior in the pages in which users are not logged in and they mess up in a form. For example:

screenshot_20160311_122019

Therefore, this new implementation is handy in most cases, but there are these corner cases that I'd rather keep the old behavior. What do you think ?

msfernandes:
@mssola I agree with you!
Correct me if I'm wrong, but only "danger" and "info" alert are used. So, we could use more alert types to group the alerts in float and fixes  alerts... What do you think?

mssola:
@msfernandes agreed. Feel free to add more alert classes.

msfernandes:
Now, only error messages are displayed as "float-alert". So, when the messages are only for information the alert is shown as in the old behaviour.

I also updated the icon in "danger" alerts to one "!" instead of the "i".

mssola:
@msfernandes looks good to me. Note however that some errors are treated as information though! For example, in the pictures you uploaded on your first comment show some of the alerts that should be considered float-alert :) (there are others though, like creating an invalid team).

mssola:
From the UX and UI point of view, it looks good to me. That being said, you should rebase your commits onto current master, because right now it's impossible for me to review your changes :sweat_smile:

mssola:
I can see that you've added more flashy classes. That's fine, by notice that the vast majority of alerts are floating. Therefore, I'd say that the new classes should be "fixed_alert" or something like that, and those will only be applied in those cases I pointed out.

This way, you would avoid all the majority of class changes, and you would also avoid to add all the time:

```
$('#alert .alert').addClass('float-alert');
refreshFloatAlertPosition();
setTimeOutAlertDelay();
```

msfernandes:
Agreed :)

But, about the code repetition, I think that's necessary because the notification is triggered by ajax and the `refreshFloatAlertPosition` was already executed. So, when the notification is displayed, the position must be adapted to the page scroll, to not override the navbar and to do this I need to execute `refreshFloatAlertPosition` again, every time that a notification is displayed...

mssola:
@msfernandes isn't there a way to do this always ? I fear that all this code repetition will be troublesome in the future.

msfernandes:
@mssola I don't think so... 
What we can do is change the notification position (put in right inferior corner, for example). In this form we don't need to use the `refreshFloatAlertPosition` because the position will be always on the same location.
The `setTimeoutAlertDelay` is important to dismiss the alert after 4 seconds. But need to be executed every time that the notification is triggered, because the time should start counting after that.

So, it's possible reduce these three lines repetition to one...

mssola:
> What we can do is change the notification position (put in right inferior corner, for example). In this form we don't need to use the refreshFloatAlertPosition because the position will be always on the same location.

I see that the alternative is worse :) Let's keep what you have here.

> The setTimeoutAlertDelay is important to dismiss the alert after 4 seconds. But need to be executed every time that the notification is triggered, because the time should start counting after that.

I did not notice that :)

> So, it's possible reduce these three lines repetition to one... 

Yeah, why not ? :) You can have just one function that calls all these functions. That would be better in this case I think.

So, long story short:

- You should treat alerts just like you did in your very first commit, and add special classes for the ones I'm mentioning in [this comment](https://github.com/SUSE/Portus/pull/781#issuecomment-206204701).
- You should write a function that includes `refreshFloatAlertPosition` and `setTimeoutAlertDelay` and use them in AJAX files.

mssola:
@msfernandes that looks better :) Now you only have to fix this failing test and rubocop, and I think it will be ready to be merged :)

mssola:
Perfect :clap: Squash your commits and I'll merge it.

msfernandes:
@mssola squashed!

mssola:
@msfernandes cool, merged ! :clap: Next time, instead of including all your commits' descriptions into your squashed one (and also including sign-offs), just write a single description and only sign it once. Otherwise it's a bit confusing :wink:

msfernandes:
sorry.. hahaha

PR #783, opened 2016-03-08
fix package name plus some refactoring

jordimassaguerpla:
Package name should be portus and not Portus. This rename was necessary
to overcome an issue with the version of the package (see previous
commits).

However, we fixed the package name (and the obsoletes tags) in the spec
file *after* setting the tag 2.0.3, thus we need to fix it in the release
script for this version.

When we release next version from master we'll be able to clean this.

Also, rename Portus.spec.in to portus.spec.in, plus some other
refactoring.


mssola:
Thanks :+1:

PR #803, opened 2016-04-05
Don't allow access to the hidden global team

### LINKED ISSUE #658: Don't allow access to the hidden global team
If you go to the global namespace, you can click into the following link:

snapshot1

1. This link should be hidden for this case.
2. Access to the global team page should never be allowed.

monstermunchkin:
The link to the global team is now hidden. If a user tries to access it
anyway, the user is presented with a 404 error.

Fixes #658

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -14,6 +14,8 @@ class TeamsController < ApplicationController
   # GET /teams/1
   # GET /teams/1.json
   def show
+    raise ActiveRecord::RecordNotFound if @team.name.starts_with?("portus_global_team_")
+
     authorize @team
     @team_users = @team.team_users.enabled.page(params[:users_page]).per(10)
     @team_namespaces = @team.namespaces.page(params[:namespaces_page]).per(15)
--- a/app/views/namespaces/show.html.slim
+++ b/app/views/namespaces/show.html.slim
@@ -43,9 +43,10 @@
       'Namespace:
       strong
         = @namespace.clean_name
-    h6.label.label-info
-      | <span>Belongs to: </span>
-      = link_to "#{@namespace.team.name}", @namespace.team
+    - unless @namespace.global?
+      h6.label.label-info
+        | <span>Belongs to: </span>
+        = link_to "#{@namespace.team.name}", @namespace.team
   .panel-body
     .table-responsive
       table.table.table-stripped.table-hover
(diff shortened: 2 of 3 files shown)


app/views/namespaces/show.html.slim
```diff
@@ -43,9 +43,10 @@
       'Namespace:
       strong
         = @namespace.clean_name
-    h6.label.label-info
-      | <span>Belongs to: </span>
-      = link_to "#{@namespace.team.name}", @namespace.team
+    - unless @namespace.team.name.starts_with?("portus_global_team_")
```
mssola: Wouldn't the following just work ?

```ruby
@namespace.global?
```

app/controllers/teams_controller.rb
```diff
@@ -14,6 +14,10 @@ def index
   # GET /teams/1
   # GET /teams/1.json
   def show
+    if @team.name.starts_with?("portus_global_team_")
```
mssola: I think that you can just raise an ActiveRecord::NotFound or something like that and you wouldn't need the `not_found` method ;)

app/controllers/teams_controller.rb
```diff
@@ -14,9 +14,16 @@ def index
   # GET /teams/1
   # GET /teams/1.json
   def show
+    raise ActiveRecord::RecordNotFound if @team.name.starts_with?("portus_global_team_")
     authorize @team
     @team_users = @team.team_users.enabled.page(params[:users_page]).per(10)
     @team_namespaces = @team.namespaces.page(params[:namespaces_page]).per(15)
+  rescue ActiveRecord::RecordNotFound
```
mssola: you can remove all this. Raising an ActiveRecord::RecordNotFound sends a 404.

monstermunchkin: But then it doesn't show the pretty 404 page.

mssola: It doesn't on development mode, but it does on staging/production :) Just try it with staging (so you don't have to do anything special in the config as in production), you'll get the fancy 404 page.

app/controllers/teams_controller.rb
```diff
@@ -14,9 +14,16 @@ def index
   # GET /teams/1
   # GET /teams/1.json
   def show
+    raise ActiveRecord::RecordNotFound if @team.name.starts_with?("portus_global_team_")
```
mssola: Nitpicking: maybe add an empty line afterwards, to differentiate better this case with the rest of the "block" :)

mssola:
Just fix my two comments and squash your commits :wink:

mssola:
Tests are failing. You should change your test to reflect that you are expecting an exception

mssola:
Thanks :+1:

PR #809, opened 2016-04-11
Set proper rails generator defaults

tboerger:
To avoid the requirement to provide additional flags while using the
builtin rails generators I have added some flags to the application
configuration. Now the generator always uses rspec, slim and
factory_girl by default, and test_unit as a fallback if any generator
doesn't support rspec.


mssola:
Thanks :clap:

PR #810, opened 2016-04-11
Annotate the models, factories and specs

tboerger:
In order to make the start for newbies easier I have added the annotate gem
which adds the available columns including some meta information as
documentation to the top of all models, factories and model specs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,7 @@ end
 unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"
 
   group :development do
+    gem "annotate"
     gem "rails-erd"
     gem "quiet_assets"
     gem "pry-rails"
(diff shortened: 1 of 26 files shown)


mssola:
How did you generate this ? Should be re-run some command whenever we add/remove a column ?

tboerger:
> How did you generate this ? Should be re-run some command whenever we add/remove a column ?

Currently it's not done automatically, at least as far as I know, but as documented at https://github.com/ctran/annotate_models#configuration-in-rails we can enable auto-update on database migration.

mssola:
Well, in there it says that by default it will do that on migrate. If that's the case, then LGTM

tboerger:
Than it should be fine.

mssola:
Thanks :clap:

PR #813, opened 2016-04-12
Enable automated annotation

tboerger:
I have enable the automated annotation, now the annotations are getting
updated on every migrate command within development environment. Beside
that we are listing the indices now as well.


mssola:
Thanks :clap:

Issue #794, opened 2016-03-24
Add webhoooks


flavio:
Webhooks allow to build or set up integrations which subscribe to certain events on Portus. When one of those events is triggered, we'll send a HTTP POST payload to the webhook's configured URL. Webhooks can be used to update an external issue tracker, trigger CI builds, deploy to your production server...

tboerger:
So we will integrate on the namespace level the functionality to send webhooks based on the ideas of https://github.com/drone-plugins/drone-webhook/blob/master/DOCS.md excluding the custom templating for now.

Optionally we can add it also to the repo level later on.

tboerger:
For the first draft version I would suggest the following database structure. Currently it's a mix of drone webhooks and the GitHub webhook functionality.

Initially I would trigger the webhooks and don't care about if it succeeds or not. But to keep it debuggable somehow I would store the last ~10 webhooks within the `webhook_deliveries` table. This table can be truncated automatically with an `after_save` hook. I would simply add a button to retrigger a webhook, that's pretty much the same what GitHub does.

To reduce the pain and keeping it mostly simple I would store JSON blobs for request headers, request body, response headers and response body.

```
bin/rails g model webhook url username password request_method:enum content_type:enum enabled:bool
bin/rails g model webhook_header webhook:references name value
bin/rails g model webhook_delivery webhook:references uuid status request_header:text request_body:text response_header:text response_body:text
```

flavio:
> I would simply add a button to retrigger a webhook, that's pretty much the same what GitHub does.

Fine for me. Who is going to resend the notification, rails or a crono process?

flavio:
Also where should we show the failure? Should it be shown inside of the namespace or should it be shown as a notification to all the owners of the namespace?

I fear that showing that only inside of the namespace page would just bury the error.

Also what happens if there's a delivery failure, but then another notification is received and this time the delivery is fine? I think in that case we could just remove the old failure.

We could also create a new Portus event about failed deliveries to show inside of the activities stream.

I fear this is going to make everything more complex... maybe we should just log the error and that's it. It would be better to add support for 3rd parties logger and leave the system administrator in charge of collecting and dealing the failures.

tboerger:
I would add a simple entry to the activity stream and details within the webhook detail page that needs to be created.

I just took GitHub as an example implementation, there you can simple retrigger every webhook (see screenshot attached).

> Fine for me. Who is going to resend the notification, rails or a crono process?

[...]

PR #814, opened 2016-04-12
[WIP] Webhook integration

tboerger:
I have started to integrate the webhook implementation, this PR is work in progress, I just created this to gather feedback as soon as possible.


app/models/webhook_header.rb
```diff
@@ -0,0 +1,22 @@
+# == Schema Information
+#
+# Table name: webhook_headers
+#
+#  id         :integer          not null, primary key
+#  webhook_id :integer
+#  name       :string(255)
+#  value      :string(255)
```
holgerreif: This limit of 255 chars might be to low, I'd feel more safe with perhaps 1024.

It's not a strict and founded requirement but a gut feeling.

tboerger: I don't think that anybody would send additional header values with a length >255 characters.

holgerreif: I had the impression that there additional information are encoded. Or did I miss the place where a notification payload (HTTP request body) is defined?


--Viele GrüßeHolger Reif

-------- Ursprüngliche Nachricht --------
Von: Thomas Boerger <notifications@github.com> 
Datum: 13.04.2016  09:16  (GMT+01:00) 
An: SUSE/Portus <Portus@noreply.github.com> 
Cc: Holger Reif <holger-git@reif.net> 
Betreff: Re: [SUSE/Portus] [WIP] Webhook integration (#814) 

In app/models/webhook_header.rb:
> @@ -0,0 +1,22 @@
> +# == Schema Information
> +#
> +# Table name: webhook_headers
> +#
> +#  id         :integer          not null, primary key
> +#  webhook_id :integer
> +#  name       :string(255)
> +#  value      :string(255)

I don't think that anybody would send additional header values with a length >255 characters.

tboerger: The body gets mostly defined by the JSON payload from the registry. While sending it we will store it in a text column within the `webhook_deliveries` table

tboerger:
screenshot from 2016-04-13 13-12-11

tboerger:
screenshot from 2016-04-13 13-59-25

app/models/webhook_delivery.rb
```diff
@@ -0,0 +1,26 @@
...
+#  id              :integer          not null, primary key
+#  webhook_id      :integer
+#  uuid            :string(255)
+#  status          :string(255)
+#  request_header  :text(65535)
+#  request_body    :text(65535)
+#  response_header :text(65535)
+#  response_body   :text(65535)
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
```
holgerreif: `created_at`== first delivery attempt, `updated_at` == last delivery attempt?

mssola: Sounds reasonable. @tboerger what do you think ?

app/models/webhook_header.rb
holgerreif: Then with current data structures you can only forward notifications from the registry as they come in. 
   * You cannot forward notifications from portus itself (e.g. someone created a team space.
Solution: Add a column `source` to `webhook` table if this is a desired future feature (I'm neutral). First defined Source is hardcoded and means notification from registry.
   * You cannot filter specific for specific events (e.g. only e specfic repo or specific tag)
Solution: Add a column `filter` of type string(255) to `webhook` table which might be somehow matched against the payload of the event. (I'm neutral too)

flavio: > You cannot forward notifications from portus itself (e.g. someone created a team space. Solution: Add a column source to webhook table if this is a desired future feature (I'm neutral). First defined Source is hardcoded and means notification from registry.

Right now we want to focus on notifications about new repositories/tags being created. In the near future we will support about notifications for delete events.

What kind of other notifications would you like to receive and why?

> You cannot filter specific for specific events (e.g. only e specfic repo or specific tag) Solution: Add a column filter of type string(255) to webhook table which might be somehow matched against the payload of the event. (I'm neutral too)

We decided to leave filtering to the final endpoint.

holgerreif: > Right now we want to focus on notifications about new repositories/tags being created. In the near future we will support about notifications for delete events

I'd hope _all_ registry's notifications referring to some repo in the namespace get forwared. Why filter (see you own comment)?

> What kind of other notifications would you like to receive and why?

There could be notifications e.g. about failed logins, new registrations etc. But as I sad personally I don't have a specific need.

> We decided to leave filtering to the final endpoint.

I'm fine with this design decision. I'd suggest to state this scope explicitely in the opening comment of #794.

app/models/webhook_delivery.rb
tboerger: Yeah will be handled like that.

app/models/webhook_header.rb
flavio: >>  Right now we want to focus on notifications about new repositories/tags being created. In the near future we will support about notifications for delete events
>
>I'd hope all registry's notifications referring to some repo in the namespace get forwared. Why filter (see you own comment)?

Right now we don't see the need to forward events triggered from portus (like adding a new user, changing the visibility of the repository,...). They are already tracked by the audit feature, that can be later on opened-up via an API if needed.

I see the webhooks as something useful for integrating Portus/distribution into a workflow. So I would expect these webhooks to be used in conjunction with something like Jenkins.

holgerreif: > I see the webhooks as something useful for integrating Portus/distribution into a workflow.

I'm absolutely with you. This is actually my use case ;-)

> Right now we don't see the need to forward events triggered from portus

Fine with me as well.

What I'm concerned is

> Right now we want to focus on *notifications about new repositories/tags being created*. In the near future we will support about notifications for delete events

Why do you want to filter out delete events? Why don't we pass _all_ notifcations from registry regarding the namespace to the regsitered endpoint?

holgerreif: > Why do you want to filter out delete events?

@flavio use case for delete notifications would be synchronisation of repos between different registries.

mssola: Good point. See discussion in the [mailing list](https://groups.google.com/forum/#!topic/portus-dev/x-xuJDRjrIQ).

flavio: >>    Why do you want to filter out delete events?
>
> @flavio use case for delete notifications would be synchronisation of repos between different registries.

I never said I want to filter delete events. On the opposite I said in the future the will support them (they have just been introduced, I don't know if we have the code to handle them yet).

tboerger:
Closed in favor of https://github.com/SUSE/Portus/pull/823

PR #815, opened 2016-04-12
Integrated guard into development environment

tboerger:
In order to make the testing while writing code easier I have integrated
guard which watches for file changes and executes the responsible specs
to get feedback from the test suite as soon as possible.

Additionally to that I have also integrated breakman checks to detect
possible security issues and I have also added rubocop to guard to keep
the code style clean.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -64,6 +64,10 @@ unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"
     gem "pry-rails"
     gem "git-review", require: false
     gem "rack-mini-profiler", require: false
+    gem "brakeman", require: false
+    gem "guard", require: false
+    gem "guard-rubocop", require: false
+    gem "guard-rspec", require: false
   end
 
   group :development, :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -61,6 +61,16 @@ GEM
       sass (>= 3.2.19)
     bootstrap-typeahead-rails (0.10.5.1)
       railties (>= 3.0)
+    brakeman (3.2.1)
+      erubis (~> 2.6)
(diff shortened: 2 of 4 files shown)


tboerger:
Maybe I should mention that you just need to execute `bin/guard` after bundling the ruby gems.

mssola:
Cool, could you squash your commits ?

tboerger:
> Cool, could you squash your commits ?

Which one? I have separate commits intentional.

tboerger:
I have dropped the other commits.

mssola:
Remove `lib/tasks/breakman.rake` and `bin/guard`. Besides this, LGTM

flavio:
> Remove lib/tasks/breakman.rake and bin/guard. Besides this, LGTM

Why do you want to remove them, aren't they needed to run the code?

Also, do we have to integrate breakman into our CI tests?

tboerger:
At least we should keep the breakman task as this configures brakeman. `bin/guard` is not a hard requirement as this can also be invoked by `bin/bundle exec guard`, even if i prefer the shorter one :)

tboerger:
> Also, do we have to integrate breakman into our CI tests?

Makes sense, than it will fail if somebody introduces a possible security issue.

mssola:
> Why do you want to remove them, aren't they needed to run the code?

The `bin/guard` file is redundant with the executable itself or `bundle exec guard`. 

> At least we should keep the breakman task as this configures brakeman.

Fair enough, seems like I read that too fast :sweat_smile: Then this file can be kept. Note though that you spelled it like breakman instead of brakeman :)

> Also, do we have to integrate breakman into our CI tests?

+1

mssola:
Thanks :+1:

PR #816, opened 2016-04-12
Added a bin script and rake task for easier rubocop invocation

tboerger:


mssola:
As discussed on IRC, I don't like adding a new script on `bin` for each tool. I expect developers to run `bundle exec rubocop`

PR #823, opened 2016-04-18
Webhook integration

tboerger:
**This is a replacement for https://github.com/SUSE/Portus/pull/814 so that somebody else can takeover this feature**

I have started to integrate the webhook implementation, this PR is work in progress, I just created this to gather feedback as soon as possible.

Things that still needs to be done:

* [x] Backend that triggers the defined webhooks on event
* [x] Fix activities for deleted webhooks
* [x] Webhook delete button fixed
* [x] Listing of webhook deliveries
* [x] Retrigger webhook functionality
* [x] Write full specs


monstermunchkin:
Is the build failing because of changes made in this PR, or because of some other non-related issue?

mssola:
@monstermunchkin that seems related to you :confused:

monstermunchkin:
OMG it's all green!
happy

app/controllers/webhook_deliveries_controller.rb
```diff
@@ -0,0 +1,29 @@
+class WebhookDeliveriesController < ApplicationController
```
mssola: Documentation

app/controllers/webhook_deliveries_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  respond_to :html, :js
+
+  before_action :set_namespace
+  before_action :set_webhook
+  before_action :set_webhook_delivery
```
mssola: Since you are only using these helpers in the `update` method, and they are trivial, I'd prefer to move the code into the `update` method altogether.

app/controllers/webhook_headers_controller.rb
```diff
@@ -0,0 +1,49 @@
+class WebhookHeadersController < ApplicationController
```
mssola: Documentation

app/controllers/webhook_headers_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  respond_to :html, :js
+
+  before_action :set_namespace
+  before_action :set_webhook
+  before_action :set_webhook_header, only: [:destroy]
```
mssola: Since this last one is only used in the `destroy` method and both the `destroy` method and this one are trivial, it would be better to merge the code into the `destroy` method.

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,94 @@
+class WebhooksController < ApplicationController
```
mssola: Documentation :wink:

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+  before_action :set_webhook, only: [:update, :show, :destroy, :toggle_enabled]
+
+  after_action :verify_authorized, except: [:index]
+  after_action :verify_policy_scoped, only: :index
+
+  # GET /namespaces/1/webhooks
+  # GET /namespaces/1/webhooks.json
+  def index
+    @webhooks = policy_scope(Webhook)
+      .where("namespace_id = ?", @namespace.id)
```
mssola: `where(namespace: @namespace)`

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+      .page(params[:page])
```
mssola: And you can probably put this as a one-liner :wink:

app/models/namespace.rb
```diff
@@ -48,6 +49,22 @@ def self.get_from_name(name)
     [namespace, name]
   end
 
+  # Find the namespace for the given push event.
+  def self.find_from_event(event)
```
mssola: This method is almost identical to the `Namespace.get_from_name` one. Are you sure that you can't re-use the code from the other method ? (either in the caller, or by adding an extra parameter in `Namespace.get_from_name` or something).

app/models/repository.rb
```diff
@@ -7,7 +7,6 @@
 #  namespace_id :integer
 #  created_at   :datetime         not null
 #  updated_at   :datetime         not null
-#  marked       :boolean          default("0")
```
mssola: Nope :wink:

app/models/tag.rb
```diff
@@ -10,7 +10,6 @@
 #  user_id       :integer
 #  digest        :string(255)
 #  image_id      :string(255)      default("")
-#  marked        :boolean          default("0")
```
mssola: Same as before

mssola:
I started reviewing your code but I have to go. I'll continue tomorrow :wink:

app/controllers/webhook_deliveries_controller.rb
tboerger: But IMHO it's best practice to put that into an action. Already prepared if it gets more methods.

mssola: Whenever that happens, it will be trivial to separate it into a separate action :wink:

app/models/webhook.rb
```diff
@@ -0,0 +1,113 @@
...
+#
+#  index_webhooks_on_namespace_id  (namespace_id)
+#
+
+require "base64"
+require "typhoeus"
+require "securerandom"
+require "json"
+
+class Webhook < ActiveRecord::Base
```
mssola: Documentation

app/models/webhook.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  enum request_method: ["GET", "POST"]
+  enum content_type: ["application/json", "application/x-www-form-urlencoded"]
+
+  belongs_to :namespace
+
+  has_many :deliveries, class_name: "WebhookDelivery"
+  has_many :headers, class_name: "WebhookHeader"
+
+  validates :url, presence: true
```
mssola: Maybe it would be even better to validate whether the given URL is valid. A validation of this can perfectly be done with Ruby's stdlib like this (just googled for this :P):

```ruby
require 'uri'

begin
  uri = URI.parse(url)
  uri.kind_of?(URI::HTTP)
rescue URI::InvalidURIError
  false
end
```

This could be put in a validor (e.g. what we do with `NamespaceValidator`).

app/models/webhook_delivery.rb
```diff
@@ -0,0 +1,70 @@
...
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#
+# Indexes
+#
+#  index_webhook_deliveries_on_webhook_id           (webhook_id)
+#  index_webhook_deliveries_on_webhook_id_and_uuid  (webhook_id,uuid) UNIQUE
+#
+
+class WebhookDelivery < ActiveRecord::Base
```
mssola: Documentation

app/models/webhook_delivery.rb
```diff
@@ -0,0 +1,70 @@
...
+  belongs_to :webhook
+
+  validates :uuid, uniqueness: { scope: :webhook_id }
+
+  def success?
```
mssola: Documentation

app/models/webhook_delivery.rb
```diff
@@ -0,0 +1,70 @@
...
+      # update `updated_at` field
+      touch
+    end
+
+    request.run
+  end
+
+  private
+
+  def fetch_headers
```
mssola: Documentation

app/models/webhook_header.rb
```diff
@@ -0,0 +1,22 @@
...
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+# Indexes
+#
+#  index_webhook_headers_on_webhook_id           (webhook_id)
+#  index_webhook_headers_on_webhook_id_and_name  (webhook_id,name) UNIQUE
+#
+
+class WebhookHeader < ActiveRecord::Base
```
mssola: Documentation

app/policies/webhook_policy.rb
```diff
@@ -0,0 +1,46 @@
...
+          .joins(team: [:team_users])
+          .where(
+            "(namespaces.public = :public OR team_users.user_id = :user_id) AND " \
+            "namespaces.global = :global AND namespaces.name != :username",
+            public: true, user_id: user.id, global: false, username: user.username)
+          .pluck(:id)
+
+        scope
+          .includes(:headers, :deliveries)
+          .where("namespace_id = ?", namespaces)
```
mssola: `where(namespace: namespaces)`

lib/portus/registry_notification.rb
```diff
@@ -18,7 +18,9 @@ def self.process!(data, handler)
         next unless HANDLED_EVENTS.include?(action)
         Rails.logger.info "Handling '#{action}' event:\n#{JSON.pretty_generate(event)}"
 
-        handler.send("handle_#{action}_event".to_sym, event)
+        handlers.each do |handler|
```
mssola: Prefer a one-liner with braces.

spec/controllers/webhook_headers_controller_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      :namespace,
+      team:        team,
+      description: "short test description",
+      registry:    registry)
+  end
+  let(:webhook) { create(:webhook, namespace: namespace) }
+
+  before :each do
+    # trigger creation of registry
+    registry
```
mssola: Just don't :smile:. Simply create the registry with `let!` instead of `let` previously. The same should go for the rest of the variables defined in `let` blocks. For what I've seen in other places in our own tests, using `let` ends up with "funny surprises". For this, I prefer using `let!` as much as possible, even if it's slower than `let`.

spec/controllers/webhooks_controller_spec.rb
```diff
@@ -0,0 +1,400 @@
...
+  let!(:webhook) do
+    create(
+      :webhook,
+      namespace: namespace
+    )
+  end
+
+  before :each do
+    # trigger creation of registry
+    registry
```
mssola: Same as before

spec/factories/repositories.rb
```diff
@@ -7,7 +7,6 @@
 #  namespace_id :integer
 #  created_at   :datetime         not null
 #  updated_at   :datetime         not null
-#  marked       :boolean          default("0")
```
mssola: Nope :smile:

spec/jobs/catalog_job_spec.rb
```diff
@@ -8,6 +8,10 @@ def update_registry!(catalog)
 end
 
 describe CatalogJob do
+  before :each do
+    VCR.turn_on!
```
mssola: Why ? As in, which piece of code is triggering a remote access ?

spec/models/repository_spec.rb
```diff
@@ -7,7 +7,6 @@
 #  namespace_id :integer
 #  created_at   :datetime         not null
 #  updated_at   :datetime         not null
-#  marked       :boolean          default("0")
```
mssola: Nope

spec/models/tag_spec.rb
```diff
@@ -10,7 +10,6 @@
 #  user_id       :integer
 #  digest        :string(255)
 #  image_id      :string(255)      default("")
-#  marked        :boolean          default("0")
```
mssola: Nope :yum:

spec/models/repository_spec.rb
```diff
@@ -28,6 +27,10 @@ def get_url(repo, tag)
   it { should have_many(:tags) }
   it { should have_many(:stars) }
 
+  before :each do
+    VCR.turn_on!
```
mssola: Same question as before :)

spec/jobs/catalog_job_spec.rb
monstermunchkin: Good question. All I know is that this gives us 100% coverage. Maybe that's just a happy little coincidence.

mssola: My gut feeling is that the `handle_push_event` method is to blame :grin: Maybe you should track which test triggers this situation and enable VCR there, or provide a VCR for that case.

mssola:
First and foremost, you did a great job, thanks :clap:. That being said, since this is a big feature, there's also some points I'd like you to consider :grin:.

I think that the following actions should only be available to owners and admins:

- Create/remove a webhook.
- Enable/disable a webhook.
- Edit a webhook
- Create a new header
- Re-trigger a delivery

This should be forbidden in the controller and simply not shown on the client if the current user does not meet the required privileges.

On the UX side:

- When you remove a webhook/header there should be a flashy message. The same when you enable/disable them.
- The flashy message for creating a header could be more explicit. Instead of "New header created", we could say "Header 'myheader' created successfully". The same applies to flashy messages from other actions (yes, this is something that we should also improve in other pages :sweat_smile:)
- There should be more links. For example, on the webhooks index page, there should be a link to the namespace. This could be accomplished in the table header like this "Webhooks for namespace", where "namespace" is a link to the namespace.
- There should be a help section in the webhooks index page explaining what are webhooks (similarly to what we do in the namespaces page, even though we could do a better job there...). This also applies for the webhooks show page, in which it would be could have help elements for each table (like we do on the teams show page for example).
- In the tables regarding namespaces (see pages namespaces#index and teams#show), there should be a counter of webhooks for each namespace. It would be even better if said counters where in fact a link to the webhooks#index page for that specific namespace.

I created a webhook called `www`. When I wanted to rename it, I clicked Save and nothing happened :confused: Then, afterwards I realized that the changes where saved, but in order to get them I had to reload the page. So here I can see two issues:

- There is no way for the user to understand what happened (this is something that happens on other actions, as stated previously).
- The JS code should handle this, updating the proper fields on update.

monstermunchkin:
Thanks for the feedback, I have addressed the commented issues. Regarding VCR, I inserted a few `VCR.turn_on!` lines because it had reported that VCR was off — don't ask me why. However, they were only inserted where errors were reported.

The UX issues have not been addressed yet.

monstermunchkin:
On webhooks#show, the name of the webhook is not the actual URL. This is intentional as it's neither pretty nor a good idea to have it read `http://www.example.com:12345/ridiculously/long/path/to/somewhere/ webhook`. Instead, it only shows the host part if the URL which would result in `www.example.org webhook`.

mssola:
@monstermunchkin sounds reasonable to me.

monstermunchkin:
Do not merge yet, there seems to be a small bug deleting webhooks when there are headers tied to the object.

mssola:
@monstermunchkin sure. Just ping me whenever I can start reviewing this PR again :wink:

app/controllers/webhooks_controller.rb
mssola: *Edit*: done

app/controllers/webhook_deliveries_controller.rb
```diff
@@ -0,0 +1,30 @@
+# WebhookDeliveriesController manages the updates of webhook deliveries.
+class WebhookDeliveriesController < ApplicationController
+  respond_to :html, :js
+
+  after_action :verify_authorized
+
+  # PATCH/PUT /namespaces/1/webhooks/1/deliveries/1
+  def update
```
mssola: I said that I wanted you to move these (before) callbacks into the `update` method. I'd prefer to move the *code* altogether, not that you call them explicitly.  Thus, imo this method should be:

```ruby
def update
  @namespace = Namespace.find(params[:namespace_id])
  @webhook = @namespace.webhooks.find(params[:webhook_id])
  @webhook_delivery = @webhook.deliveries.find(params[:id])

  authorize @webhook_delivery

  @webhook_delivery.retrigger
  render template: "webhooks/retrigger", locals: { webhook_delivery: @webhook_delivery }
end
```

Moreover, it looks like some instance variables like `@namespace` and `@webhook` can be just mere locals. Am I right ?

app/models/webhook.rb
```diff
@@ -0,0 +1,153 @@
...
+#
+
+require "base64"
+require "typhoeus"
+require "securerandom"
+require "json"
+require "uri"
+
+# A Webhook describes a kind of callback to an endpoint defined by an URL.
+# Futher parameters are username and password, which are used for basic
```
mssola: Futher -> Further

app/models/webhook.rb
```diff
@@ -0,0 +1,153 @@
...
+  # default to http if no protocol has been specified. If unspecified, the URL
+  # validator will fail.
+  before_validation do
+    unless url.nil? || url.strip.empty?
+      self.url = "http://#{url}" unless url.start_with?("http://") ||
+          url.start_with?("https://")
+    end
+  end
+
+  before_destroy do
```
mssola: You can achieve this by saying that `headers` and `deliveries` depend on this model. If I'm not mistaken, this can be done like this in the `has_many` calls:

```ruby
has_many :deliveries, class_name: "WebhookDelivery", dependent: :delete_all
has_many :headers, class_name: "WebhookHeader", dependent: :delete_all
```

app/models/webhook.rb
```diff
@@ -0,0 +1,153 @@
...
+
+  has_many :deliveries, class_name: "WebhookDelivery"
+  has_many :headers, class_name: "WebhookHeader"
+
+  validates :url, presence: true, url: true
+
+  # default to http if no protocol has been specified. If unspecified, the URL
+  # validator will fail.
+  before_validation do
+    unless url.nil? || url.strip.empty?
```
flavio: Do we really need to do a `url.strip`, I doubt the user can add carriage return symbols via the UI.

I think we can replace this line with a simpler: `unless url.blank?`

app/views/namespaces/_namespace.html.slim
```diff
@@ -1,6 +1,7 @@
 tr id="namespace_#{namespace.id}"
   td= link_to namespace.clean_name, namespace
   td= namespace.repositories.count
+  td= namespace.webhooks.count
```
flavio: I just wonder, wouldn't it better to have a link leading to the page where all the webhooks are shown? Would that be possible? If yes I think we can deal with that by opening an issue and making a dedicated PR. This one is already pretty big.

app/views/public_activity/webhook/_create.csv.slim
```diff
@@ -0,0 +1 @@
+= CSV.generate_line(['webhook', activity.trackable.url, 'create', '-', activity.owner.username, activity.created_at, "owned by namespace #{activity.trackable.namespace.name}"])
```
flavio: This is totally unrelated with this PR, however I just realized one thing.

Suppose:

  1. user triggers webhook for namespace "foo"
  2. user downloads CSV audit file
  3. namespace "foo" is renamed "bar"
  4. user download CSV 

The audit entry for event number 1 is reported in two different ways depending from the generation time. I think this is really wrong, the past events of the audit log should always stay the same!

What do you think about that, this is probably worth a dedicated issue.

monstermunchkin:
The issues have been addressed and everything should work as expected.

Regarding the missing alerts, that maybe also had to do with #883. Furthermore, users don't have access to webhooks, and viewers and contributors cannot create/delete/edit anything related to webhooks. The "weird" issue has been fixed now as well; I hadn't noticed that one before.

Once everyone is satisfied, I'll squash some of my commits.

app/models/webhook.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    namespace, = Namespace.get_from_name(event["target"]["repository"], registry)
+    return if namespace.nil?
+
+    hydra = Typhoeus::Hydra.hydra
+
+    Webhook.where(namespace: namespace).find_each do |webhook|
+      next unless webhook.enabled
+
+      headers, auth = headers_from_webhook(webhook)
```
tboerger: I would suggest an instance method instead of a class method for `headers_from_webhook` like `webhook.processed_headers`

tboerger: And I would create a separate instance method to get the credentials for authentication..

app/models/webhook.rb
```diff
@@ -0,0 +1,148 @@
...
+
+      args = {
+        method:  webhook.request_method,
+        headers: headers,
+        body:    JSON.generate(event),
```
tboerger: Why do you need `JSON.generate`? Hydra can't handle it automatically?

app/models/webhook.rb
```diff
@@ -0,0 +1,148 @@
...
+      }
+    )
+  end
+
+  # headers_from_webhook gathers all headers belonging to the provided webhook.
+  # Also, it creates a basic auth string if username and password are provided.
+  # The function returns both headers and auth string.
+  def self.headers_from_webhook(webhook)
+    headers = { "Content-Type" => webhook.content_type }
+    WebhookHeader.where(webhook: webhook).find_each do |header|
```
tboerger:
```
webhook.headers.each do |header|
```

app/models/webhook.rb
```diff
@@ -0,0 +1,148 @@
...
+      headers[header.name] = header.value
+    end
+    if webhook.username.empty? || webhook.password.empty?
+      auth = ""
```
tboerger: Maybe simply `nil`?

tboerger: For the check if it is set you can also use `if auth.present?` which works also for nil values.

app/models/webhook_delivery.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  validates :uuid, uniqueness: { scope: :webhook_id }
+
+  # success? returns whether or not the request was successful (status 200).
+  def success?
+    status == 200
+  end
+
+  # Retrigger a webhook unconditionally.
+  def retrigger
```
tboerger: Feels like some duplicated code

app/models/webhook_delivery.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    request.run
+  end
+
+  private
+
+  # fetch_headers gathers all headers belonging to the webhook.
+  # Also, it creates a basic auth string if username and password are provided.
+  # The function returns both headers and auth string.
+  def fetch_headers
```
tboerger: Feels like some duplicated code

app/views/public_activity/webhook/_create.html.slim
```diff
@@ -0,0 +1,23 @@
+li
+  .activitie-container
```
tboerger: Is this class name really correct?

app/views/webhook_headers/_webhook_header.html.slim
```diff
@@ -0,0 +1,15 @@
+tr id="webhook_header_#{webhook_header.id}"
+  td= webhook_header.name
+  td= webhook_header.value
```
tboerger: Should we really display all the values to non-owners?

app/models/webhook.rb
monstermunchkin: Nope, rubocop will complain about this.

mssola: why ? Could you paste the complain ?

app/views/public_activity/webhook/_create.html.slim
monstermunchkin: If so, it seems to be wrong in other places as well, e.g. `app/views/public_activity/namespace/_create.html.slim`. Not sure if this matters though.

app/models/webhook.rb
monstermunchkin: If I remember correctly, there were some issues here when not explicitly using `JSON.generate`. I'll look into that again.

app/models/webhook.rb
tboerger: Sorry but the current approach is IMHO not really good. `webhook.headers.each do |header|` should be the way to go.

app/models/webhook.rb
mssola: @monstermunchkin maybe add some specific tests for this, to make sure that there is nothing wrong on the sent body.

app/models/webhook.rb
mssola: Makes sense

app/models/webhook.rb
monstermunchkin: The message is `Use find_each instead of each` [1]. I totally agree that `each` would be better. We have to then change rubocop's behavior regarding this, if possible.

[1] http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/FindEach

app/models/webhook.rb
monstermunchkin: Yes, tests for that would be good. @tboerger, nope, Typhoeus cannot handle it, and the documentation doesn't mention JSON either.

app/models/webhook.rb
mssola: Oh, right, so `webhook.headers.find_each do |header|` :sweat_smile:

tboerger: I would say `webhook.headers.each do |header|` :)

tboerger: > Note: This method is only intended to use for batch processing of large amounts of records that wouldn’t fit in memory all at once. If you just need to loop over less than 1000 records, it’s probably better just to use the regular find methods.

app/models/webhook.rb
```diff
@@ -92,6 +92,21 @@ def host
     host
   end
 
+  # process_headers returns a hash containing the webhook's headers.
+  def process_headers
+    headers = { "Content-Type" => content_type }
```
tboerger: What about this?

```ruby
{
  "Content-Type" => content_type
}.tap do |result|
  headers.each do |header|
    result[header.name] = header.value
  end
end
```

app/views/namespaces/_namespace.html.slim
monstermunchkin: Should this link replace the existing "Show webhooks" link on the namespace page (`app/views/namespaces/show.html.slim`), or should they coexist?

app/views/public_activity/webhook/_create.html.slim
mssola: @tboerger it is correct. I don't recall exactly why it is like this, but it was done to fix an style issue. So, even if ugly, this is not something to be dealt in this PR :wink:

app/views/webhooks/show.html.slim
```diff
@@ -0,0 +1,154 @@
...
+    h5
+      ' Deliveries
+      small
+        a[data-placement="right" data-toggle="popover" data-container=".panel-heading" data-content="A delivery is created once a webhook has been triggered. They are not re-created but updated after retriggering." data-original-title="What's this?" tabindex="0"]
+          i.fa.fa-info-circle
+  .panel-body
+    .table-responsive
+      table.table.table-stripped.table-hover
+        colgroup
+            - if current_user.admin? || @namespace.team.owners.include?(current_user)
```
mssola: Change this line with `can_manage_namespace?(@namespace)`. The same with some lines below

mssola:
LGTM :tada:. Just fix my nitpick and rebase :)

mssola:
:tada: :tada:

flavio:
fireworks

PR #853, opened 2016-05-02
Evaluate ERB in config-local.yml

leshik:
Hello,

This change makes it possible to evaluate ERB code in `config-local.yaml`. This is useful if we don't want to store sensitive information, such as SMTP password, in the config file. Instead we can set the environment variable, then change our config to:
```
...
password: "<%= ENV['SMTP_PASSWORD'] %>"
...
```
I took the idea from quite popular `sidekiq` project, so I suppose it should be safe from security point of view.

### DIFF
--- a/lib/portus/config.rb
+++ b/lib/portus/config.rb
@@ -17,7 +17,7 @@ module Portus
       local = {}
       if File.file?(@local)
         # Check for bad user input in the local config.yml file.
-        local = YAML.load_file(@local)
+        local = YAML.load(ERB.new(IO.read(@local)).result)
         unless local.is_a?(Hash)
           raise StandardError, "Wrong format for the config-local file!"
         end


mssola:
As described in the [documentation](http://port.us.org/docs/Configuring-Portus.html#override-specific-configuration-options), you can already do that. In fact, environment variables will take precedence above all else. Am I missing anything ?

tboerger:
IMHO this change doesn't hurt, this is also possible for the database config of rails.

mssola:
@tboerger but as I said, this is already doable. I wouldn't want to introduce *yet* another way of changing the configuration, when it's basically the same as one of the ways we have :wink: 

In his case, you just have to use the `PORTUS_SMTP_PASSWORD` environment variable, and it will be automatically picked up by Portus.

leshik:
@mssola Ah, sorry — I missed that piece somehow. This definitely solves my problem. Thanks!

mssola:
Closing then.

PR #867, opened 2016-05-06
portusctl: disable automatic generation of certificates

monstermunchkin:
Introduce two new flags regarding SSL certificates:

	* --ssl-gen-self-signed-certs
	* --ssl-certs-dir <dir>

The former, as the name suggests, creates self-signed certificates. The
latter uses certificates located in the specified directory. From there,
it will copy the certificates to the apache directory and continue the
process as before.
Should either one of the certificates be missing, it will fail with a
meaningful message.

If neither flag is specified, it assumes the certificates are already
located in `/etc/apache2/ssl.{key,crt}/`.
As before, should either one of the certificates be missing, it will
fail with a meaningful message.

These two flags are mutually exclusive.

Resolves bsc#978374


mssola:
LGTM. @flavio what do you think ? Moreover, I think that this should not be backported into the `v2.0` branch.

mssola:
@monstermunchkin rubocop complains :sweat_smile:

packaging/suse/portusctl/lib/cli.rb
```diff
@@ -4,6 +4,13 @@ class Cli < Thor
   option "secure", desc: "Toggle SSL usage for Portus", type: :boolean, default: true
 
   # SSL certificate options
+  option "ssl-gen-self-signed-certs",
+    desc:    "Generate self-signed certificates",
+    type:    :boolean,
+    default: false
+  option "ssl-certs-dir",
+    desc:    "Location of own certificates",
+    default: ""
```
flavio: I would also add:
```ruby
  long_desc <<-LONGDESC
Looks for the certificates into the specified folder.
The folloing files must be found:
   * `<custom dir>/file.key`: the certificate key
...
  LONGDESC
```

packaging/suse/portusctl/lib/helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  exit 1
+end
+
+def handle_own_certs(path)
+  puts "Using keys from #{path}"
+  key_file = File.join(path, "#{HOSTNAME}-ca.key")
+  crt_file = File.join(path, "#{HOSTNAME}-ca.crt")
+
+  missing_file(key_file) unless File.exist?(key_file)
+  FileUtils.cp(key_file, "/etc/apache2/ssl.key")
```
flavio: I'm not entirely sure about the usage of this method shared between the regular certificate code and the "bring your own certs".

The output in this case would be a bit confusing, like "Use --ssl-gen-self-signed-certs <...> or --ssl-certs-dir".

I would prefer something more clear like: "Cannot find hostname-ca.crt inside of custom_dir".

flavio:
The code looks good, there are a couple of minor things about the whole user experience that can be improved.

There might be more work related with this PR like:
  * update `portusctl` documentation of the website (do we have it?)
  * update `portusctl` man page (do we have it?)

@mssola I know this is kinda of a new feature, but I don't see any problem in backporting this code to the v2 branch. What worries you?

mssola:
> update portusctl documentation of the website (do we have it?)

We don't. Thus #868.

> update portusctl man page (do we have it?)

Nope, I've also added it on #868.

>  I know this is kinda of a new feature, but I don't see any problem in backporting this code to the v2 branch. What worries you?

Generally speaking, it's discouraged to add a new feature into a patch-level release. That being said, there's been users complaining about this, so it can be considered a fix for a UX bug. So, let's backport it.

PR #883, opened 2016-05-18
hide alert box instead of closing it

monstermunchkin:
After closing an alert box, it is removed from the DOM, and won't show
unless the page is refreshed. This fix hides the alert box instead of
closing it.


mssola:
:clap:

Issue #579, opened 2015-11-18
Ignoring event coming from unknown registry


thenewguy:
So I remember reading about the issue with notifications and the registry/portus being behind compose.

In the case of "Ignoring event coming from unknown registry" (it detects the registry as "registry:5000" instead of docker.internal.example.com) would it not just be simplest to follow up with a http call and pull the event from the registry api?

I am not sure of the notification payload because it isn't logged, but it would seem that one could query the registry api for the notification supplied by the rejected payload and use the reponse.

mssola:
Could you provide more information ? Versions of docker, distribution and Portus (also note that we have released Portus 2.0.3 not so long ago, and that it fixes some known problems with the registry). Even better, we may want to continue this conversation in our new mailing list: https://groups.google.com/forum/#!forum/portus-dev. If we realize that it's an actual bug from Portus, then we can submit a new issue. What do you think ?

jmahowald:
I've just run into something similar.  Finally got my nginx proxy working with both portus and registry on same host.  I don't think it's a problem with Portus and the event  spec per-se, but rather in the ways in which you determine if it is a known registry.  Too tired to do a full write up, but I'll try to follow up tomorrow

mssola:
Closing this issue since it's old and we never got info back.

@jmahowald feel free to leave a comment here anyways, any help that we can get is always appreciated :blush:

jmahowald:
Here's what I believe the fundamental issue is.  There isn't a way to configure the "host" field of the event that the registry uses when it pushes an event.  It appears to use the name that the client used when interacting with the registry.   I have portus though communicate directly with the registry, as SSL termination is happening through nginx.  So that "hostname" in portus is registry:5000.   That's fine since they are linked.  But my clients that want to push go to registry.docker.mycompany.com over 443.  

I think the most correct answer might be to propose a "fix" to the registry to allow the incoming host to be configurable, as the service in the auth token is.   I also think the easier thing for me to fix would be to patch portus to have another field in the registry configuration like external name, or to break up the "hostname" field into host and port, then change the matching logic in the event handler.

Thoughts?

mssola:
Thanks for taking the initiative and submitting a PR :blush: As I said in your PR, I didn't think about this option really :sweat_smile:

PR #888, opened 2016-05-19
Allow external registry name used by clients to diverge from internal name for connections when receiving events

jmahowald:
Description from https://github.com/SUSE/Portus/issues/579#issuecomment-220350888

Also, I know I have no spec tests yet.  My rails skills aren't that great, but I'm happy to add in.  Wanted first to see whether you approve the high level changes


config/config-local.yml
```diff
@@ -0,0 +1,74 @@
...
+  filter: ""
+
+  # The LDAP attribute where to search for username. The default is 'uid'.
+  uid: "uid"
+
+  # LDAP credentials used to search for a user.
+  authentication:
+    enabled: true
+    bind_dn: "CN=dockerregistry_srv,OU=Accounts System,DC=us,DC=int,DC=genesyslab,DC=com"
+    password: "8L2SF5hJ8BWq8tsbxrSqBzuKpUetdf"
```
cyphar: I would change this password if I were you. ;)

jmahowald: Doh.   Thanks, certainly hadn't meant for that file to be in.

mssola: Yeah, just ignore the whole `config-local.yml` file altogether :wink:

mssola:
This is something I certainly didn't think about, thanks :+1: In principle it looks good to me. Just ignore the `config-local.yml` file and provide tests :wink: 

@flavio what do you think ?

flavio:
Sorry for the late response. @jmahowald could you provide some background about this feature. What use case are you trying to solve?

mssola:
Hey @jmahowald, are you still working on this ? Could you rebase your patch ?

jmahowald:
Hey, sorry.  I can work to close this out, but the reason I haven’t done as much is because we decided to roll with gitlab which just added in support for registries and we already have deployed and integrated with our corp AD

mssola:
Closing then. I don't have any objections on this idea (as long as it doesn't over-complicate the setup for regular use cases).  So, if anyone wants to pick it up again and submit a new PR, feel free to do so.

PR #890, opened 2016-05-20
fix release script

jordimassaguerpla:
set disabled mode to the _service file so that the tarball does not
get updated while we copy it from/to different projects


flavio:
LGTM

PR #896, opened 2016-05-24
helper modified to create a <i> instead of <img> . Fix# 312

cyntss:
…ses adapted


mssola:
LGTM. Let me deal with the tests.

PR #900, opened 2016-05-25
Fixed a bug in which deleting images didn't work

### LINKED ISSUE #897: removing image not working in namespace
Hi,

Juste tested image deletion in a docker-compose setup with registry 2.4.0 it's working great with image in the "root" namespace (registry:5000/image).

But not on other namespace for example registry:5000/user1/fedora:23.
In registry log i can see that portus is trying to get the manifest for the wrong repo :

```
registry | "GET /v2/fedora/manifests/23 HTTP/1.0" 401 138 "" "Ruby"
registry | level=error msg="response completed with error" auth.user.name=portus environment=production err.code="manifest unknown" err.detail="unknown tag=23" err.message="manifest unknown" go.version=go1.6.1 http.request.host=registry:5000 http.request.id=9c294197-c201-420c-8ca4-90a101f68442 http.request.method=GET http.request.remoteaddr=172.17.0.1 http.request.uri="/v2/fedora/manifests/23" http.request.useragent=Ruby http.response.contenttype="application/json; charset=utf-8" http.response.duration=5.153083ms http.response.status=404 http.response.written=92 instance.id=79f84872-9b54-45be-be30-c384b9bc6921 service=registry vars.name=fedora vars.reference=23 version=v2.4.0
```

i think it should be /v2/user1/fedora/manifests/23 ?

Best Regards,

mssola:
Apparently trying to delete images and tags in a repository contained inside of
a namespace that is not the global one didn't work. This commit fixes this
while also making sure that this does not happen in many other cases.

Fixes #897

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -40,6 +40,13 @@ class Repository < ActiveRecord::Base
     # options :namespace_name, type: :fulltext
   end
 
+  # Returns the full name for this repository. What this means is that it
+  # returns the bare name if it belongs to the global namespace, otherwise
+  # it prefixes the name with the name of the namespace.
+  def full_name
+    namespace.global? ? name : "#{namespace.name}/#{name}"
+  end
+
   # Set this repo as starred for the given user if there was no star
   # associated. Otherwise, remove the star.
   def toggle_star(user)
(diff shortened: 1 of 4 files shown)


flavio:
LGTM, we should also write an integration test for the delete feature...

mssola:
@flavio there's already integration tests for that actually :) The problem was that the name passed to the client was not the one that was supposed to be given (that's what the tests that I added are checking). For this I've added the issue #903. What do you think ?

PR #913, opened 2016-05-30
create activity on successful action

### LINKED ISSUE #909: failed renaming action creates activity
When attempting to change a team's name to one which already exists, it will notify the user that this is not possible. However, there will be an activity stating that the team has been renamed.

The expected behavior would be to only see an activity if it indeed was successful.

monstermunchkin:
Check whether the update was successful before creating an activity.
This prevents misleading activities.

Fixes #909


mssola:
LGTM

PR #926, opened 2016-06-13
Ui review and fixes

cyntss:
#Main changes in UI

Comments:

image

Header and stars:

image

Login button:

image

Links with standard font:

image

Menu on mouse-over change of colours for more contrast (these colours are only visible onMouseOver):

image

and many more details in the UI !


mssola:
This looks great, thanks :clap:.

Now, for the bad news :wink:. I am experiencing some regressions with this code. First of all, links that are on the right corner of a table/container header are missaligned. See: 

missaligned

This happens in all tables and containers as far as I can tell. Moreover, the stars widget now has a small regression with its title:

title

Finally, for regressions (maybe this one already existed but I'm not sure), when there are no comments we see the following text:

comments

As you can see, there should be at least some left padding for the text.

Finally, I really appreciate the "dots" widget that is shown in the namespaces container for the "Show webhooks" link. See:

webhooks

That being said, it's not really straight-forward to realize what it is from a first glance. Maybe it would be a good idea to add a title to this widget with a text like "More" ?

cyntss:
@mssola 

I made the links not aligned on purpose, because it is not part of the title and rather options, but, if it doesnt not look good at first glance, I will have to fix it :)

Same with the options button. It is very normal through out most applications in the internet to see that icon and understand there are options behind it. But, if it really isnt easy to notice, then I will fix it as well.
Can I ask for some opinion regarding this last topic to @flavio @tboerger ?

monstermunchkin:
The options button/menu is a good idea, however it looks a bit odd next to the big buttons. When I see such a menu (e.g. in the Chromium browser), this options button usually is as big as the other buttons. Would it look weird if this menu button were a bit bigger, or the big buttons a bit smaller?

cyntss:
@monstermunchkin I had that feeling as well. Let me see how I can improve that part then ;)

cyntss:
@mssola I cannot reproduce:

> Moreover, the stars widget now has a small regression with its title:

image

Testing in Chrome and Firefox, neither of them is giving me that issue. What browser are you using?

cyntss:
Fixed the issue with the Options btn:

image

mssola:
> I made the links not aligned on purpose, because it is not part of the title and rather options, but, if it doesnt not look good at first glance, I will have to fix it :)

Fair enough. Let's keep it this way.

> Same with the options button. It is very normal through out most applications in the internet to see that icon and understand there are options behind it. But, if it really isnt easy to notice, then I will fix it as well.

Agreed with @monstermunchkin here.

> I cannot reproduce

This happens to me in Firefox when I hover the mouse over the number. Most surely is because the number also has a title assigned to it (because this is how it was done before this change :wink:)

> Fixed the issue with the Options btn:

Have you added a title saying "More options" to it ?

monstermunchkin:
The option buttons on the left hand side look really nice, and are much more noticeable. However, I guess moving them broke something else:
portus

Also, I can reproduce the regression regarding the stars widget in Chromium.

cyntss:
> Have you added a title saying "More options" to it ?

No, but I can do it of course. Just, first I would like to know how to see that issue you are seeing in the "star repo" button with the tooltip being overlapped by the title of the link. Because I would like to apply the same tooltip to the options button.

cyntss:
@monstermunchkin I didnt read your comment before i sent the previous one :p .. ok, I will see what's the problem with Chromium then, because this is a 100% bootstrap componen (the tooltip).

I didnt see that regression after moving the options.. but I know what the problem is, I will fix it asap :)

cyntss:
@mssola 

> This happens to me in Firefox when I hover the mouse over the number. Most surely is because the number also has a title assigned to it (because this is how it was done before this change 😉)

Will fix it :)


> Have you added a title saying "More options" to it ?

And will add this as well ;)

cyntss:
@mssola @monstermunchkin  I tried in Firefox in Mac, chrome in Mac, Firefox in openSUSE Leap, and Chromium in openSUSE Leap... but there is no way I can reproduce the bug you see with the titles in the stars :(

The code is perfectly fine, there is no duplication of the titles whatsoever.

Are you sure you guys have the latest version of these browsers? It may be a bug that was already fixed in the current versions of chrome and firefox.

cyntss:
@monstermunchkin fixed!

image

@mssola  So.. only the issue with the star is the one I cannot fix, because it is something from Bootstrap in those browsers i cannot reproduce.

I would suggest we create an issue in github if more people see the same issue and perhaps open the issue in bootstrap's github so they can fix it for the versions of Firefox and Chromium you are using.

monstermunchkin:
I looked into this widget behavior again, and it turns out that it only behaves this way in a certain situation, that is when the star is toggled without a subsequent page refresh. Once you refresh the page, all is good again. Also, before you toggle the star, there is no issue.

This can be reproduced on Chromium and Firefox.

cyntss:
@monstermunchkin With that information I was able to reproduce and fix the problem 💃 

hope it is good to be merged now :D

mssola:
LGTM :clap:

monstermunchkin:
Thanks for fixing the double-title issue :+1:

However… I noticed some other strange behavior while testing. When I hover over the star, then click (toggle the star), and then move the cursor away again, the tooltip doesn't disappear. This is reproducible in Chrome and Firefox.

mssola:
@monstermunchkin oops, merged :sweat_smile:. Let's fix it in another issue+PR :wink:

PR #928, opened 2016-06-14
Fixed double alerts bug

msfernandes:
With #781 I added a bug :(
When the user sign up or add a new registry two notifications (one float and one fixed) were triggered.
This PR fixes it :+1:


mssola:
Thanks :+1:

PR #929, opened 2016-06-14
Search improvements

msfernandes:
Fixes: #753 

In this PR I updated the search feature to search through Teams and Namespaces too! So, I added a category menuto filter by repository, team and namespace.

screenshot-localhost 5000 2016-06-14 00-06-26
screenshot-localhost 5000 2016-06-14 00-06-54

The repository search result table is the same as before, I only added "updated_at" column. Then I show in this print a empty result :)

screenshot-localhost 5000 2016-06-14 00-07-13

And in small screens:

screencapture 1

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -30,3 +30,4 @@
 @import 'comments';
 @import 'repositories';
 @import 'responsive-utilities';
+@import 'search';
--- a/app/assets/stylesheets/search.scss
+++ b/app/assets/stylesheets/search.scss
@@ -1,0 +1,44 @@
+.search {
+  .category-panel {
+    .result-counter {
+      float: right;
+      background-color: $primary-colour;
+      border-radius: 0.5rem;
+      padding: 0 0.5rem;
+      font-weight: bold;
+      color: $white;
+      :hover {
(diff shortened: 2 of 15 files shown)


mssola:
The code looks good to me, but apparently code coverage is not at 100%. Could you look into that ?

cyntss:
Looks excellent!

One minor change I would suggest: remove the space between the tabs and the table and then add the results numbers in the tabs =>

image

cyntss:
I made that quickly in photoshop :p .. I didnt put the numbers in the other tabs, but they should be there as well.

cyntss:
One last thing.. I think we could add that style in the tabs in the dashboard:

image

It would look much better!

mssola:
Totally agree with @cynthia on the dashboard. Bear in mind though that this can be done in another PR (it is part of his GSoC to improve the dashboard :wink:).

msfernandes:
Hi @cyntss ! Thanks for the review :)

With your suggestions:

screencapture

mssola:
Now rubocop is the one complaining :sweat_smile:

mssola:
Great :clap: ! Squash your commits and I'll merge it :yum:

PR #930, opened 2016-06-15
New design for documentation menu

msfernandes:
Hi :)

@mssola notify me about one bug on "Created with <3 by the SUSE team" documentation menu: 

image

Analyzing this problem, I saw other issue: when the screen have a small height (and need not to be too small), the quantity of items on the list overflows the screen, then the user lost a piece of menu.

So I designed a new interface for this documentation menu, with a scrollabe menu and added the "Created with <3 by the SUSE team" at the bottom:

screencapture 2

screencapture 3

And in small screen I put the menu in a dropdown:

screencapture 4

screencapture 5

I hope that you enjoy it! :)

### DIFF
--- a/_includes/desktop-topics.html
+++ b/_includes/desktop-topics.html
@@ -1,0 +1,40 @@
+<aside class="hidden-xs hidden-sm content-list">
+  <ul class="topics">
+    <li>Documentation</li>
+    <ul class="subtopics">
+      {% assign docs = (site.docs | sort: 'order') %}
+      {% for subtopic in docs %}
+        <li>
+          <a href="{{ subtopic.url | prepend: site.baseurl }}">
+            {{ subtopic.title }}
+          </a>
+        </li>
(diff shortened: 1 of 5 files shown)


mssola:
That looks perfect. Thanks !

mssola:
Argh, it looks like I merged too fast :sweat_smile:. There's a regression here for example: http://port.us.org/features/1_Synchronizing-the-Registry-and-Portus.html (the titles have lost their style).

mssola:
@msfernandes nevermind. As always, this was due that our setup could be better :sweat_smile:. Basically, when you add a new image, modify an asset, etc., you have to also include the modification from the `build` directory. I've pushed a fix to it (see [commit](https://github.com/SUSE/Portus/commit/a92877ff564647b4384afac6dc2a0409d863569c)).

Let us both remember about this next time :sweat_smile:

PR #953, opened 2016-07-06
Assets structure refactor

msfernandes:
Hi!
First, this PR is still in progress, but I would to know your opinion about that :)

Well... This PR is a new proposal to serve Portus assets (CSS and JS), using controller specific assets.
By now, I reorganized the `application.scss`, with only common styles (to the most pages) and added the necessary controllers specific SCSS, and each of them imports the `application.scss`. Finally, in the `application.html.slim` I added a condition to import the stylesheets: if the controller specific asset exists, import it, else, import `application.scss`.

And the idea is do the same with JS files...

What you think?


config/initializers/assets.rb
```diff
@@ -1,2 +1,11 @@
...
 Rails.application.config.assets.precompile += %w( auth/cover.js *.woff2 )
+Rails.application.config.assets.precompile += %w( admin/dashboard.css )
+Rails.application.config.assets.precompile += %w( dashboard.css )
+Rails.application.config.assets.precompile += %w( namespaces.css )
+Rails.application.config.assets.precompile += %w( repositories.css )
+Rails.application.config.assets.precompile += %w( teams.css )
+Rails.application.config.assets.precompile += %w( comments.css )
+Rails.application.config.assets.precompile += %w( search.css )
+Rails.application.config.assets.precompile += %w( errors.css )
+Rails.application.config.assets.precompile += %w( authentication.css )
```
mssola: I think that it can be simplified as:

```ruby
Rails.application.config.assets.precompile += %w( *.css )
```

app/views/layouts/application.html.slim
```diff
@@ -2,7 +2,10 @@ doctype html
 html
   head
     title Portus
-    = stylesheet_link_tag 'application', media: 'all'
+    - if Rails.application.assets.find_asset(params[:controller])
+      = stylesheet_link_tag params[:controller], media: 'all'
+    - else
+      = stylesheet_link_tag 'application', media: 'all'
```
mssola: I understand that this way is safer, but what if we just assumed that there was *always* a controller asset ? This would even make this simpler and it would remove an `if` statement on such a heavily used page. Yes, we would be quite explicit on controllers with no style whatsoever, but I think that it would be better for code clarity actually. What do you think ?

mssola:
All in all, this looks really good. Thanks for taking care of this :clap:

tboerger:
I don't think that this is a good idea. A single asset will be cached after first load, with that you have to reload the base content over and over again.

mssola:
> I don't think that this is a good idea. A single asset will be cached after first load, with that you have to reload the base content over and over again.

This is something worth testing because I'm not sure that this statement is 100% true either :grin:. The current situation is:

- Clients will get all the files, and it has happened more than once already that conflicts may arise, and thus becoming hard to debug bugs.
- Clients will get contents for a page that they really don't care anyways.

Admittedly, I don't have a deep knowledge on the asset pipeline, but I'd assume that the differently generated assets would be cached anyways once they get requested. I don't think that it's such a big trouble to cache again the base content for each separate asset on first use.

But as I said, it should be properly tested after the PR is ready to be reviewed.

app/views/layouts/application.html.slim
msfernandes: In most cases the asset loaded is the `application.css` and all controller assets needs to import `application.scss` because of some variables and classes defined there. So, I thought that will be better add this `if` than repeat only this import in a "empty" file (and for each controller).

msfernandes:
Maybe we can refactor the `application.scss`, removing all variables and classes (used in all `scss` files) and put them in another file, that will be imported in all controller assets and `application.scss`.

With this way, we can always import both `application.css` and controller specific `css`, because the controller assets will not need to import `application.scss` anymore. Also, the `application.scss` will be cached in the first access and the controllers scss when exists.

msfernandes:
@mssola @tboerger 
Hi!
I updated the PR. Now `application.css` and `controller.css` are imported. I did the things that I said on my previous comment. What you think?

mssola:
@msfernandes looks good to me. Now the only thing left is to add files for each controller (that's why tests are failing :wink:)

msfernandes:
Ok! I'll create the files for each controller and see js files to do the same thing

mssola:
Cool, so now tests are passing with the new layout for stylesheets. Now it's time for JS :wink:

msfernandes:
I was thinking...
The most part of JS will be served in `application.js`, so we'll have many empty js files being requested.
Which costs less: render the template with an `if` statement (to request the asset only when exists) or always request the asset (an empty file have more or less 250 bytes)?
Maybe the best solution will be always request the application asset and request the controller assets only if it exists..
What you think?

mssola:
> Maybe the best solution will be always request the application asset and request the controller assets only if it exists..

I think that this is the best idea.

tboerger:
And why are we doing it for css differently?

mssola:
> And why are we doing it for css differently?

That's true. It looks like CSS has ended up in the same situation (having lots of empty CSS files). We should re-consider on this and also apply the same rule for CSS.

msfernandes:
@tboerger I thought this solution while working on JS files... After finish with them I'll do the same with the CSS :)

tboerger:
I still got the opinion that all those separate files are overhead we don't need. I would prefer to reduce the css and optimize the js to load only single files for the hole app

config/initializers/assets.rb
```diff
@@ -1,2 +1,4 @@
 Rails.application.config.assets.version = "1.0"
-Rails.application.config.assets.precompile += %w(auth/cover.js *.woff2)
+Rails.application.config.assets.precompile += %w(*.woff2)
+Rails.application.config.assets.precompile += %w(*.css)
+Rails.application.config.assets.precompile += %w(*.js)
```
mssola: Nitpicking here, but why not simply?

```ruby
Rails.application.config.assets.precompile += %w(*.js *.css *.woff2)
```

mssola:
Besides my nitpick, it looks good to me. Could you fix my comment and squash your commits ?

msfernandes:
@mssola done!

mssola:
> I still got the opinion that all those separate files are overhead we don't need. I would prefer to reduce the css and optimize the js to load only single files for the hole app 

Reducing the CSS and the JS of Portus doesn't look doable for the next version. Bear in mind that the vast majority of CSS and JS is vendor (bootstrap, jQuery, etc.). If you debug it on a browser, you'll notice that the Portus' frontend code is nothing compared to that. Moreover, for context, take into consideration that "extra files" are just one more JS and CSS files on some pages (out of a gazillion of other JS and CSS coming from bootstrap + jQuery + others). Everything is properly cached, and at least by using it on my dev machine I cannot see any performance penalties (although we might want to do more research on performance in general (not just on the frontend) in the future :wink:). In my opinion, having everything in a single file makes sense for SPA, which I don't think is a suitable model for Portus (it's a small-medium sized web app and it is not heavy on JS really, only in the namespaces#show and the teams#show pages we do "fancy" stuff with JS).

Therefore, for now I think that this is the right way to handle the problems we have for now (unorganized JS/CSS that keeps clashing with other JS/CSS code). We might want to investigate this further in the future, but I think that we are good to go for now.

@msfernandes thanks for taking care of this :clap:

mssola:
@msfernandes I'm experiencing a regression for which the "Push", "pull", "owner" circles in the namespaces/teams page are no longer shown properly. Could you look into that ?

msfernandes:
@mssola it's working on my environment... Can you disable cache on you browser or reload the page with `ctrl + shift + r`?

mssola:
It's working now. Sorry for the noise

PR #961, opened 2016-07-07
update digests when updating tags

### LINKED ISSUE #826: Update digests when updating tags
If a user pushes a tag that already exists in Portus, Portus should update the digest if it has changed.

monstermunchkin:
If an already existing tag is pushed again, the digest is updated.

Fixes #826

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -135,11 +135,14 @@ class Repository < ActiveRecord::Base
     return if actor.nil?
 
     # Get or create the repository as "namespace/repo". If both the repo and
-    # the given tag already exists, return early.
+    # the given tag already exists, update the digest and return early.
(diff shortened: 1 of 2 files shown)


monstermunchkin:
I'm not quite sure about this. In theory it should work, and the test also succeeds. However, it didn't seem to work in a live environment.

mssola:
> I'm not quite sure about this. In theory it should work, and the test also succeeds. However, it didn't seem to work in a live environment.

It works for me :sweat_smile:. What I did:

1. Update a digest from a tag manually (through DB access).
2. Push again the same tag, and the DB gets updated.

I see no issues with the code, so if you are confident about it let's merge it, what do you think ?

monstermunchkin:
Awesome, thanks for testing! :smile: Feel free to merge.

Issue #878, opened 2016-05-16
newly successfully pushed images still not appear in Recent activities feeds.


nancykyo:
#version: Portus-2.0.1
also could not find error msg in log/production.log
![image](https://cloud.githubusercontent.com/assets/1332885/15279522/cc1c1b74-1b57-11e6-875f-187b003f29ba.png)
![image](https://cloud.githubusercontent.com/assets/1332885/15279577/68ad343c-1b58-11e6-8d2d-056c6d01743c.png)

here is the log before, this seems to be the normal behavior.
![image](https://cloud.githubusercontent.com/assets/1332885/15279465/3a8cae62-1b57-11e6-83ec-b8c4cd57d647.png)

how can I debug this to find out what's wrong?

nancykyo:
the portus interface return 301
![image](https://cloud.githubusercontent.com/assets/1332885/15281725/5c938774-1b70-11e6-9979-dc53545cba05.png)

nancykyo:
updates:
after I sync manually, the images can be searched.
but newly pushed images still not appear in **Recent activities** feeds.
![image](https://cloud.githubusercontent.com/assets/1332885/15282922/298d2c8c-1b79-11e6-8bb2-0a0afc75b93a.png)

mssola:
Which version of Docker are you using, and which version of docker distribution ? As stated in the [README](https://github.com/SUSE/Portus#supported-versions) file, not all versions are supported in 2.0.1. Maybe you should be considering updating to 2.0.4 ?

boyand:
I have the same problem and I am using Docker 1.9.1, Portus 2.0.5, Registry 2.4.0

mssola:
@boyand could you provide more information on this ? I'm asking because this is one of those issues that are really hard to debug because there are a lot of factors that can trigger this issue, and it may very well be a problem in your current setup (not a bug in Portus itself).

jovandeginste:
I did some more debugging and in the (current, 8532207a390d2ddb73cff59cf6385e31cab63210) code, after changing this I saw every (new) push to an existing tag in the activity feed:

```diff
diff --git a/app/models/repository.rb b/app/models/repository.rb
index 0491214..057bc35 100644
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -142,7 +142,13 @@ class Repository < ActiveRecord::Base
     elsif repository.tags.exists?(name: tag)
       # Update digest if the given tag already exists.
[...]

PR #963, opened 2016-07-07
This will update existing tags, including the "recent activities"

jovandeginste:
Just updating the tag will not update recent activities (why are we using created_at there instead of updated_at?)

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -141,8 +141,9 @@ class Repository < ActiveRecord::Base
       repository = Repository.create(namespace: namespace, name: repo)
     elsif repository.tags.exists?(name: tag)
       # Update digest if the given tag already exists.
-      _, digest = Repository.id_and_digest_from_event(event, repository.full_name)
-      repository.tags.find_by(name: tag).update!(digest: digest)
+      id, digest = Repository.id_and_digest_from_event(event, repository.full_name)
+      repository.tags.find_by(name: tag).destroy!
+      repository.tags.create!(name: tag, author: actor, digest: digest, image_id: id)
       return
     end
 
@@ -193,11 +194,10 @@ class Repository < ActiveRecord::Base
     repository = Repository.find_or_create_by!(name: name, namespace: namespace)
     tags = repository.tags.pluck(:name)
 
-    to_be_created_tags = repo["tags"] - tags
     to_be_deleted_tags = tags - repo["tags"]
 
     client = Registry.get.client
-    to_be_created_tags.each do |tag|
+    repo["tags"].each do |tag|
       # Try to fetch the manifest digest of the tag.
       begin
         id, digest, = client.manifest(repository.full_name, tag)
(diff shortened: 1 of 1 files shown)


mssola:
Since you actually destroy the tag, the id gets lost. Therefore, previous activities will lose the information of which tag had been pushed. An image:

asd

These two pushes are for the same image:tag, and it's a bit weird that the information of the tag that had been pushed previously is now gone. Maybe it would be a better idea to just create a new activity for that tag (something like "Updated the metadata of ...").

app/models/repository.rb
```diff
@@ -206,8 +206,19 @@ def self.create_or_update!(repo)
         digest = ""
       end
 
-      Tag.create!(name: tag, repository: repository, author: portus, digest: digest, image_id: id)
-      logger.tagged("catalog") { logger.info "Created the tag '#{tag}'." }
+      if current_tag = repository.tags.find_by_name(tag)
```
mssola: Why? I mean, we were using the `to_be_created_tags` array because it contained tags that we were sure didn't exist, therefore there is nothing to update, only to create. Do you have a use case in which this is not true ? As in, maybe we've got a bug on the `to_be_created_tags`, is that what you are saying ? :smile:

jovandeginste: When I push a new version of `mynamespace/myimage:latest`, I don't see it appear in the "recent activities", nor do I see the digest id being updated. Maybe I'm missing something?

mssola:
Moreover, check out our `CONTRIBUTING.md` file. In your case, check the [Mind the Style](https://github.com/SUSE/Portus/blob/master/CONTRIBUTING.md#mind-the-style), the [Provide tests](https://github.com/SUSE/Portus/blob/master/CONTRIBUTING.md#provide-tests) and the [Sign your work](https://github.com/SUSE/Portus/blob/master/CONTRIBUTING.md#sign-your-work) sections. Hope you don't mind, but every contributor (including me :grin:) has to follow these rules :wink:

jovandeginste:
@mssola why would I take the trouble of "signing my work" (which looks like much more work than just adding a line at the end)?

jovandeginste:
@mssola OK I'm not sure this can be fixed 'easily'. You lose 'previous activities' on the tag, but in fact the tag is overwritten (so the previous activities are not per se relevant). Without this PR, you never get new activities (or do you ?? if so, I'm missing something on my side!) which is IMHO worse... When I update an existing image:tag, the digest was never updated in portus (not by the callback, not by the cronotab)

Ideally and if I'm not mistaken, there should be a "taglog" which contains anything that happens to any namespace/repo:tag in time (and only gets added to). This will remain even if a tag is actually destroyed on the repo server.

jovandeginste:
I'll open it as an issue instead - easier for everyone

jovandeginste:
I think the issue I'm experiencing is actually #878; I will post there.

Issue #676, opened 2016-01-22
Enable/disable team/namespace creation for normal users


lonewulf:
In my opinion, it should be possible to disable team/namespace creation for normal users. 
Admins should be able to create teams/repos on behalf of normal users.

Eventually it should also be also be possible to disable personal repos.

mssola:
> In my opinion, it should be possible to disable team/namespace creation for normal users.
Admins should be able to create teams/repos on behalf of normal users.

I agree

> Eventually it should also be also be possible to disable personal repos

I'm not so sure about this though. @flavio what do you think ?

flavio:
I agree with the 1st point (disable team/namespace creation for normal users).

About the 2nd point, I think the easiest way to do that would be to disable a certain user. This is already implemented. @lonewulf would that work? If not can you provide more details about your usecase?

lonewulf:
Hello,

Well, disabling the user would disable his repo, though it's not what I meant.

The thing is we (admins) want to control the namespaces and teams so we can be sure of who has access to what, we have regulars audits on that point.
So if a user needs a personal repository, admins would enable it for that user and decide if it's public or not.

We come back to the first point of the initial post, a need for admins to be in full control of the access rights and the namespaces/repositories structure.

Regards,

lonewulf:
Hi @flavio 

Do you have any update on the second point ?

Regards,

flavio:
Sorry, the issue slipped though a lot of emails...

When a user logs into Portus for the 1st time a personal namespace is created. By default this namespace is private.

I think we should not change this work flow. However we could add an option that would not allow users to change the visibility (public/private) of their personal namespaces. Hence everybody would get a hidden read/write space (user win!), but only admins would be able to toggle the public flag (ops win!).

What do you think?

lonewulf:
It's a good compromise :)

So if I may sum it up:

- There should be an option that makes the teams/namespaces only editable by admins (on behalf of users) and that enables only the admins to toggle the public/private flag on those.
- There should be an option that enables only the admins to toggle the public/private flag on personal namespaces.

mssola:
@lonewulf @flavio I also think that it's a good compromise. Let's do this :)

monstermunchkin:
> There should be an option that makes the teams/namespaces only editable by admins (on behalf of users) and that enables only the admins to toggle the public/private flag on those.

Should this option be per-namespace and per-team, or just a global setting?

mssola:
[...]

PR #975, opened 2016-07-12
add option to disable change of visibility

monstermunchkin:
There is a new option 'user_change_visibility' which allows users to
change the visibility of their personal namespace. If this is option is
disabled, only admins can perform this change. This option is enabled by
default and therefore doesn't change the current behavior.

This commit is part of issue #676.


mssola:
LGTM. Just fix the complains by rubocop :wink:

monstermunchkin:
Done.

PR #979, opened 2016-07-13
use `enabled?` for nested options

monstermunchkin:
Allow the use of `enabled?` for nested options in the config file. Nodes
are separated with a period and only work for up to 2 nodes at the
moment.

Example:
`APP_CONFIG.enabled?("foo.bar")` instead of
`APP_CONFIG["foo"]["bar"]["enabled"].eql?(true)`


mssola:
This is something that I thought in the past but then forgot about it. Thanks! :clap:

PR #1003, opened 2016-07-21
introduce more optional user restrictions

### LINKED ISSUE #676: Enable/disable team/namespace creation for normal users

monstermunchkin:
Options regarding user permissions can be found under the
`user_permission` node in the config file.

The three options are:

* **change_visibility**
permits the user to change the visibility of namespaces he/she owns
* **modify_team**
permits the user to change team attributes, e.g. name or description,
given that he/she is an owners
* **modify_namespace**
permits the user to change namespace attributes, e.g. name, team, or
description, given that he/she is an owners

Note that the previous user restriction regarding namespace visibility
only applied to a user's personal namespace. This has been extended to
all namespaces the user owns.

Resolves #676

### DIFF
--- a/app/policies/namespace_policy.rb
+++ b/app/policies/namespace_policy.rb
@@ -40,20 +40,25 @@ class NamespacePolicy
     user.admin? || namespace.team.users.exists?(user.id)
   end
 
+  def create?
+    raise Pundit::NotAuthorizedError, "must be logged in" unless user
+    (APP_CONFIG.enabled?("user_permission.manage_namespace") || user.admin?) && push?
+  end
+
+  alias update? create?
   alias all? push?
-  alias create? push?
-  alias update? push?
 
   def change_visibility?
     raise Pundit::NotAuthorizedError, "must be logged in" unless user
-    (namespace.global? && user.admin?) || \
-      (!namespace.global? && (user.admin? || namespace.team.owners.exists?(user.id)))
+    user.admin? || (APP_CONFIG.enabled?("user_permission.change_visibility") &&
+                    !namespace.global? && namespace.team.owners.exists?(user.id))
   end
 
   # Only owners and admins can change the team ownership.
   def change_team?
     raise Pundit::NotAuthorizedError, "must be logged in" unless user
-    user.admin? || namespace.team.owners.exists?(user.id)
+    user.admin? || (APP_CONFIG.enabled?("user_permission.manage_namespace") &&
+                    namespace.team.owners.exists?(user.id))
   end
 
   class Scope
--- a/app/policies/team_policy.rb
+++ b/app/policies/team_policy.rb
@@ -15,8 +15,12 @@ class TeamPolicy
     user.admin? || @team.owners.exists?(user.id)
   end
 
+  def create?
+    APP_CONFIG.enabled?("user_permission.manage_team") || user.admin?
+  end
+
   def update?
-    !@team.hidden? && owner?
+    create? && !@team.hidden? && owner?
   end
 
   alias show? member?
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -49,6 +49,15 @@ RSpec.configure do |config|
       "jwt_expiration_time" => { "value" => 5   },
       "catalog_page"        => { "value" => 100 }
     }
+
+    APP_CONFIG["user_permission"] = {
+      # This allows non-admins to change the visibility of their personal namespace
+      "change_visibility" => { "enabled" => true },
+      # This allows non-admins to modify namespaces
+      "manage_namespace"  => { "enabled" => true },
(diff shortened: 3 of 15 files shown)


spec/spec_helper.rb
```diff
@@ -48,6 +48,14 @@
     APP_CONFIG["registry"] = {
       "jwt_expiration_time" => { "value" => 5   },
       "catalog_page"        => { "value" => 100 }
```
mssola: Missing comma. That's the compiler error you're getting :wink:

monstermunchkin: Actually, it should be a missing curly brace :wink:

mssola:
Moreover, you have to add options for portusctl. Bear in mind that they don't necessarily have to be named like `user-permission-change-visiblity`. Take a look at the `format_key` method in the `packaging/suse/portusctl/spec/options_spec.rb` file.

Other than that, everything looks good to me :+1:

monstermunchkin:
Oh right, totally forgot about those additions. Thanks!

spec/spec_helper.rb
mssola: Ah, sure. I was reading it as if it was all in the same hash :grin:

monstermunchkin:
Done.

Now, that I'm thinking about it, we should also hide the create/edit buttons from users if needed. Otherwise, these would only lead to confusion.

monstermunchkin:
This should be it.

app/policies/namespace_policy.rb
```diff
@@ -40,20 +40,25 @@ def index?
...
   alias all? push?
-  alias create? push?
-  alias update? push?
 
   def change_visibility?
     raise Pundit::NotAuthorizedError, "must be logged in" unless user
-    (namespace.global? && user.admin?) || \
-      (!namespace.global? && (user.admin? || namespace.team.owners.exists?(user.id)))
+    user.admin? || (APP_CONFIG.enabled?("user_permission.change_visibility") &&
+                    !namespace.global? && namespace.team.owners.exists?(user.id))
```
holgerreif: The message "must be logged in" as at least misleading (to be correct: it is wrong) in the case of prohibition by APP_CONFIG. Actually the user is logged in but he is not allowed to do so.

This applies to other code changes in this commit as well.

holgerreif:
In #676 there was a consence about

> In my opinion, it should be possible to disable team/namespace creation for normal users.
Admins should be able to create teams/repos on behalf of normal users.

From looking at the code I do not see, that this is accomplished with this commit (of course, I'm not much of a ruby expert)

app/policies/namespace_policy.rb
monstermunchkin: The message "must be logged in" is only shown if a user is not logged in. In case of prohibition by APP_CONFIG, the user should just get a 401. The Ruby code might look a bit confusing at first.

holgerreif: Ah, now I see. The raise statement ends with "unless user". Sorry for confusion and thanks for explaination.

monstermunchkin:
The options `modify_team` and `modify_namespace` take care of this. If they are disabled, users cannot manage (create/modify) any of these. This is checked in the `*_policy.rb` files. Does this help?

holgerreif:
> The options modify_team and modify_namespace take care of this. 

Thanks, I wasn't sure that creation is deemed a modifaction ;-)

We will test this (together with some other fixes) after it gets merged and an images is published on dockerhub.

monstermunchkin:
> Thanks, I wasn't sure that creation is deemed a modifaction ;-)

Good point. Perhaps we should rename them to something like `manage_team` and `manage_namespace`. On the other hand, the comments above both options clearly state what effect these options have.

mssola:
@monstermunchkin maybe renaming it to `manage_team` and `manage_namespace` is more clear. That being said, I don't have any strong feelings on this.

app/policies/team_policy.rb
```diff
@@ -15,8 +15,12 @@ def owner?
     user.admin? || @team.owners.exists?(user.id)
   end
 
+  def create?
+    APP_CONFIG.enabled?("user_permission.modify_team") || user.admin?
+  end
+
   def update?
-    !@team.hidden? && owner?
+    (APP_CONFIG.enabled?("user_permission.modify_team") || user.admin?) && !@team.hidden? && owner?
```
mssola:
```ruby
create? && !@team.hidden? && owner?
```

monstermunchkin:
The options have been renamed to `manage_team` and `manage_namespace`.

mssola:
Thanks :clap:

PR #1016, opened 2016-07-30
Closes #454 : Allow filtering of namespaces, teams and users

### LINKED ISSUE #454: Allow filtering of namespaces, teams and users
When Portus has tons of data it become pretty hard to go through it.

msfernandes:
**Work in Progress**
Closes #454 

Hi :)

I started to implement the dynamic filtering by the search page, and this is the current behavior:

www gifcreator me_ptzis5

I opened the PR before finish to receive some feedback about the presentation and implementation, also certify that is the correct behavior.

Cheers


app/controllers/search_controller.rb
```diff
@@ -1,8 +1,19 @@
...
+    @repositories = repositories.page(params[:page]).per(2)
+    @repositories_count = repositories.count()
+
+    teams = policy_scope(Team).search(search)
+    @teams = teams.page(params[:page]).per(2)
+    @teams_count = teams.count()
+
+    namespaces = policy_scope(Namespace).search(search)
+    @namespaces = namespaces.page(params[:page]).per(2)
+    @namespaces_count = namespaces.count()
```
mssola: I think that more than 2 items would be better. I guess that you set this value just for testing purposes though :wink:

app/controllers/search_controller.rb
```diff
@@ -1,8 +1,19 @@
 class SearchController < ApplicationController
   def index
     search = params[:search].split(":").first
-    @repositories = policy_scope(Repository).search(search)
-    @teams = policy_scope(Team).search(search)
-    @namespaces = policy_scope(Namespace).search(search)
+    @search_query = params[:search]
+    @search_type = params[:type]
```
mssola:
```ruby
params.require(:type)
```

The same for the above line.

app/views/search/index.html.slim
```diff
@@ -9,10 +9,14 @@
...
             = render template: "search/_mobile_categories.html.slim"
       .col-xs-12.col-md-8.col-lg-9.search.search-result
-        - case params[:type]
-          - when 'repositories'
-          = render template: "search/_repositories.html.slim"
-          -when 'namespaces'
-          = render template: "search/_namespaces.html.slim"
-          -when 'teams'
-          = render template: "search/_teams.html.slim"
+        .search-filter
```
mssola: This smells like something that you will be able to re-use in other places. Maybe it would be interesting to abstract it into a helper if you repeat yourself on this on other pages :wink:

app/views/search/index.js.erb
```diff
@@ -0,0 +1,14 @@
+$('#search_results').html('<%= escape_javascript(render("search/#{@search_type}")) %>');
+$('#teams_count').html(<%= @teams_count %>);
+$('#namespaces_count').html(<%= @namespaces_count %>);
+$('#repositories_count').html(<%= @repositories_count %>);
+
+function updateCategoryLink(selector) {
+    var new_href = $(selector).attr('href');
+    new_href = new_href.replace(/([\&?]search=)\w*(\&?)/g, '$1<%= @search_query %>$2');
+    $(selector).attr("href", new_href);
+}
```
mssola: This function looks like it can be re-used somehow when you apply this same idea in other places. Maybe you could create a `pagination_helpers.js` or something like that where you can put common functions.

app/views/shared/_header.html.slim
```diff
@@ -4,13 +4,15 @@
...
 
 .text-right.user-header
-  .search-header.hidden-xs
-    .header-search-form
-      = form_tag search_index_path, method: 'get', class: 'navbar-form' do
-        = search_field_tag 'search', params[:search], class: 'form-control search-field', placeholder: 'Search'
-        = hidden_field_tag 'type', 'repositories'
-        button[type="submit" class="btn btn-default"]
-          i.fa.fa-search
+  - if not current_page?(url_for controller: 'search')
```
mssola: This is not going to work whenever you want to extend this to other places. Maybe it would be a good idea to create a new helper for this.

mssola:
This looks like magic :tada:  I remember that we tried this but couldn't figure out how to keep pagination in place. Thanks a lot for taking care of this. The design looks good to me, so from my side feel free to extend this everywhere :+1:

app/views/shared/_header.html.slim
msfernandes: I added this conditional just to remove the search from the topbar on search page, because on the search page already have a search form (with the ajax filter)

app/views/search/index.js.erb
msfernandes: I did this function to update the link (dynamically updates the `search` get parameter) of the search result categories (repositories, namespaces and teams , I think this function won't be used in another context, because of his specificity

app/controllers/search_controller.rb
msfernandes: Ok!  Rails is a little bit new for me hahaha

app/controllers/search_controller.rb
msfernandes: just for testing :)

app/views/search/index.html.slim
msfernandes: yes! I'll work on it

msfernandes:
I have some doubts...
Which pages/tables I have to add the dynamic filter? 

Notice that the search is hidden in `search_controller`'s pages (just for prevent the user to user the form without the dynamic filter), and on the other pages, which place would be better to put the search form?

app/views/shared/_header.html.slim
msfernandes: The search form from topbar and small screens are a little bit different...
Maybe we can do a helper for the others places where the dynamic filter appears. What you think?

app/views/search/index.html.slim
msfernandes: This form has one hidden field (type) that won't be repeated on the others places with the dynamic field...
Maybe the helper would be better for these places. What you think?

app/controllers/search_controller.rb
msfernandes: With `params.require(:type)` the search broke when the `search` parameter is empty :(

mssola:
So, theoretically:

- The "Namespace you have access to" table in the namespaces#index page.
- The "Namespaces owned" in the teams#show page.

Those are the biggest offender iirc :grin:

app/views/shared/_header.html.slim
mssola: Makes sense to me.

app/views/search/index.html.slim
mssola: Or maybe the helper can have an optional parameter about this. But I don't know, I don't have any strong opinions here :wink:

app/controllers/search_controller.rb
mssola: mmm, I see, then use params[:search] as you did before

msfernandes:
Added the feature on the new pages:
www gifcreator me_ybuvnp

And some refactor....

mssola:
I guess that this also works with pagination in place right ? If so, then it generally looks good to me :+1: Besides that, tests are failing. Just fix those and we would be good to go :)

mssola:
You are probably going to kill me, but testing it I thought of some other pages in which this could also be applied :sweat_smile: :

- teams#index
- teams#show: members table
- admin_namespaces#index: other namespaces table
- admin_teams#index
- admin_users#index

msfernandes:
Hi!
I did some refactoring on dynamic filter to turn more generic and to allow more than one filter on the same page.

But even the DOM problem solved (before refactoring the helper generated the `<form>` code with hardcoded id's) I cannot figure out how to allow the more than one filter per page. On `teams#show`, when I type anything in one filter a search occurs on `namespaces` AND `team_users`, but one parameter always be empty because the request get only the current `form` parameters. Then `show.js.erb` is executed and all "filter results" are re-rendered with the wrong search result :(

Can you understand?

mssola:
@msfernandes I just try it out and it seems to work for me. Could you present a specific case in which this doesn't work ?

msfernandes:
@mssola On `teams#show`, try filter one table, then filter another table. When you filter the second, the second lost the filter

mssola:
Oh, now I see! One way to do it (don't know if it's the best way :grin:), would be to store in a cookie the last results of the filter. This way if `params[:filter_users]` is not set, then you can read it from the cookie. I'd advice to come up with a helper method for that :wink: 

Another thing to consider is if the user then goes to another page and finally comes back: the filter will be empty, but then the `show` method will remember the last filters. There are two ways of handling this:

- If the cookie is set, set it in the HTML too when loading the page. I don't like this idea though.
- If we are coming from another page, even if any of both cookies are set, we can know for sure that both params will be empty. Therefore, in the beginning of the method you can do something like:

```ruby
if params[:filter_users].blank? && params[:filter_namespaces].blank?
  # clear cookies
end
```

What do you think ?

msfernandes:
I don't like to use cookies this way, but I also can't think in another solution :joy: 

So, let me understand... 

* If `filter_user param` is empty, uses cookie as filter
* If `filter_namespaces param` is empty, uses cookie as filter
* If `filter_user param` AND  `filter_namespaces param` is empty, clear cookies
* Always set the cookies with `filter_user param` and  `filter_namespaces param` values

Did I understand right?

mssola:
Yeah :+1:

app/controllers/admin/namespaces_controller.rb
```diff
@@ -1,6 +1,10 @@
 class Admin::NamespacesController < Admin::BaseController
+  respond_to :html, :js
+
   def index
     @special_namespaces = Namespace.where(global: true)
-    @namespaces = Namespace.not_portus.where(global: false).page(params[:page])
+    namespaces = Namespace.not_portus.where(global: false)
+    @namespaces = namespaces.search(params[:filter]).page(params[:page])
```
shlomizadok: Won't this create two queries to the DB?

app/controllers/search_controller.rb
```diff
@@ -1,8 +1,19 @@
...
+
+    repositories = policy_scope(Repository).search(search)
+    @repositories = repositories.page(params[:page])
```
shlomizadok: Ditto, this will create two queries

app/views/namespaces/_namespaces.html.slim
```diff
@@ -0,0 +1,19 @@
...
+        tr
+          th Name
+          th Repositories
+          th Webhooks
+          th Created At
+          th Visibility
+      tbody#accessible-namespaces
+        - @namespaces.each do |namespace|
+          = render partial: 'namespaces/namespace', locals: {namespace: namespace}
+.panel-footer= paginate(@namespaces)
```
shlomizadok: Nitpick: missing end-of-line

app/views/search/index.html.slim
shlomizadok: +1 for a helper method to DRY out the code. The hidden field can be set via a parameter - something like:
(psuedo code)
```ruby
def form_helper_name(with_hidden = false)
  form_tag that_creates_hidden_field if with_hidden
end
```

(just my 2 cents :smiley: )

app/controllers/admin/namespaces_controller.rb
msfernandes: I'm not very experienced with rails... But in Django, the similar code create only one query on db...

shlomizadok: I believe it would. First time with `Namespace.not_portus.where` (1st where query) and second time with `namespaces.search` (the [search method](https://github.com/mrkamel/search_cop#how-does-it-work) is also a where query)
If they where on the same assignment, so Rails would know to unify them - but when are split - they would be called one after the other.

app/controllers/search_controller.rb
BBBernsteyn: Likewise,
```ruby
@repositories = policy_scope(Repository).search(search).page(params[:page])
``` 
will result in a single query and put less stress on the DBMS.

BBBernsteyn:
@msfernandes, just in case: To resolve the conflicts, [this SO post](http://stackoverflow.com/a/31863550) could serve as inspiration.

msfernandes:
Olá a todos, desculpe pelo atraso. O fim do GSoC coincidiu com o início do meu trabalho final de formatura e ontem o apresentei. Eu não gosto de deixar o meu trabalho ao meio, mas eu não tive tempo, desculpe :(

Eu vou continuar este PR agora!

msfernandes:
I tried use cookie, but didn't work :(
```
if params[:filter_users].blank? && params[:filter_namespaces].blank?
  # clear cookies
end
```
When you clear the input of any filter this `if` returns `true` and the cookies are cleared...

So I did using JS. I added the `extra_data` parameter to `dynamic_filter_input` helper. Then, when the user is typing on `filter_namespaces` the request sent contains `filter_users` too and vice versa.

msfernandes:
@mssola

app/controllers/admin/namespaces_controller.rb
mssola: @msfernandes could you add what @shlomizadok and @BBBernsteyn have mentioned ?

PI-Victor:
old PR, seems a lot of work went into it, would be a shame not to merge it if it's still up-to-date.

msfernandes:
Sorry @PI-Victor , I'm very ashamed because of this PR is still open, I have been focused on my graduation work and I don't have much time now :(

But I promise that I'll finish this PR, as soon as I can!

mssola:
Hey @msfernandes, thanks a lot for the work 👍 

Unfortunately though, we have been introducing quite some changes on the frontend of Portus for some time now:

- We are using modern Javascript (ES6, webpack, babel, etc.), and so we have removed CoffeeScript because we thought it didn't made sense anymore for us.
- We are using Vue, and so we'd prefer a component-based approach, instead of what we've been doing so far (including this PR).

Moreover, since it's been quite some time, there are conflicts, and a lot of stuff would have to be re-worked. Therefore, I will close this PR, but feel free to open a new PR with the proper code.

Thanks 👏

PR #1004, opened 2016-07-21
New FAQ page

msfernandes:
As part of #289 @mssola and I created a FAQ page with the initial questions :)

image

You can suggest questions too!

PR #1039, opened 2016-08-24
When the user click on question header the answer is shown

msfernandes:
Hi :)

I added a FAQ page in #1004 , but the answer was shown only clicking on the tab text... Now when the user click in any place of question header the answer is shown


mssola:
Agreed :+1:

mssola:
/me is dumb and thought when he saw this that it was an issue. Embarrassing... Merging :sweat_smile:

PR #1042, opened 2016-08-26
bcrypt is not a Rails dependency so it's absolutely required in Gemfile

guilleiguaran:
bcrypt stop being a hard dependency of rails since this [commit](https://github.com/rails/rails/commit/67790644372ad3a771810f1d6d99687d795789ea), so it's totally required here for application tokens.


mssola:
Thanks ! :clap:

Issue #1040, opened 2016-08-25
Make password length configurable


mssola:
I'm not entirely sure about this, but it was brought up in the mailing list (see [discussion](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/portus-dev/5jKp70sGqHE/vq05vWxOAAAJ)).

The gist of it is: in LDAP environments, passwords might not match the minimum length required by Portus (and Portus administrators might not be in charge of the LDAP server), so trying to login with Portus with valid LDAP credentials is simply not possible. Moreover, in this specific case, the Portus instance is run internally, so security concerns shouldn't be that important.

Of course, this would be really discouraged but it would solve this scenario. I'm not entirely sure about adding this, and that's why I'm adding the `discussing` label.

maksonlee:
Security is important, but if we decide to authenticate to ldap, maybe we should let ldap to check the password policy, portus should do the check only when using built-in authentication.

Mika56:
:+1: I'm currently having issues with this :)
I think that while Portus is right to enforce a password policy (that should be configurable) for internal users, I can't enforce one for external users, the LDAP directory should

BBBernsteyn:
When delegating authentication to LDAP it really stops being Portus' responsibility to enforce password policies. I dug around in `devise` and found the responsible validation in the `validatable` module [here](https://github.com/plataformatec/devise/blob/69bee06ceee6280b54304928bb6e55c5064abad8/lib/devise/models/validatable.rb#L35).

What do you think about disabling the `validatable` module in the user model (see gist [here](https://gist.github.com/BBBernsteyn/11b2e507da343ec0f191ab84f3f5698a)) in LDAP setups?

holgerreif:
I am wondering why password policy is enforced at logon time. Usually a policy is enforced at password generation/change time, eventually accompanied by the possibility for an admin user to override this policy.

BBBernsteyn:
I skimmed the documentation but to no avail. What if we asked the generous folks to give us devise to explain their rationale?

mssola:
Fixed with #1064

PR #1064, opened 2016-09-23
Skip validations of devise's validatable module in LDAP setups

BBBernsteyn:
Should solve #1040.
The mechanism is easy, the specs are tricky (Rubocop and Travis CI agree). Please, bear with me:

1. Once the user model class is loaded, `devise` configures it by mixing in its modules. In our case it mixes in the `validatable` module that contains the validation for the minimum length of a user's password (See [here](https://github.com/plataformatec/devise/blob/v3.5.1/lib/devise/models/validatable.rb#L35)). We have no way to to get rid of this validation during runtime. 
2. In case of an LDAP setup we want devise to leave out the `validatable` module because its validation of password length might be too strict  (#1040). 
3. In the specs the user model has already been loaded. It is not possible to enable the LDAP setting and _renew_ the class definition without a trick. This is done by reloading the user model class into a separate module (_the clean room_), where it can coexist with the original user model without side effects. 

In practice, we know beforehand if an LDAP setup is present or not and this is not subject to change during runtime - so this shouldn't be a problem.

### DIFF
--- a/spec/features/auth/login_feature_spec.rb
+++ b/spec/features/auth/login_feature_spec.rb
@@ -34,6 +34,29 @@ feature "Login feature" do
     expect(page).to_not have_content("The first user to be created will have admin permissions !")
   end
 
+  scenario "Skips validation of minimum password length when authenticating via LDAP" do
+    APP_CONFIG["ldap"] = { "enabled" => true }
+
+    # Skipping validation for LDAP users is configured when the user model is first interpreted
+    # Use a clean room to guard against side effects
+    module CleanRoom
+      # rubocop:disable Lint/Eval
+      eval File.read(File.join(Rails.root, "app", "models", "user.rb"))
+      # rubocop:enable Lint/Eval
+    end
+
+    ldap_user = CleanRoom::User.first
+    ldap_user.password = "short"
+    expect(ldap_user.save).to eql(true)
+
+    fill_in "user_username", with: ldap_user.username
+    fill_in "user_password", with: ldap_user.password
+    click_button "Login"
+
+    expect(page).to have_content("Recent activities")
+    expect(page).to have_content("Repositories")
+  end
+
   scenario "Existing user is able using his login and password to login into Portus", js: true do
     expect(page).to_not have_content("Invalid username or password")
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -36,8 +36,12 @@
 class User < ActiveRecord::Base
   include PublicActivity::Common
 
-  devise :database_authenticatable, :registerable, :lockable,
-         :recoverable, :rememberable, :trackable, :validatable, authentication_keys: [:username]
+  enabled_devise_modules = [:database_authenticatable, :registerable, :lockable,
+                            :recoverable, :rememberable, :trackable, :validatable,
+                            authentication_keys: [:username]]
+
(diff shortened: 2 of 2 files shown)


BBBernsteyn:
> The use of eval is a serious security risk.

Rubocop, you need to relax:
I wholeheartedly agree when `eval` is used in the productive coding in an _application_; here it's being useful in the _specs_.

.rubocop.yml
```diff
@@ -39,3 +39,8 @@ AllCops:
     - db/schema.rb
     - db/migrate/*
     - vendor/**/*
+
+# Let eval work its magic in the specs
+Lint/Eval:
+  Exclude:
+    - 'spec/**/*'
```
mssola: Instead of doing this, I'd prefer to write inline comments on the special offending line (we do it in other places :grin:)

mssola:
Besides that nitpick, it looks good to me. What do you think @tboerger and @monstermunchkin ?

tboerger:
I don't really see why we need eval in that case but if @mssola is fine with that you get my LGTM

spec/features/auth/login_feature_spec.rb
```diff
@@ -34,6 +34,27 @@
     expect(page).to_not have_content("The first user to be created will have admin permissions !")
   end
 
+  scenario "Skips validation of minimum password length when authenticating via LDAP" do
+    APP_CONFIG["ldap"] = { "enabled" => true }
+
+    # Skipping validation for LDAP users is configured when the user model is first interpreted
+    # Use a clean room to guard against side effects
+    module CleanRoom
+      eval File.read(File.join(Rails.root, "app", "models", "user.rb"))
```
mssola: @tboerger made me think. It should be possible to avoid `eval` by doing the following no ?

```ruby
module CleanRoom
  User = ::User
end
```

BBBernsteyn: Unfortunately this is not enough; Ruby will simply reuse the existing User model - we need it to interpret the model class again; that's why we need a sprinkle of `eval` magic here...

.rubocop.yml
BBBernsteyn: Can you give me a hint on that inline comment? Then I'll just revert delete the offending commit and use that.

BBBernsteyn: Done via c14b777466b1bf3cd581d9d5da29f19336062313.

spec/features/auth/login_feature_spec.rb
BBBernsteyn: @mssola, if that is OK with you, I'd close this pull request so you can go ahead and merge.

mssola:
Could you squash your commits into a single one and sign it off ? (git commit's `-s` flag). Thanks!

BBBernsteyn:
@mssola: Aye, done.Closes #1040.

BBBernsteyn:
Closes #1040.

mssola:
@BBBernsteyn why did you close this ? :confused:

BBBernsteyn:
Sorry @mssola, I was getting ahead of myself.

mssola:
Thanks :clap:

BBBernsteyn:
Was a pleasure! :bowtie:

PR #1065, opened 2016-09-25
Allow us to turn off smtp authentication

jmahowald:
In a minimal docker compose type setting it's useful to have smtp turned on and to let the relay handle all the authentication.  Otherwise, you'll have errors about 

`
portus_1       | Net::SMTPSyntaxError (503 AUTH command used when not advertised
portus_1       | ):

`

if using a basic smtp image like namshi/smtp

### DIFF
--- a/config/initializers/smtp.rb
+++ b/config/initializers/smtp.rb
@@ -3,17 +3,25 @@
 
 if defined?(APP_CONFIG) && APP_CONFIG["email"]["smtp"]["enabled"]
   Portus::Application.config.action_mailer.delivery_method = :smtp
-
   smtp = APP_CONFIG["email"]["smtp"]
-  ActionMailer::Base.smtp_settings = {
+  smtp_settings = {
     address:              smtp["address"],
     port:                 smtp["port"],
-    user_name:            smtp["user_name"],
-    password:             smtp["password"],
     domain:               smtp["domain"],
-    authentication:       :login,
-    enable_starttls_auto: true
+    enable_starttls_auto: false
   }
+  if smtp["user_name"].blank?
+    Rails.logger.info "No smtp username supplied, not using smtp authentication"
+  else
+    auth_settings = {
+      user_name:            smtp["user_name"],
+      password:             smtp["password"],
+      authentication:       :login,
+      enable_starttls_auto: true
+    }
+    smtp_settings = smtp_settings.merge(auth_settings)
+  end
+  ActionMailer::Base.smtp_settings = smtp_settings
 else
   # If SMTP is not enabled, then go for sendmail.
   Portus::Application.config.action_mailer.delivery_method = :sendmail


config/initializers/smtp.rb
```diff
@@ -3,17 +3,26 @@
...
-    user_name:            smtp["user_name"],
-    password:             smtp["password"],
-    domain:               smtp["domain"],
-    authentication:       :login,
-    enable_starttls_auto: true
+    port:                 smtp["port"]
   }
+
+
+  if(!ENV['PORTUS_EMAIL_SMTP_AUTHENTICATION_DISABLED'])
```
tboerger: You should use APP_CONFIG for that

mssola:
Why don't you simply disable SMTP with an environment variable as explained [here](http://port.us.org/docs/Configuring-Portus.html#override-specific-configuration-options) ? :confused:

jmahowald:
I want to turn off SMTP Authentication, not SMTP itself.  

When I'm running portus in a container, or at least the one I was trying, there is no local smtp.  So sendmail was not working.  I decided to use what looked to be the most popular [smtp container on docker hub](https://hub.docker.com/r/namshi/smtp/).  

When using that, I encountered a relatively obscure [smtp authentication issue](
http://serverfault.com/questions/313562/exim-configuration-503-auth-command-used-when-not-advertised)  that seems to stem from issues with using the extended SMTP standard.  Rather than hack the exim4  configuration to allow for that (well, after trying for an hour), I tried to NOT authenticate by monkey patching and that worked.


A little more of a tangential background:

I was running portus in a container based off of a [minimal alpine portus](https://github.com/h0tbird/docker-portus) rather than the new [official portus container](https://hub.docker.com/r/opensuse/portus/)  because I couldn't correctly setup a new container that didn't have apache.  I'm having ELBs in some cases, and  [traefik](https://traefik.io/) in others terminate the SSL.  I did try to use the [minimal dockerfile you created](https://github.com/openSUSE/docker-containers/blob/master/derived_images/portus/docker/Dockerfile) but after failing with compiling some gems with ssl went to that alpine image.

All of that is to say that for all I know, using the officially blessed container might have allowed me to just use sendmail but otherwise I couldn't make it meet my needs.

config/initializers/smtp.rb
jmahowald: Sorry, should have been more clear in my original description, as I just wanted to start the conversation, make sure we were on the same page, and then make it right.  I did start off trying that, but then I would have also had to change config.yml I think, and at the time of the request was concerned with introducing breaking changes.

jmahowald: that being said, my change did apparently break things.  I should be able to update soon with that chagne as well.

mssola:
I think that a better solution would be:

- The administrator (in this case you), explicitly disables the authentication part (we can say that if you set the proper env. variables empty for the username and the password, then SMTP authentication won't be done).
- Portus checks in the SMTP initializer that even though SMTP is enabled, the authentication variables are empty, so this part will be skipped (we can also log a warning/info message).

This way we don't introduce new environment variables and the desired functionality will be accomplished. What do you think ?

> I was running portus in a container based off of a minimal alpine portus rather than the new official portus container because I couldn't correctly setup a new container that didn't have apache. I'm having ELBs in some cases, and traefik in others terminate the SSL. I did try to use the minimal dockerfile you created but after failing with compiling some gems with ssl went to that alpine image.

Yeah, we are still working on that. Hopefully we'll have something soon in this regard.

app/jobs/catalog_job.rb
```diff
@@ -13,7 +13,7 @@ def perform
         # the DB in an unknown state because of an update failure.
         ActiveRecord::Base.transaction { update_registry!(cat) }
       rescue StandardError => e
-        Rails.logger.warn "Exception: #{e.message}"
+        Rails.logger.warn "Exception: #{e.message}" + e.backtrace
```
mssola: I don't think this is needed

config/initializers/smtp.rb
```diff
@@ -3,17 +3,27 @@
...
-    user_name:            smtp["user_name"],
-    password:             smtp["password"],
     domain:               smtp["domain"],
-    authentication:       :login,
-    enable_starttls_auto: true
+    enable_starttls_auto: false
+
   }
+
+  if (smtp["user_name"].blank?)
```
mssola: Remove the parenthesis, they are not needed

config/initializers/smtp.rb
```diff
@@ -3,17 +3,27 @@
...
+  else
+    smtp_settings = smtp_settings.merge({
+      user_name:            smtp["user_name"],
+      password:             smtp["password"],
+      authentication:       :login,
+      enable_starttls_auto: true
+      })
+  end
+  ActionMailer::Base.smtp_settings = smtp_settings
+
```
mssola: Remove this blank line

mssola:
Rubocop is still complaining. Fix those issues and squash your commits :wink:

mssola:
The other failures seem unrelated to your change. So, simply squash your commits into one and sign it off. I'll merge your change then :wink:

jmahowald:
Trying to figure out the mechanics of  that.  I guess close this pull request and open new one with squashed commits?  Or is there some way I'm unaware of where I can squash commits in here after the fact on this branch?

mssola:
Sorry for the trouble, but could you sign your commit off (`-s` flag) ?

jmahowald:
Figuring out how to sign commits was on my todo list anyway . .

All good now?

mssola:
Thanks :clap:

PR #1077, opened 2016-10-05
Fix portus:update_tags task.

Typz:
The task did not work, since it tried to get the manifest from the
repository name without the namespace.

### DIFF
--- a/lib/tasks/portus.rake
+++ b/lib/tasks/portus.rake
@@ -171,7 +171,7 @@ HERE
       puts "[#{index + 1}/#{tags.size}] Updating #{repo_name}/#{t.name}"
 
       begin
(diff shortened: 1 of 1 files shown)


lib/tasks/portus.rake
```diff
@@ -171,7 +171,7 @@ HERE
       puts "[#{index + 1}/#{tags.size}] Updating #{repo_name}/#{t.name}"
 
       begin
-        id, digest, = client.manifest(t.repository.name, t.name)
+        id, digest, = client.manifest(t.repository.namespace.name + "/" + t.repository.name, t.name)
```
mssola: Use `t.repository.full_name` instead.

mssola:
Thanks!

PR #1088, opened 2016-10-14
fix assets precompilation in the RPM (remove man pages and fix

jordimassaguerpla:
initialization)

We don't need the rake task for creating the man pages in the RPM.
Otherwise, running the assets precompilation fails.

Also, fix the initialization of the assets or the variables can't be
found when precompiling the assets.

### DIFF
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -1,2 +1,2 @@
 Rails.application.config.assets.version = "1.0"
-Rails.application.config.assets.precompile += %w(*.woff2 *.css *.js)
+Rails.application.config.assets.precompile += %w(*.woff2 *.js)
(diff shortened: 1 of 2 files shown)


config/initializers/assets.rb
```diff
@@ -1,2 +1,2 @@
 Rails.application.config.assets.version = "1.0"
-Rails.application.config.assets.precompile += %w(*.woff2 *.css *.js)
+Rails.application.config.assets.precompile += %w(*.woff2 *.js)
```
mssola: This conflicts with #1087, and I'd prefer to keep them separate

jordimassaguerpla:
I've created a PR just for the man pages alone because the other change is already on the mentioned pull request.

This is the PR for the removal of the man pages from the RPM

https://github.com/SUSE/Portus/pull/1089

PR #1094, opened 2016-10-18
Add registry port cmd

rodesousa:
In my use cases, I needed to run registry with port custom.
With this commit we can set registry port on compose-setup.sh


mssola:
LGTM. Could you squash your commits into a single one and sign it off ? (git commit's `-s` flag). Thanks!

rodesousa:
done

mssola:
Thanks :clap:

PR #1111, opened 2016-10-26
webhooks: hide passwords

monstermunchkin:

### DIFF
--- a/app/views/webhooks/_detail.html.slim
+++ b/app/views/webhooks/_detail.html.slim
@@ -21,6 +21,6 @@
         th Password
         td
           - if webhook.password.present?
-            = webhook.password
+            = '*' * 11
           - else
             ' &mdash;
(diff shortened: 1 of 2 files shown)


app/views/webhooks/_detail.html.slim
```diff
@@ -21,6 +21,6 @@
         th Password
         td
           - if webhook.password.present?
-            = webhook.password
+            = '*' * webhook.password.length
```
jdsn: Why expose the length of the password? I would use a static or random length.

app/views/webhooks/_detail.html.slim
```diff
@@ -21,6 +21,6 @@
         th Password
         td
           - if webhook.password.present?
-            = webhook.password
+            = '*' * 11
```
tboerger: And are you sure we never want to display it? Maybe make it toggable?

app/views/webhooks/_detail.html.slim
monstermunchkin: I doesn't really matter if it's a random or fixed length. The fixed length string however is really easy to implement.

PR #1120, opened 2016-10-28
prevent validation error in the specs

MaximilianMeister:
due to the randomness there is a chance that a number occurs multiple
times, which causes a validation error and the test to fail

Signed-off-by: Maximilian Meister mmeister@suse.de


mssola:
Good catch! :clap:

PR #1121, opened 2016-10-28
enhance the way of running local specs

MaximilianMeister:
now you can just run `bundle exec rake spec:models` after making some change to a model e.g. and the tests run much faster, as the successful activerecord logs aren't written anymore in the test env

you still have to provide the mysql test database and `phantomjs` though...


MaximilianMeister:
@mssola squashed

PR #1122, opened 2016-10-28
provide phantomjs via rubygem

MaximilianMeister:
this removes the obs dependency for the deb package and also
enables running local rspec without having to care about providing
phantomjs

Signed-off-by: Maximilian Meister mmeister@suse.de


MaximilianMeister:
failing travis will be fixed by https://github.com/SUSE/Portus/pull/1120

PR #1130, opened 2016-11-30
Fix notifications url

Evalle:
Right now if you will click on 'notifications' link on the http://port.us.org/features/1_Synchronizing-the-Registry-and-Portus.html you will get 404 page. This PR will fix it.


mssola:
Thanks :clap:

PR #1146, opened 2017-01-16
add patch management for making the spec file

jordimassaguerpla:
We need to add some patch management because otherwise we add the
patches later in the build service but then, if we rerun the script,
those patches are erased


mssola:
Shouldn't this be targeted to the v2.2 branch ?

jordimassaguerpla:
I would merge it on master and then cherry-picked to v2.2 ... what about it?

mssola:
Cherry-picked into v2.2.

PR #1148, opened 2017-01-18
add gem patches to the rpm

jordimassaguerpla:
otherwise the resulting Gemfile is different and this causes errors


mssola:
Gràcies 👏

PR #1180, opened 2017-02-13
packaging: a more generic make_spec.sh

MaximilianMeister:
providing the package name as an argument enables reusage for similar
projects

### DIFF
--- a/packaging/suse/make_spec.sh
+++ b/packaging/suse/make_spec.sh
@@ -1,4 +1,14 @@
 #!/bin/bash
+if [ -z "$1" ]; then
+  cat <<EOF
+usage:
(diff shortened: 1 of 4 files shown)


packaging/suse/make_spec.sh
```diff
@@ -36,11 +46,11 @@ additional_native_build_requirements() {
   fi
 }
 
-mkdir -p build/Portus-$branch
```
MaximilianMeister: @jordimassaguerpla do you have an idea if the capitalization is important here? the new version will render `build/portus-master` instead of `build/Portus-master`

jordimassaguerpla: I don't think it matters.

PR #1181, opened 2017-02-13
fix ownership for tmp/cache directory

jordimassaguerpla:
This should be owned by wwwrun:www but if this is an update, it is
owned by root. Thus, we need to force it to be owned by wwwrun:www

This follows up with the fix in 496976


mssola:
what about doing so for all `tmp`. I think there's also a `tmp/pids` for example.

jordimassaguerpla:
There is on only tmp/cache

PR #1189, opened 2017-02-15
[rpm] fix tumbleweed build

jordimassaguerpla:


jordimassaguerpla:
I applied this same fix in Virtualization:containers:Portus:2.2

PR #1203, opened 2017-02-17
packaging: fix patch name

MaximilianMeister:
the make_spec.sh script expects a gem in the name of the patch

follow up for #1202 

@mssola thanks for merging the patches, there was just this detail with the patch identifier in the name that was recently introduced in 89b118a97d775321a6d190239592985f77c30f40


mssola:
Got it, thanks 👍

PR #1205, opened 2017-02-21
assets: removed redundant bootstrap reference

mssola:
bootstrap-sprockects is the same as bootstrap, but with the files
splitted so it's easier to embed in sprocket's workflow.


jordimassaguerpla:
ok. I'll trust you with that :). lgtm

PR #1207, opened 2017-02-21
packaging: updated patch on gemfile

mssola:


mssola:
Merging myself since it's a tiny fix from a packaging patch.

PR #1210, opened 2017-02-22
Rely more on environment variables

mssola:
The secrets.yml file will rely more on environment variables, so it's
easier to deploy on Docker images. Morever, the RPM no longer needs to
remove the database.yml and secrets.yml files


jordimassaguerpla:
lgtm

PR #1211, opened 2017-02-22
packaging: fixed key-base new env. variable name

mssola:


mssola:
Merging myself since it's a small fix needed after my previous PR

PR #1212, opened 2017-02-22
Autofocus on registrations/new

vitoravelino:
After changing the email to become the second field in order, `autofocus` attribute was left there instead of added to the first field.


mssola:
Thanks 👏

PR #1213, opened 2017-02-22
team: description field cleared

vitoravelino:
Team description field now being cleared when adding a new team without reloading the page.

If the previous value was left uncleared on purpose, please do not consider this PR. I'm new to the project, just trying to contribute as much as I can. :)


mssola:
Thanks ! 👏

PR #1214, opened 2017-02-22
namespace: description and team cleared when form is displayed

vitoravelino:
Similar to https://github.com/SUSE/Portus/pull/1213 but on namespace's form.

### DIFF
--- a/app/assets/javascripts/namespaces.coffee
+++ b/app/assets/javascripts/namespaces.coffee
@@ -6,12 +6,11 @@ $(document).ready ->
     set_typeahead('/teams/typeahead/%QUERY')
 
   $('#add_namespace_btn').unbind('click').on 'click', (event) ->
-    $('#namespace_namespace').val('')
-
-    # When we are creating this on the namespaces page.
-    if $('#namespace_team') && $('#namespace_team').is(':visible')
-      $('#namespace_team').val('')
+    # clear only when creating a new namespace
+    if $('#new-namespace-form').length
+      $('#namespace_team.typeahead').val('')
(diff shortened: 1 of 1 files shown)


mssola:
Cool. Do you need any help with the tests ?

vitoravelino:
Fixed. I didn't remember rails generates id for hidden inputs too. Also, it's was bit confusing to understand the `if` condition since code runs every time and was being reused for different parts. 

I was hoping to maybe refactor a bit and make it more reliable on which code runs when a specific page is loaded. Not sure if is it worth it since I saw you mentioning about migrating to React/VueJS.

mssola:
> I was hoping to maybe refactor a bit and make it more reliable on which code runs when a specific page is loaded. Not sure if is it worth it since I saw you mentioning about migrating to React/VueJS.

On making that separation for each specific case, we already tried that once and failed miserably, and that's why we are considering more modern approaches for the frontend. That being said, we are still considering it, so take it with a grain of salt 😉 

Thanks for the PR 👍

vitoravelino:
@mssola 

That's sad to hear. I have a quick question: are you considering to move the front-end to become an independent SPA or just add a view layer on the js side to manage the components better? 

Not sure if you already have an answer, I'm just being curious.

mssola:
> That's sad to hear. I have a quick question: are you considering to move the front-end to become an independent SPA or just add a view layer on the js side to manage the components better?

The latter. We don't want to move to a full-fledged SPA (it doesn't make much sense in our case), but we'd like to introduce more modern approaches. So adding a view layer to manage the components better would be desirable.

Issue #1023, opened 2016-08-08
Extend version info on Web page


holgerreif:
Admin user gets version (like 2.1.0-dev) info at the lower right corner, normal user doesn't.

Admin user should get a detailed version (meaning for a -dev release the number and date of latest commit should be displayed as well) and the normal user should get the short version. 

Rationale: admin could easily check what actual version is deployed (esp. if pulled as a prebuild image), user could at least get a glue which online documenation is relevant to his version.

mssola:
> Admin user should get a detailed version (meaning for a -dev release the number and date of latest commit should be displayed as well) and the normal user should get the short version. 

This is already happening, no ? At least from current master I get `Version: master@5a698eac9d55`. Isn't that the same for you ? How are you deploying Portus. As far as I remember, 2.1.0-dev will only be shown if there is no git info on the deployed Portus.

> user could at least get a glue which online documenation is relevant to his version.

Agreed, we should allow regular users to check at least the current version.

holgerreif:
We are using the latest Image from dockerhub containng RPM with the last but one commit from master (last commit failed to build on OBS).

Admin gets the following displayed: ![4ff612e8beeddf5f](https://cloud.githubusercontent.com/assets/12717933/17488997/38626744-5d9c-11e6-925d-e17b61deec4d.png)

mssola:
I see, if I recall correctly we strip git info when producing the RPM, so that's why you are getting that version. I'll take a look at this too :wink:

PR #1216, opened 2017-02-22
#1023: showing current version for all users

vitoravelino:
Not sure if #1023 is only this change but kinda of it makes sense to display the current version to all users.


mssola:
Not exactly. The PR #1023 is about the version not being there because we strip it inside of the RPM. The change should be more on the lines of preventing the RPM build to do this (e.g. but saving a `.gitcommit` file containing the commit in which it was based or something like that).

cyntss:
lgtm

mssola:
Let's do this in another PR.

PR #1217, opened 2017-02-22
Removed unused bootstrap plugins

vitoravelino:
As you already know. `#= require bootstrap-sprockets` will include [all bootstrap plugins](https://github.com/twbs/bootstrap-sass/blob/master/assets/javascripts/bootstrap-sprockets.js). As far as I could investigate, Portus is using only 3 (tooltip, popover, tab) of them.

**Update:** this one different from the others has testing failing in all environments, so I probably broke something. :sweat_smile: I'll investigate and get back to this.

### DIFF
--- a/app/assets/javascripts/application.coffee
+++ b/app/assets/javascripts/application.coffee
@@ -1,6 +1,9 @@
 #= require jquery
 #= require jquery_ujs
-#= require bootstrap-sprockets
+#= require bootstrap/transition
+#= require bootstrap/tab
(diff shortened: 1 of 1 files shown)


mssola:
Nice catch! If you need any help with tests just tell me 😉

vitoravelino:
Fixed. Investigated and found that tooltip relies on the transition that wasn't included previously. 😄

mssola:
Thanks 👏

PR #1233, opened 2017-03-04
js: users module refactoring

vitoravelino:
So, this is something I've been thinking about for a while since I started contributing to Portus. After #1238 were merged, we are now able to write js using the new ES2015 syntax. This will improve the code readability and maintainability.

One of the coolest feature on ES2015 is the module support. The folder structure I'm proposing is something similar to:

```
javascripts/
    modules/
        (...)      
        namespaces/
        teams/
            components/
                form.js
                list.js
            pages/
                show.js
                list.js
            index.js
    application.js
modules
```

The `index.js` in the `teams`folder is simple and will be responsiable to know which which page component should be used on the loaded page. Example:

```js
import TeamsShow from './pages/show';
import TeamsIndex from './pages/index';

// we can move this to a separated file 
// and import as `import { TEAMS_SHOW...
const TEAMS_SHOW_ROUTE = 'teams/show';
const TEAMS_INDEX_ROUTE = 'teams/index';

$(() => {
  const $body = $('body');
  const route = $body.data('route');

  if (route === TEAMS_SHOW_ROUTE) {
    TeamShow.render($body);
  }

  if (route === TEAMS_INDEX_ROUTE) {
    TeamIndex.render($body);
  }

  (...)
```
I know that it's quite a poor technique but since we have the module system on our side, it guarantees that only the code involved on the page component will be executed.

A `BaseComponent` was created as an artifact to avoid code duplication and create some convention.

Suggestions are more than welcome. 😃


vitoravelino:
Tests are not passing because coverage got reduced. Any chance that we could expose the coverage report on coveralls or anything similar? 

I have problem testing portusctl and couldn't reproduce the same coverage problem.

app/assets/javascripts/auth/cover.js
```diff
@@ -1,8 +0,0 @@
-jQuery(function ($) {
-  var rndNum;
-
-  if ($('body section.sign-up, body section.login').length) {
-    rndNum = Math.floor((Math.random() * 2) + 1);
-    $('body').addClass('massive-background-' + rndNum).hide().fadeIn(1000);
```
vitoravelino: Any idea what this `massive-background-n` is/was used for? I didn't find anything.

vitoravelino:
Most of the work is done. I'd appreciate some feedback.

mssola:
@vitoravelino I'll review this PR later, but in principle it looks like a nice idea. In the meantime take a look at #1238. This is still WIP and it's something I've done during the weekend, and I think that it clashes a bit with your PR (not too much). My idea would be to get that one merged before yours, because the importing of modules will change (and also note that using webpack allows us to ignore the dependency on the `execjs` gem). After that we can review this one.

vitoravelino:
@mssola Oh, that looks great. I didn't consider webpack because I thought it would add another thing to the build process and that would be an issue. 

I'll take a look at your work. As soon as you merge your PR I'll rebase my branch and you can review it. Feeling excited for both changes. Woohooo! 😄

mssola:
@vitoravelino the PRs regarding webpack and ruby 2.4 have been merged. You can go ahead 👍

vitoravelino:
@mssola I've rebased and fixed the issues but there are some feature tests failing in travis but passing on localhost. It doesn't make sense. Any idea what could be? =/

vitoravelino:
Enabled `js_errors` and I got a clue. Working on it.

vitoravelino:
Finally figured out the problem. For some unknown (at least for me) reason `this.onKeyup.bind(this)` was throwing an error on travis but working in localhost. First, I tried to change it to use `$.proxy(this.onKeyup, this)`and it worked. Since we have ES2015 syntax support, I preferred to use `e => this.onKeyup(e)` instead because it's cleaner.

I know that the project is moving forward to use a library to manage the view layer but the structure is not gonna change much from what I've written here. 😃

app/assets/javascripts/modules/users/components/application-token-panel.js
```diff
@@ -0,0 +1,41 @@
+import BaseComponent from '~/base/component';
+
+const TOGGLE_LINK = '#add_application_token_btn';
+const TOGGLE_LINK_ICON = `${TOGGLE_LINK} i`;
+const APP_TOKEN_FORM = '#add_application_token_form';
+const APP_TOKEN_FIELD = '#application_token_application';
+
+class ApplicationTokenPanel extends BaseComponent {
```
mssola: Could you add documentation ?

app/assets/javascripts/modules/users/components/password-form.js
```diff
@@ -0,0 +1,40 @@
+import BaseComponent from '~/base/component';
+
+const CURRENT_PASSWORD_FIELD = '#user_current_password';
+const NEW_PASSWORD_FIELD = '#user_password';
+const NEW_CONFIRMATION_PASSWORD_FIELD = '#user_password_confirmation';
+const SUBMIT_BUTTON = 'input[type=submit]';
+
+class UsersPasswordForm extends BaseComponent {
```
mssola: Same here

app/assets/javascripts/modules/users/components/profile-form.js
```diff
@@ -0,0 +1,40 @@
+import BaseComponent from '~/base/component';
+
+const EMAIL_FIELD = '#user_email';
+const DISPLAY_NAME_FIELD = '#user_display_name';
+const SUBMIT_BUTTON = 'input[type=submit]';
+
+class UsersProfileForm extends BaseComponent {
```
mssola: ditto

app/views/devise/registrations/edit.html.slim
```diff
@@ -27,44 +27,45 @@
...
-    #add_application_token_form.collapse
-      = form_for :application_token, url: application_tokens_path, remote: true, html: {id: 'new-application-token-form', class: 'form-horizontal', role: 'form'} do |f|
-          .form-group
-            = f.label :application, {class: 'control-label col-md-2'}
-            .col-md-7
-              = f.text_field(:application, class: 'form-control', required: true, placeholder: "Name")
-          .form-group
-            .col-md-offset-2.col-md-7
-              = f.submit('Create', class: 'btn btn-primary')
+    .app-token-wrapper
```
mssola: Instead of wrapping the form, why don't you simply add this class into the form itself ?

vendor/assets/javascripts/lifeitup_layout.js
```diff
@@ -1,130 +1,130 @@
-// to render the layout correctly in every browser/screen
```
mssola: I'm not sure I understand what changed here 😅

vitoravelino: Me neither. Maybe because of the rebase? LOL. 😅

app/views/devise/registrations/edit.html.slim
vitoravelino: The idea is to have an element that contains both the link and the form. I know I could just add classes/ids to the elements and access from the global scope but restricting the component to access only the `this.$el` give us more control of what's going on and also a good practice.

Ideally, panel and form could be separated components and when user clicked on the link an event is triggered and the component that contains both panel and form will tell the form to show itself. What do you think? 

But yeah, wrapping them in a component to restrict dom element scope would still be needed.

app/assets/javascripts/modules/users/components/application-token-panel.js
vitoravelino: Is there any format/rule that I should follow?

vitoravelino:
Added some documentation. I would appreciate some guidance since I don't know what you are expecting.

mssola:
@vitoravelino you overdid it! :smile: So, the documentation should be one the same lines as the ruby code: human explanation of what it does, and when there is a parameter explain what is it expecting and all that. You can take [the LDAP class](https://github.com/SUSE/Portus/blob/master/lib/portus/ldap.rb) as reference :wink: So, avoid all the @ references and such, it should be simpler.

mssola:
Also, for comments I'd only use // instead of /**

vitoravelino:
KKKKK. Will do. 😅

vitoravelino:
@mssola Done. Let me know if it still needs improvement, please.

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["es2015"]
+}
```
mssola: I'm ok with this, but I think that we already provide the babel config through the package.json file no ? Why would we need this ? (asking just in case we can avoid duplication)

app/assets/javascripts/base/component.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable class-methods-use-this */
+
+//
```
mssola: You can remove this line and the one in the end. So, all comments can be like this:

```javascript
// BaseComponent class to avoid boilerplate code
// into component classes and add some convention.
//
// The subclasses are not obligated to implement all
// of these methods but if they need to perform an action
// that fits one of the descriptions below, it's highly
// recommended to use the proper method.
```

mssola: This same thing applies to the other classes, functions, etc.

app/views/devise/registrations/edit.html.slim
mssola: Got it. Thanks

vendor/assets/javascripts/lifeitup_layout.js
mssola: Then, rebase again and `checkout --` this file 😉

app/assets/javascripts/base/component.js
vitoravelino: 👍

.babelrc
vitoravelino: I didn't see the package.json config. I prefer having babel config on their own file. This is actually from my work of adding ES2015 support that forgot to delete it. 

I just realized that we might not need both of them since you already set 

```js
query: {
    presets: ['es2015'],
},
```

on the babel-loader config.

mssola: Maybe just the babelrc file.

vitoravelino:
No worries, @mssola. 😄

mssola:
Thanks 👏

Issue #1235, opened 2017-03-06
Events should be handled asynchronously to avoid deadlocks


codablock:
I ran into the same problem described here: https://github.com/SUSE/Portus/issues/940

As I understand it (based on my own debugging), the problem is that when events come in, the processing of these results in requests to the registry to get some additional information about the images being pushed. When this happens, the registry returns 401 to indicate a missing token. Portus then acts with an authorization attempt by doing a request to itself. This request however does not finish as for some reason Portus has to wait for the current request to finish, which is never happening as it's waiting for the auth request to finish.

I do not understand why Portus can not handle multiple requests at the same time here. I have setup puma to use multiple workers and threads, so it should be able to handle multiple requests. Maybe there is something in the code that causes database locking, but I'm not familiar enough with Ruby to debug this properly.

Nevertheless, my feeling is that the events processing should actually happen asynchronously to avoid any type of possible deadlock or request performance degradation. What should happen is:
1. Registry notifies Portus about events
2. Events endpoint of Portus gets called, entering the request handler
3. Add the event to something that is capable of background/asynchronously processing
4. Return from the event request handler immediately
5. Let the background processing do its work

An alternative idea to avoid the deadlock at least for now is to somehow call the token generation code directly instead of going through a HTTP request. If the deadlock is caused by a limitation on parallel requests, this would fix it. Also, if it's happening due to DB locks, this would probably also fix it as it could happen in the same transaction. This however feels like a dirty trick as it kind of ignores the registries auth settings.

codablock:
After some more investigation, I figured out that the manifest fetching was actually a bug. It tried to fetch the manifest even though the required info (digest) was already in the event. A fix to this is https://github.com/SUSE/Portus/issues/1236

Also, after reading through details how Ruby on Rails handles concurrency, I figured out that there are some connections between these issues and concurrency handling. As I understand it, RoR disables concurrency in development environments with the help of a mutes around all request handling, so that only one request can be served at the same time. I'm not sure how production is different in this behavior.  

Also, at least for development, caching of classes is disabled, resulting in the cached token property of the Registry model to being reused, increasing the number of needed authentication calls. So the behavior with these issues seen in development may be different from what you see in production.

Disclaimer: I'm not a Ruby/RoR coder, so my understanding of RoR internals may be wrong. I base my assumptions on the experience from other frameworks in other languages.

mssola:
This is fixed already. We've moved quite some tasks into the new background process. Thanks for the  feedback 👍

Issue #940, opened 2016-06-24
Breaking: Portus hangs for a very long time after image push


charles-m-knox:
Hello! 
I've deployed Portus using `compose-setup.sh` and it works perfectly. However there is a breaking issue that I have seen in every deployment of Portus that I've done, and would like to know what the issue is and how I can help fix it, if at all.

**The problem:**
After logging in via `docker login portus-site.site.com:5000`, and then subsequently pushing an image via `docker push portus-site.site.com:5000/mynamespace/image:latest`, the Portus `web` container appears to hang for a few minutes. Any and all HTTP requests sit around, waiting for a response. For the purposes of clear discussion, I'm going to refer to this down time as "halt time".

**Investigation**
- I ran weaveworks to see if there were any memory issues or CPU issues with the containers during halt time. There were no outstanding measurements.
- Weaveworks did show that there were ~3 processes running in the Portus web container, all of which appeared to be typical ruby processes. If needed, I can provide specific logs (I just simply don't know where to look, I am at your disposal here).
- I checked the host system's disk usage during the halt time with `iotop`. No outstanding measurements were present, disk usage was normal.
- Running `top` showed no issues either, most processes that were running were idle.

**Specific Logs**
At halt time, here are the logs for Portus web. Note that the last event that occurred is a DB `select` method.

```
Started POST "/v2/webhooks/events" for 172.17.0.1 at 2016-06-24 16:50:37 +0000
Cannot render console from 172.17.0.1! Allowed networks: 127.0.0.1, ::1, 127.0.0.0/127.255.255.255
Processing by Api::V2::EventsController#create as JSON
Incoming event:
{
  "id": "b2f4586a-4dc8-41f9-9d1d-c0a5d529a6f8",
  "timestamp": "2016-06-24T16:50:37.324040756Z",
  "action": "push",
  "target": {
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "size": 502,
    "digest": "sha256:e52be8ffeeb1f374f440893189cd32f44cb166650e7ab185fa7735b7dc48d619",
    "length": 502,
    "repository": "chuck/hello-world",
    "url": "http://portus-demo1.site.com:5000/v2/chuck/hello-world/manifests/sha256:e52be8ffeeb1f374f440893189cd32f44cb166650e7ab185fa7735b7dc48d619"
  },
  "request": {
    "id": "4120caf6-da29-4c0a-9aaf-6060228c0d22",
[...]

PR #1236, opened 2017-03-06
Fix check for missing digest in id_and_digest_from_event

codablock:
The check for blank digest was wrong. It has to execute the manifest
fetching in case it is blank and not when it is not blank.

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -160,7 +160,7 @@ class Repository < ActiveRecord::Base
     digest = event.try(:[], "target").try(:[], "digest")
     id = ""
 
-    unless digest.blank?
+    if digest.blank?
       begin
         id, = Registry.get.client.manifest(repo, digest)
       rescue StandardError => e


codablock:
This PR fixes one of the cases where Portus hangs on push (e.g. https://github.com/SUSE/Portus/issues/940). The reason was that even though incoming events already had a digest set, it still tried to get it from the registry, resulting in a deadlock due to authentication hanging (as described here https://github.com/SUSE/Portus/issues/1235)

codablock:
@mssola PTAL

I'm not sure why tests are failing with this PR. Maybe a test case was wrong before and thus always succeeded?

codablock:
@mssola Friendly PING :) I'm going through all my open PRs atm which are queued since some time already.

It's just a one-liner fix and it should be very easy to verify by only reading the code.

mssola:
Hey @codablock! Sorry for taking so much to reply...

This method is called whenever we got an event from the registry and we want to fetch the manifest for the given digest. Therefore, the digest should *not* be blank: it's fine the way it is. Try to trace the HTTP request sent to the registry in order to check what it looks like.

codablock:
Ahh sorry, I see. I mixed up the id and digest and didn't realize the method tries to fetch the id and not the digest. Closing this PR now.

I'm currently not working for the client that I supported with the Portus installation but will maybe continue to work for him in a few weeks. I'll try to debug further when doing so.

PR #1237, opened 2017-03-06
Honor external_hostname in token generation

codablock:
Otherwise token generation may fail in some cases.


codablock:
@mssola I fixed the logger issue.

Regarding the method in the Registry model...I'm really not a Ruby coder and everything that involves more then adding log entries and copy+pasting code is currently quite some effort for me :D Would it be ok if you accept it as it is right now and change it later in a way that is more acceptable?

mssola:
Fair enough. Merging this, I'll write a helper method later. Thanks 👍

Issue #861, opened 2016-05-06
signup: perform checks on the client side


mssola:
It's a bit cumbersome to allow users to enter whatever they want in the signup form, just to tell them they did it wrong afterwards. We should perform some client side checks (e.g. format of the `username`, that both the confirmation and the password match, etc.), and disallow the creation of the user if the requirements are not met.

flavio:
Maybe this can be addressed during the GSoC?

alexis19apl:
Hello good night mssola;

One of the full time engineer from Suse let me to this project to help and also learn, knowing how i reach this project want to know how can i contribute with you guys?

krishnak9:
Should we use the client_side_validation gem for this purpose or have a custom implementation for that. I am interested in doing this fix. Please help me in this regard. And what do we mean by the format of username exactly? I mean is there any predefined format for that.

mssola:
@krishnak9 this gem looks unnecessary. That being said, I'm not entirely sure what would be the best way to handle this. You could either do some validation entirely on the client side, or add a method on the registrations controller that the client code can call.

PR #1251, opened 2017-03-15
Client side validation of Signup form

krishnak9:
#861 This implements the client side validation to the signup page.

These are few screenshots:-

username
Submit button gets disabled when user enters an invalid entry.
email
screenshot from 2017-03-15 18 20 31
screenshot from 2017-03-15 18 21 47
screenshot from 2017-03-15 18 22 35
The submit button gets enabled after all fields are correct.
screenshot from 2017-03-15 18 28 44

### DIFF
--- a/app/assets/javascripts/modules/users/components/signup-form.js
+++ b/app/assets/javascripts/modules/users/components/signup-form.js
@@ -1,0 +1,55 @@
+import BaseComponent from '~/base/component';
+
+const PASSWORD_FIELD = '#user_password';
+const USERNAME_FIELD = '#user_username';
+const EMAIL_FIELD = '#user_email';
+const CONFIRMATION_PASSWORD_FIELD = '#user_password_confirmation';
+const SUBMIT_BUTTON = '#submit_btn';
+
+// UsersPasswordForm component that handles user password form
+// interactions.
+class UsersSignUpForm extends BaseComponent {
+  elements() {
+    this.$password = this.$el.find(PASSWORD_FIELD);
+    this.$passwordConfirmation = this.$el.find(CONFIRMATION_PASSWORD_FIELD);
+    this.$username = this.$el.find(USERNAME_FIELD);
+    this.$userEmail = this.$el.find(EMAIL_FIELD);
+    this.$submit = this.$el.find(SUBMIT_BUTTON);
+  }
+
+  events() {
+    this.$el.on('focusout', PASSWORD_FIELD, e => this.onFocusout(e));
+    this.$el.on('focusout', CONFIRMATION_PASSWORD_FIELD, e => this.onFocusout(e));
+    this.$el.on('focusout', USERNAME_FIELD, e => this.onFocusout(e));
+    this.$el.on('focusout', EMAIL_FIELD, e => this.onFocusout(e));
+  }
+
+  onFocusout() {
+    const usernameInvalid = !this.$username[0].validity.valid;
+    const emailInvalid = !this.$userEmail[0].validity.valid;
+    const passwordInvalid = !this.$password.val() || this.$password.val().length < 8;
+    const passwordConfirmationInvalid = !this.$passwordConfirmation.val() ||
+      this.$password.val() !== this.$passwordConfirmation.val();
+
+    if (passwordConfirmationInvalid) {
+      this.$passwordConfirmation[0].setCustomValidity('Please enter same password as above');
+    } else {
+      this.$passwordConfirmation[0].setCustomValidity('');
+    }
+
+    if (passwordInvalid) {
+      this.$password[0].setCustomValidity('Password should be of at least 8 characters minimum.');
+    } else {
+      this.$password[0].setCustomValidity('');
+      this.$passwordConfirmation.attr('pattern', this.$password.val());
+    }
+
+    if (passwordInvalid || passwordConfirmationInvalid || emailInvalid || usernameInvalid) {
+      this.$submit.attr('disabled', true);
+    } else {
+      this.$submit.removeAttr('disabled');
+    }
+  }
+}
+
+export default UsersSignUpForm;
--- a/app/assets/javascripts/modules/users/pages/sign-up.js
+++ b/app/assets/javascripts/modules/users/pages/sign-up.js
@@ -1,12 +1,20 @@
 import BaseComponent from '~/base/component';
 
 import { fadeIn } from '~/utils/effects';
+import SignUpForm from '../components/signup-form';
+
+const SIGN_UP_FORM = 'form.new_user';
 
 // UsersSignUpPage component responsible to instantiate
 // the user's sign up page components and handle interactions.
 class UsersSignUpPage extends BaseComponent {
+  elements() {
+    this.$signupForm = this.$el.find(SIGN_UP_FORM);
+  }
+
   mount() {
     fadeIn(this.$el);
+    this.signupForm = new SignUpForm(this.$signupForm);
   }
 }
 
--- a/app/views/devise/registrations/new.html.slim
+++ b/app/views/devise/registrations/new.html.slim
@@ -6,14 +6,14 @@ section.sign-up class=(@have_users ? '' : 'first-user')
       = image_tag 'layout/portus-logo-login-page.png', class: 'login-picture'
       = form_for(resource, as: resource_name, url: { action: 'create' }, :html => {:autocomplete => "off"}) do |f|
         = f.text_field :username, class: 'form-control input-lg', placeholder: 'Username', required: true, autofocus: true
-        = f.email_field :email, class: 'form-control input-lg first', placeholder: 'Email', required: true
-        = f.password_field :password, class: 'form-control input-lg', placeholder: 'Password (8 characters min.)', required: true
+        = f.email_field :email, class: 'form-control input-lg first', placeholder: 'Email', required: true 
+        = f.password_field :password, class: 'form-control input-lg', placeholder: 'Password (8 characters min.)', required: true, pattern: '^.{8,}$'
         = f.password_field :password_confirmation, class: 'form-control input-lg last', placeholder: 'Password again', required: true
 
         - if !@admin && @first_user_admin
           = f.hidden_field :admin, value: true
 
-        button.btn.btn-primary.btn-block.btn-lg type="submit"
+        button.btn.btn-primary.btn-block.btn-lg#submit_btn type="submit"
           i.fa.fa-check
           - if @admin || !@first_user_admin
             | Create account
--- a/spec/features/auth/signup_feature_spec.rb
+++ b/spec/features/auth/signup_feature_spec.rb
@@ -143,6 +143,30 @@ feature "Signup feature" do
                                   text: 'Password confirmation doesn\'t match Password')
   end
 
+  scenario "Submit Button gets disabled when any field is filled wrong", js: true do
+    visit "/users/sign_up"
+    wait_for_effect_on("#new_user")
+    fill_in "Username", with: user.username
+    fill_in "user_email", with: "gibberish"
+    fill_in "user_password", with: "12341234"
+    fill_in "user_password_confirmation", with: "532"
+    page.execute_script "$('#user_username').trigger('focusout')"
+    wait_for_effect_on("#submit_btn")
+    expect(page).to have_button("submit_btn", disabled: true)
+  end
+
+  scenario "Submit Button should be enabled when all fields are filled right", js: true do
+    visit "/users/sign_up"
+    wait_for_effect_on("#new_user")
+    fill_in "Username", with: user.username
+    fill_in "user_email", with: user.email
+    fill_in "user_password", with: user.password
+    fill_in "user_password_confirmation", with: user.password
+    page.execute_script "$('#user_username').trigger('focusout')"
+    wait_for_effect_on("#submit_btn")
+    expect(page).to have_button("submit_btn", disabled: false)
+  end
+
   scenario "If there are users on the system, and can move through sign_in and sign_up" do
     click_link("Login")
     expect(current_path).to eql(new_user_session_path)


app/assets/stylesheets/components/layout.scss
```diff
@@ -189,3 +189,14 @@ i.fa {
     border-right: 1px solid $primary-colour;
   }
 }
+
+//popover css used in signup page for client side validation
+
+#signup_page .popover{
+    background-color: rgba(0,0,0,0.6);
```
vitoravelino: Please use 2 spaces and spaces after the commas.

app/assets/stylesheets/components/layout.scss
```diff
@@ -189,3 +189,14 @@ i.fa {
...
+
+//popover css used in signup page for client side validation
+
+#signup_page .popover{
+    background-color: rgba(0,0,0,0.6);
+    color:white;
+}
+
+#signup_page .popover.right > .arrow:after{
+    border-right-color: rgba(0,0,0,0.6);
```
vitoravelino: Same here.

app/views/devise/registrations/new.html.slim
```diff
@@ -1,23 +1,90 @@
...
-        button.btn.btn-primary.btn-block.btn-lg type="submit"
+        button.btn.btn-primary.btn-block.btn-lg#submit_btn type="submit"
           i.fa.fa-check
           - if @admin || !@first_user_admin
             | Create account
           - else
             | Create admin
         - if @have_users
           .text-center = link_to 'I already have an account. Login.', new_user_session_path, class: 'btn btn-link'
+javascript:
```
vitoravelino: Move this out of the slim file. 

I suggest you to follow the same structure done on #1233. Since I was responsible for that, I can help you if you need any help.

app/views/devise/registrations/new.html.slim
```diff
@@ -1,23 +1,90 @@
...
     .col-md-4.col-sm-2.col-xs-1
     .col-md-4.col-sm-8.col-xs-10.text-center
       = render 'shared/notifications'
       = image_tag 'layout/portus-logo-login-page.png', class: 'login-picture'
       = form_for(resource, as: resource_name, url: { action: 'create' }, :html => {:autocomplete => "off"}) do |f|
-        = f.text_field :username, class: 'form-control input-lg', placeholder: 'Username', required: true, autofocus: true
-        = f.email_field :email, class: 'form-control input-lg first', placeholder: 'Email', required: true
-        = f.password_field :password, class: 'form-control input-lg', placeholder: 'Password (8 characters min.)', required: true
-        = f.password_field :password_confirmation, class: 'form-control input-lg last', placeholder: 'Password again', required: true
+        = f.text_field :username, class: 'form-control input-lg', placeholder: 'Username', required: true, autofocus: true,rel: 'popover'
```
vitoravelino: Add a space after comma.

app/views/devise/registrations/new.html.slim
```diff
@@ -1,23 +1,90 @@
...
+        = f.email_field :email, class: 'form-control input-lg first', placeholder: 'Email', required: true, rel: 'popover'
+        = f.password_field :password, class: 'form-control input-lg', placeholder: 'Password (8 characters min.)', required: true,rel: 'popover'
```
vitoravelino: Same here.

app/views/devise/registrations/new.html.slim
```diff
@@ -1,23 +1,90 @@
...
+        = f.password_field :password_confirmation, class: 'form-control input-lg last', placeholder: 'Password again', required: true,rel: 'popover'
```
vitoravelino: Same here.

vitoravelino:
Popover width too thin.

Submit button disabled if form is empty since is invalid.

app/assets/stylesheets/components/layout.scss
krishnak9: like this `rgba(0, 0, 0, 0.6);`

krishnak9: I will use the 2 spaces for Indent level :)

app/assets/stylesheets/components/layout.scss
krishnak9: Changing Now..:D

app/views/devise/registrations/new.html.slim
krishnak9: So the submit button must be disabled at the starting. I will change the popover width too.

app/views/devise/registrations/new.html.slim
krishnak9: Should I move the javascript to the Portus/app/assets/javascripts/modules/users/pages/sign-up.js ??

app/assets/stylesheets/components/layout.scss
vitoravelino: Yes. Also, move this to a separated file. It could be a `registration.scss` in the `pages` folder since you are overriding the default popover style on the sign up page.

One question I have: is it okay to override the popover style only on this page?  /cc @mssola

app/views/devise/registrations/new.html.slim
vitoravelino: Not only move to that but also restructure the code itself to follow a more component based style.

vitoravelino:
Tests are very welcome too. You can check the some of the rspec feature tests. 😄

krishnak9:
I am implementing the changes.... will push them soon...! I will also try to make travis pass.

vitoravelino:
Travis is already passing. That's a false negative, don't worry. It happens sometimes. 😅

krishnak9:
I did something similar to password-form.js by making signup-form.js and then using it from signup.js in pages . But when i check for event listener on the user_password and user_password_confirmation and email, Its not showing any. Please help me in this. I am new to this component based style.

vitoravelino:
@krishnak9, could you push the code?

app/assets/stylesheets/components/layout.scss
mssola: @vitoravelino I don't think it's ok to override it. All popovers should look and feel in the same way throughout the app.

app/views/devise/registrations/new.html.slim
mssola: I agree with @vitoravelino. Please change this to a more component based style.

app/assets/stylesheets/components/layout.scss
vitoravelino: That's what I thought, these styling can be removed then. 

I didn't see any form validation using popover. It's using the built-in validators. I think you can use [setCustomValidity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/setCustomValidity) method to achieve the same look and feel.

What do you think, @mssola?

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,36 @@
...
+
+const PASSWORD_FIELD = '#user_password';
+const CONFIRMATION_PASSWORD_FIELD = '#user_password_confirmation';
+const SUBMIT_BUTTON = '#submit_btn';
+
+// UsersPasswordForm component that handles user password form
+// interactions.
+class UsersSignUpForm extends BaseComponent {
+  elements() {
+    this.$Password = this.$el.find(PASSWORD_FIELD);
```
mssola: Could you rename variable to lower case ? (e.g. `this.$Password` should be `this.$password`).

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,36 @@
...
+
+  events() {
+    this.$el.on('keyup', PASSWORD_FIELD, e => this.onKeyup(e));
+    this.$el.on('keyup', CONFIRMATION_PASSWORD_FIELD, e => this.onKeyup(e));
+  }
+
+  onKeyup() {
+    const Password = this.$Password.val();
+    const PasswordConfirmation = this.$PasswordConfirmation.val();
+    const PasswordInvalid = !Password;
```
mssola: I don't think this is needed.

app/assets/javascripts/modules/users/pages/sign-up.js
```diff
@@ -1,12 +1,18 @@
...
 
 // UsersSignUpPage component responsible to instantiate
 // the user's sign up page components and handle interactions.
 class UsersSignUpPage extends BaseComponent {
+  elements() {
+    this.$signupForm = this.$el.find(SIGN_UP_FORM);
+  }
+
   mount() {
-    fadeIn(this.$el);
```
mssola: Please don't remove this 😉

app/assets/stylesheets/pages/registration.scss
```diff
@@ -0,0 +1,12 @@
+//popover css used in signup page for client side validation
```
mssola: Do you really have to re-style popovers ?

mssola:
Now you can remove all that jQuery code right ?

app/assets/javascripts/modules/users/components/signup-form.js
vitoravelino: ☝️

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,36 @@
...
+// interactions.
+class UsersSignUpForm extends BaseComponent {
+  elements() {
+    this.$Password = this.$el.find(PASSWORD_FIELD);
+    this.$PasswordConfirmation = this.$el.find(CONFIRMATION_PASSWORD_FIELD);
+    this.$submit = this.$el.find(SUBMIT_BUTTON);
+  }
+
+  events() {
+    this.$el.on('keyup', PASSWORD_FIELD, e => this.onKeyup(e));
```
vitoravelino: I think the reason nothing is happening on the focus out is because is you listening to the wrong event. You should use `blur`.

vitoravelino: Also rename the method event listener.

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,36 @@
...
+    this.$el.on('keyup', CONFIRMATION_PASSWORD_FIELD, e => this.onKeyup(e));
+  }
+
+  onKeyup() {
+    const Password = this.$Password.val();
```
vitoravelino: Lowercase for variables.

krishnak9:
Yeah I will remove all that jquery codes on the slim file, but i still ididnt get this component based style structure to work. I am still doing it. I am new to this so might take some time to learn all.

app/assets/javascripts/modules/users/pages/sign-up.js
vitoravelino: ☝️

app/assets/javascripts/modules/users/components/signup-form.js
krishnak9: Yes . I will do that for sure .

app/assets/stylesheets/pages/registration.scss
vitoravelino: Github hided my comment regarding this.

"That's what I thought. So, these styling can be removed then.

I didn't see any form validation using popover. It's using the built-in validators. I think you can use setCustomValidity method to achieve the same look and feel.

What do you think, @mssola?"

krishnak9: I mean its not much necessary but if we use popover for validation messages then the width of the popover must be increased. But i think i will remove this and use [HTMLSelectElement.setCustomValidity()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/setCustomValidity) as suggested by @vitoravelino

app/assets/javascripts/modules/users/components/signup-form.js
krishnak9: Even on the 'keyup' its not listening.

app/assets/javascripts/modules/users/components/signup-form.js
krishnak9: I am on it :) .!

vitoravelino:
@krishnak9 Let me try to explain. There's a [BaseComponent](https://github.com/SUSE/Portus/blob/master/app/assets/javascripts/base/component.js) class that is a simple empty js class that tries to avoid some boilerplate of having to call common methods that repeats in every component. There's an order and purpose of those methods.

When you extend `BaseComponent`, the component inherits its constructor that expects a jQuery wrapped dom element. This element works like the scope of the component. Everything you need to access on DOM should be in its dom tree. We know that we could just use `$(selector)` and access anything but it's not recommended due lack of control and scope of the component. 

Components should do as simple things as possible, if you have many responsibilities in a component, you might need to split into separated components or delegate to a service.

Regarding migrating your initial code to this structure, you can see that you repeat yourself a lot wrapping the same dom elements, right? I guess you've already figured that out and started putting all the elements that you are gonna need to access inside the `elements()` method.

Second, and really important, thing: events. When you need to listen to any event from any dom element (ideally below this.$el, there will be cases that you are gonan need to listen from another component or service), this is where you are gonna define the event listeners (e.g keyup, blur, focus, etc). To listen to dom elements events, we use the (.on())[http://api.jquery.com/on/] method. I'm not sure if you are aware that 

```js
(e) => this.onKeyup(e)
````

is the same thing as 

```js
function (e) {
  this.onKeyUp(e);
}.bind(this)
```

That's just some syntax suger from ES2015.

This covers the basics of the current component based structure. With that knowledge I think you can do most of the migration. Let me know if you have any question.

krishnak9:
Finally found what was going wrong. I didn't reload my vagrant environment and thats why the changes didn't take place 😅. I also removed the popover and implemented `setCustomValidity()` method. Please review @vitoravelino @mssola .

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,52 @@
...
+    this.$el.on('focusout', EMAIL_FIELD, e => this.onfocusout(e));
+  }
+
+  onfocusout() {
+    const usernameInvalid = !this.$username;
+    const emailInvalid = !this.$userEmail[0].validity.valid;
+    const passwordInvalid = !this.$password.val() || this.$password.val().length < 8;
+    const passwordConfirmationInvalid = !this.$passwordConfirmation.val() ||
+      this.$password.val() !== this.$passwordConfirmation.val();
+    if (passwordConfirmationInvalid) {
```
vitoravelino: Add an empty line above.

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,52 @@
...
+      this.$passwordConfirmation[0].setCustomValidity('Please enter same Password as Above');
+    } else {
+      this.$passwordConfirmation[0].setCustomValidity('');
+    }
+    if (passwordInvalid) {
```
vitoravelino: Same here.

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,52 @@
...
+      this.$password[0].setCustomValidity('Password should be of at least 8 characters minimum.');
+    } else {
+      this.$password[0].setCustomValidity('');
+      this.$passwordConfirmation.attr('pattern', this.$password.val());
+    }
+    if (passwordInvalid || passwordConfirmationInvalid || emailInvalid || usernameInvalid) {
```
vitoravelino: Same here.

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,52 @@
...
+const SUBMIT_BUTTON = '#submit_btn';
+
+// UsersPasswordForm component that handles user password form
+// interactions.
+class UsersSignUpForm extends BaseComponent {
+  elements() {
+    this.$password = this.$el.find(PASSWORD_FIELD);
+    this.$passwordConfirmation = this.$el.find(CONFIRMATION_PASSWORD_FIELD);
+    this.$username = this.$el.find(USERNAME_FIELD);
+    this.$userEmail = this.$el.find(EMAIL_FIELD);
```
vitoravelino: If you are not gonna use the $element, you don't need to cache it here. So, I think you can remove both `username` and `email` elements, constants and event listeners.

app/assets/javascripts/modules/users/components/signup-form.js
krishnak9: before the `if` statement ?

app/assets/javascripts/modules/users/components/signup-form.js
vitoravelino: Oops. My bad! Forgot the disabled button. 😅

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,52 @@
...
+
+  events() {
+    this.$el.on('focusout', PASSWORD_FIELD, e => this.onfocusout(e));
+    this.$el.on('focusout', CONFIRMATION_PASSWORD_FIELD, e => this.onfocusout(e));
+    this.$el.on('focusout', USERNAME_FIELD, e => this.onfocusout(e));
+    this.$el.on('focusout', EMAIL_FIELD, e => this.onfocusout(e));
+  }
+
+  onfocusout() {
+    const usernameInvalid = !this.$username;
```
vitoravelino: I think you should do the same as below, no?

app/assets/javascripts/modules/users/components/signup-form.js
krishnak9: Yeah ..they are used in `emailInvalid` , `usernameInvalid` and for disabling the submit button. So no need to remove them right ?

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  events() {
+    this.$el.on('focusout', PASSWORD_FIELD, e => this.onfocusout(e));
+    this.$el.on('focusout', CONFIRMATION_PASSWORD_FIELD, e => this.onfocusout(e));
+    this.$el.on('focusout', USERNAME_FIELD, e => this.onfocusout(e));
+    this.$el.on('focusout', EMAIL_FIELD, e => this.onfocusout(e));
+  }
+
+  onfocusout() {
```
vitoravelino: I would suggest to rename it to `onFocusout`. The idea (at least that's what I do) is to start a function/method with `on` only if it's an event listener/callback. That will highlight that this is probably a a function that is used as an event listener/callback related to the `focusout` event. 

I know that in this case it wouldn't add much value because this component is very simple, but it's good to have consistency around.

app/assets/javascripts/modules/users/components/signup-form.js
krishnak9: Yeah ! I will do that . In that way if we use any pattern based user-name ,it will also get validated.

app/assets/javascripts/modules/users/components/signup-form.js
vitoravelino: No need. I just woke up. Wasn't thinking really straight yet. KKKK.

app/assets/javascripts/modules/users/components/signup-form.js
vitoravelino: Yes. Make code easier to read.

app/assets/javascripts/modules/users/components/signup-form.js
krishnak9: Yeah..! Sure 😀.!

krishnak9:
Did the changes! , Don't know why travis is failing here. In my fork it did passed [Travis (commit: changes)](https://travis-ci.org/krishnak9/Portus/builds/212076884)

vitoravelino:
It's passing, don't worry. As I said before, it's a false negative.

krishnak9:
Oh Cool 😀 . Any more changes !

app/assets/javascripts/modules/users/components/signup-form.js
```diff
@@ -0,0 +1,55 @@
...
+
+  onFocusout() {
+    const usernameInvalid = !this.$username[0].validity.valid;
+    const emailInvalid = !this.$userEmail[0].validity.valid;
+    const passwordInvalid = !this.$password.val() || this.$password.val().length < 8;
+    const passwordConfirmationInvalid = !this.$passwordConfirmation.val() ||
+      this.$password.val() !== this.$passwordConfirmation.val();
+
+    if (passwordConfirmationInvalid) {
+      this.$passwordConfirmation[0].setCustomValidity('Please enter same Password as Above');
```
vitoravelino: To lowercase `Password` and `Above`.

vitoravelino:
I'm sure that your code is working as expected but would you mind to write some feature tests to ensure that?

krishnak9:
Yeah Okay. I will try to do that.

app/assets/javascripts/modules/users/pages/sign-up.js
```diff
@@ -1,12 +1,19 @@
...
+import SignUpForm from '../components/signup-form';
+
+const SIGN_UP_FORM = 'form.new_user';
 
 // UsersSignUpPage component responsible to instantiate
 // the user's sign up page components and handle interactions.
 class UsersSignUpPage extends BaseComponent {
+  elements() {
+    this.$signupForm = this.$el.find(SIGN_UP_FORM);
+  }
```
mssola: nitpicking: add an empty line to separate the `elements` and the `mount` functions 😄

mssola:
And yeah, please add a feature test 👍

app/assets/javascripts/modules/users/pages/sign-up.js
krishnak9: Yeah Sure 😃

krishnak9:
To write feature tests for this, I have to execute the javascript using capybara-webkit driver, therefore I need to include this gem too. As I have to trigger the `focusout` event and check if the button is disabled or not if fields are wrong or Is there any other way to do this ? @vitoravelino @mssola

mssola:
> To write feature tests for this, I have to execute the javascript using capybara-webkit driver, therefore I need to include this gem too. As I have to trigger the focusout event and check if the button is disabled or not if fields are wrong or Is there any other way to do this ? @vitoravelino @mssola

There's no need to include any new gem. Notice that whenever you want a feature test to be able to execute JS, you need to add the `js: true` flag. That being said, maybe you can take a look at the `spec/support` directory, which include some helper methods. In that directory, I think that the `wait_for_events.rb` and the `helpers.rb` files can be of help (and don't be afraid to extend them if you think it's necessary 😉).

krishnak9:
Sorry I got busy with the hectic college schedule. For feature test I have done something like this

 `scenario "Submit Button gets disabled when any field is filled wrong", js: true do`
 `   fill_in "user_username", with: user.username`
 `   fill_in "user_email", with: "gibberish"`
 `   fill_in "user_password", with: "12341234"`
 `   fill_in "user_password_confirmation", with: "532"`
  `  page.execute_script "$('#user_username').trigger('focusout')"`
  `  expect(page).to have_button('submit_btn', disabled: true)`
    
  `end`
but there was an error saying 

`Capybara::ElementNotFound:`
 `Unable to find field "user_username"`

I can't find out what caused this error. @vitoravelino  @mssola

mssola:
@krishnak9 iirc, that only worked (as in the other tests of this same file) when `js: true` was not specified. When JS is needed, you have to take another approach (I still don't fully know why 😅). Check other tests with `js: true` on how this can be accomplished. If I recall correctly, it could work with something like this:

```ruby
fill_in "Username", with: user.username
```

or with something like this (I don't recall if this was the exact way of doing it though):

```ruby
find("#user_username").fill user.username
```

You can take a look at the `spec/features/namespaces_spec.rb` file to check some tests using `js: true`. Moreover, I'd recommend adding a wait method before this: `expect(page).to have_button('submit_btn', disabled: true)`. Otherwise this might become a flaky test I fear 😃

krishnak9:
Oh okay 😃 !. I will try to find out the approach for this . Thanks..! I will check by using this `find("#user_username").fill user.username` . I think this may work.

krishnak9:
I tried so many ways but error is still there. Can it be because Phantom JS does not support ES6 style. [https://github.com/ariya/phantomjs/issues/14506](url). @mssola @vitoravelino

krishnak9:
Finally found a way to work through 😅. Added the feature test for Client side validation. @mssola @vitoravelino Please review. 😃

spec/features/auth/signup_feature_spec.rb
```diff
@@ -143,6 +143,32 @@
...
+    visit "/users/sign_up"
+    wait_for_effect_on("#new_user")
+    fill_in "Username", with: user.username
+    fill_in "user_email", with: "gibberish"
+    fill_in "user_password", with: "12341234"
+    fill_in "user_password_confirmation", with: "532"
+    page.execute_script "$('#user_username').trigger('focusout')"
+    wait_for_effect_on("#user_username")
+    expect(page).to have_button("submit_btn", disabled: true)
+
```
vitoravelino: Remove empty line.

spec/features/auth/signup_feature_spec.rb
```diff
@@ -143,6 +143,32 @@
...
+    visit "/users/sign_up"
+    wait_for_effect_on("#new_user")
+    fill_in "Username", with: user.username
+    fill_in "user_email", with: user.email
+    fill_in "user_password", with: user.password
+    fill_in "user_password_confirmation", with: user.password
+    page.execute_script "$('#user_username').trigger('focusout')"
+    wait_for_effect_on("#user_username")
+    expect(page).to have_button("submit_btn", disabled: false)
+
```
vitoravelino: Same here.

spec/features/auth/signup_feature_spec.rb
```diff
@@ -143,6 +143,32 @@
...
 
+  scenario "Submit Button gets disabled when any field is filled wrong", js: true do
+    visit "/users/sign_up"
+    wait_for_effect_on("#new_user")
+    fill_in "Username", with: user.username
+    fill_in "user_email", with: "gibberish"
+    fill_in "user_password", with: "12341234"
+    fill_in "user_password_confirmation", with: "532"
+    page.execute_script "$('#user_username').trigger('focusout')"
+    wait_for_effect_on("#user_username")
```
vitoravelino: Did you test without this? Just curious.

krishnak9: No I didn't test without this. I just want to make sure that `focusout` get's completed before the check for disabled submit button.

krishnak9: I did it as @mssola mentioned about test might become flaky
> You can take a look at the spec/features/namespaces_spec.rb file to check some tests using js: true. Moreover, I'd recommend adding a wait method before this: expect(page).to have_button('submit_btn', disabled: true). Otherwise this might become a flaky test I fear :smiley:

krishnak9:
What more should be done here . Or is this concluded. Please review @mssola @vitoravelino .

spec/features/auth/signup_feature_spec.rb
mssola: Maybe I'd wait for the effect on the submit button, since this is the element that will change its state.

mssola:
@krishnak9 fix my nitpick, and then squash your commits into a single one. Then I'd say it will be good to be merged. Thanks 👏

krishnak9:
I messed up here sending a new PR with all these changes .

mssola:
Just one final thing, could you sign off your commit as specified in the [CONTRIBUTING.md file](https://github.com/SUSE/Portus/blob/master/CONTRIBUTING.md#sign-your-work) ? If you have your author & email in your git config, it should be a matter of just using the `-s` flag in the `commit` command.

Thanks!

krishnak9:
Oh okay. I will disable the button at the starting. I see the problem here. Is it okay that if I add a new `keyup` method which is called only on the confirmation password field and keeping others with focusout only . With this as soon the password matches, the submit button gets enabled. And having a variable set true if all the credentials before the password matching are correct .  By checking this variable and the password matching on each keyup inside the keyup method. I think this will solve the problem and also it will call the keyup method only "length of the password number of times". What are your views on this @mssola

mssola:
> I will disable the button at the starting

You can even write a very simple feature test that checks that when visiting the signup page this element is disabled.

> Is it okay that if I add a new keyup method which is called only on the confirmation password field and keeping others with focusout only . With this as soon the password matches, the submit button gets enabled. And having a variable set true if all the credentials before the password matching are correct . By checking this variable and the password matching on each keyup inside the keyup method. I think this will solve the problem and also it will call the keyup method only "length of the password number of times". What are your views on this 

It has to be applied always. I just said password confirmation as an example. Another use case:

1. I type the email wrong but I continue anyways.
2. Type the password and password confirmation correctly.
3. Go back to fix my email

In this case, the button won't be enabled when I've just corrected. So, to be fair, this should be done for each input element.

vitoravelino:
In order to make it more readable, I guess you could have individuals event listeners for each field and store the validation result in a variable. This way you are more flexible to manage the experience of giving a specific feedback back to the user based on the field. In the end of each of these listeners you would need to a common method that will check if all the validations are valid and if yes, enable the button, otherwise disable it.

Since the project is not using a lib/fw with some feature/sugar syntax and `BaseComponent` is still super simple, this will be a bit verbose and feel not elegant but, remember, it's plain js.

krishnak9:
Yeah this is true. This should be done on all the input elements .I will do the changes. I will try to do with individual event listeners or the timer based approach, I will check for it. Will push the changes soon.

krishnak9:
I have implemented this using `keyup`, having different event handlers for each field. But when writing test, I have to enable `js:true` in each test where submit button is clicked as the submit button is disabled at start . There is this error showing on tests . I have attached the screenshot 
screenshot from 2017-04-11 14 38 51 @mssola  @vitoravelino . What caused this ??

vitoravelino:
@krishnak9 sorry the delay, any luck so far?

mssola:
@krishnak9 are you still working on this ?

krishnak9:
@mssola @vitoravelino Sorry Sir ! I got busy with my college campus placements so I m not getting time to actually work on this. If required you can close this PR. When I will resume the work, I will send a new PR.

mssola:
Let's close this then. It can be re-opened if none else has worked on this in the meantime. Thanks for working on this 👍

PR #1252, opened 2017-03-17
Quick fix and improvement on team#update

vitoravelino:
Also added alert notification and removed unused variables.

### DIFF
--- a/app/assets/javascripts/teams.js
+++ b/app/assets/javascripts/teams.js
@@ -1,5 +1,5 @@
 jQuery(function ($) {
-  $('#add_team_user_btn').on('click', function (_event) {
+  $('#add_team_user_btn').on('click', function () {
     var team_id;
 
     $('#team_user_user').val('');
@@ -19,8 +19,9 @@ jQuery(function ($) {
     set_typeahead(team_id + '/typeahead/%QUERY');
   });
 
-  $('body').on('click', '.btn-edit-role', function (_event) {
+  $('body').on('click', '.btn-edit-role', function (event) {
     var el = $(this).find('i.fa');
+
     if ($(this).hasClass('add')) {
       open_close_icon(el);
       $('#team_user_' + event.currentTarget.value + ' td .role').toggle();
@@ -32,7 +33,7 @@ jQuery(function ($) {
     }
   });
 
-  $('#add_webhook_btn').unbind('click').on('click', function (_event) {
+  $('#add_webhook_btn').unbind('click').on('click', function () {
     $('#webhook_url').val('');
     $('#webhook_username').val('');
     $('#webhook_password').val('');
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/teams.js
```diff
@@ -50,7 +51,7 @@ jQuery(function ($) {
     });
   });
 
-  $('body').on('click', '.btn-edit-webhook', function (_event) {
+  $('body').on('click', '.btn-edit-webhook', function (event) {
     if ($(this).hasClass('button_edit_webhook')) {
       $('.webhook_information').toggle();
       $('#update_webhook_' + event.currentTarget.value).toggle();
```
vitoravelino: `event` was being used instead of `_event`.

app/assets/javascripts/teams.js
```diff
@@ -1,5 +1,5 @@
 jQuery(function ($) {
-  $('#add_team_user_btn').on('click', function (_event) {
+  $('#add_team_user_btn').on('click', function () {
```
mssola: Is this parameter really optional then ? I wrote `_event` to state "jQuery wants to pass a parameter but we don't use it". If so, I think that we can strip most of the parameters that start with "_" 😄

app/assets/javascripts/teams.js
mssola: Yeah, good catch 👍  Note that we write variables starting with _ when we have to have a function parameter but we don't use it (see my comment above). And it looks like I didn't see this usage.

mssola:
Thanks a bunch 👏

app/assets/javascripts/teams.js
vitoravelino: Yes, it's optional. By default there's no check on how functions are referenced/called. JS sucks on those kind of stuff and that's one of the reasons why TypeScript is getting more and more popular and being used in bigger projects.

Just saw that you've configured ESLint to ignore unused variables starting with `_`. I guess you can remove that and let ESLint do the job of checking. 

I know that eventually there will be a case like:

```js
(not_used, used) => {
  // ...
}
```

and you can use `_not_used` without needing to configure ESLint to not complain about it. ESLint is smart enough to know that `used` is used and the first parameter is necessary to be declared even if not used. So tou don't need to configure to ignore variables starting with `_`.

On the other hand, keeping the `_var` indicates that the function receives that parameter but it's not being used in it. If an app gets too big and lacks documentation, having that artifact would help a lot. In this case, it's a well documented and know scenario of listening to a dom event through jQuery. I don't think it's necessary, but it's your call to keep it or not.

I also realized that `event` is a reserved keyword and that's another reason why ESLint didn't catch the erro of `event` not being defined. For events I'm comfortable with `e` or `evt`.

What do you think?

mssola: Totally agree with you. Thanks for the explanation ! 👍

PR #1256, opened 2017-03-20
Add local webpack for building

Diokuz:
Just added local webpack: no global webpack need anymore, just
```
yarn
yarn run webpack
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,8 @@
 {
   "private": true,
   "scripts": {
-    "eslint": "eslint --max-warnings 0 --ext .js ."
+    "eslint": "eslint --max-warnings 0 --ext .js .",
+    "webpack": "webpack --watch --config config/webpack.js"
   },
   "devDependencies": {
     "babel-core": "^6.23.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,25 +1,26 @@
 {
   "private": true,
   "scripts": {
-    "eslint": "eslint --max-warnings 0 --ext .js ."
+    "eslint": "eslint --max-warnings 0 --ext .js .",
+    "webpack": "webpack --watch --config config/webpack.js"
```
mssola: This should be the only change. I don't understand the rest of the changes 😕

vitoravelino: Alphabetical order.

mssola: I was complaining because it's not really needed, and I'm sure that this alphabetical order won't be preserved over time 😀 Anyways, I don't want to be a blocker. Merging...

vitoravelino: It's automatic, it wasn't him but yarn.

xiaoxueshengyueyue:
hi,I also meet this problem,when i used yarn run webpack, it  appears another problem,ERROR in Entry module not found: Error: Can't resolve 'babel-loader' in '/root/Portus/app/assets/javascripts'?
can you help me fix this ?
and i also didn't solve the problem about Webpack::Rails::Manifest::EntryPointMissingError in Auth::Registrations#new

thank you very much

mssola:
@xiaoxueshengyueyue  Are you sure that you have all the dependencies where you have Portus installed ? As in, have you run `yarn install` there ? Moreover, how are you deploying Portus ?

xiaoxueshengyueyue:
@mssola Thank you very much,my bad,i skip to use yarn install,so it didn't work,but after when i  do the following  steps ,it works.
step1.install yarn
(1)wget https://dl.yarnpkg.com/rpm/yarn.repo -O /etc/yum.repos.d/yarn.repo,
(2)curl --silent --location https://rpm.nodesource.com/setup_6.x | bash -
(3) yum install yarn
step2.yarn install
step3.yarn run webpack
After three steps,It solved.

PR #1266, opened 2017-03-27
Fix dashboard sidebar links

### LINKED ISSUE #1265: Sidebar links don't work.
The links in the sidebar of the dashboard are currently broken due to click intercepts which get bound to them from [dashboard.js](https://github.com/SUSE/Portus/blob/d5d33ecce2332e3313e2b5c7cf55409c13e8d4c4/app/assets/javascripts/dashboard.js), e.g.:
```
  $('#starred a').on('click', function (e) {
    e.preventDefault();
    $(this).tab('show');
  });
```
These intercepts are intended for the links within the tabs themselves, rather than the tab contents (to toggle display of the 'all', 'starred', and 'personal' tabs). Because the IDs which are selected by these intercepts are being given to the tab contents rather than the element containing the links in the tab header, the links to namespaces and repositories within the tab contents can't be followed.

portus-dashboard-sidebar-bug

HartS:
Added an ID to the dashboard sidebar tabs header and modified dashboard.js to select all of tab links based on their relationship to this ID, as in the example at http://getbootstrap.com/javascript/#tabs-examples.

This closes #1265 

Note: Since we are not cache-busting our webpack assets in development, this fix may only be evident if one's browser has been configured to not load assets from its cache, or if the webpack bundle is force reloaded via a hard reload of $PORTUS_ROOT_URL/assets/webpack/application.js

Signed-off-by: Hart Simha <hart.simha@suse.com>

### DIFF
--- a/app/assets/javascripts/dashboard.js
+++ b/app/assets/javascripts/dashboard.js
@@ -1,15 +1,5 @@
 jQuery(function ($) {
-  $('#starred a').on('click', function (e) {
-    e.preventDefault();
-    $(this).tab('show');
-  });
-
-  $('#all a').on('click', function (e) {
-    e.preventDefault();
-    $(this).tab('show');
-  });
-
-  $('#personal a').on('click', function (e) {
+  $('#sidebar-tabs a').on('click', function (e) {
     e.preventDefault();
     $(this).tab('show');
   });
--- a/app/views/dashboard/index.html.slim
+++ b/app/views/dashboard/index.html.slim
@@ -12,7 +12,7 @@
       .panel-heading
         h5 Repositories
 
-        ul.nav.nav-tabs{role="tablist"}
+        ul#sidebar-tabs.nav.nav-tabs{role="tablist"}
           li.active{role="presentation"}
             a{href="#all" aria-controls="all" role="tab" data-toggle="tab"} All
           li{role="presentation"}


dgdavid:
If I'm not mistaken and according to the markup, you also could use something more generic like

```javascript

$('[role="tablist"] a').on('click', function (e) {
  ...
})
```
or even

```javascript

$('a[role="tab"]').on('click', function (e) {
  ...
})
```

This way help you to avoid using a specific ID for a _pseudo-common_ behaviour. Furthermore, there are a lot of literature encouraging you to avoid the use of ID selectors; right now, I remember [this](https://dev.to/buntine/dom-elements-with-ids-are-global-variables) that I read recently.

vitoravelino:
I personally prefer classes over ids but for this case it's fine. Never or almost never use elements/pseudo on selectors because they are harder to read/understand and slower. Same applies when writing css.

LGTM.

mssola:
@vitoravelino correct me if I'm wrong, but this is fixed in #1289, right ?

vitoravelino:
> @vitoravelino correct me if I'm wrong, but this is fixed in #1289, right ?

Yeah, [here](https://github.com/SUSE/Portus/pull/1289/files#diff-0dbec4de505397dcbaade37c663f02b1).

mssola:
Since we have just merged it, we can close this PR. Thanks for your time.

PR #1267, opened 2017-03-28
fix vagrant setup

saurabhsurana:
- npm install of yarn, webpack
- yarn install inside /vagrant directory
- running webpack locally on the portus web vm


mssola:
Thanks 👏

PR #1275, opened 2017-04-06
imported typeahead from npm and upgraded it

vitoravelino:
Initially I kept the same version but I took the liberty to upgrade it to the newer one because it exports the object properly instead of just exposing it as a global variable. 

The behavior still the same but if the project prefers to keep the previous version, just let me know and I'll downgrade it.


vitoravelino:
Wrong commit.

PR #1276, opened 2017-04-06
imported typeahead from npm and upgraded it

vitoravelino:
Initially I kept the same version but I took the liberty to upgrade it to the newer one because it exports the object properly instead of just exposing it as a global variable.

The behavior still the same but if the project prefers to keep the previous version, just let me know and I'll downgrade it.


app/assets/stylesheets/components/typeahead.scss
```diff
@@ -1,7 +1,46 @@
+.tt-menu {
```
mssola: This file should be kept in vendor I think, since we took it from the gem iirc.

vitoravelino: You are right. I made 2 or 3 changes due upgrade and previous enhancement but I'll keep it in vendor folder. Thanks for reminding me that. :)

vitoravelino:
Done!

mssola:
Thanks 👏

PR #1289, opened 2017-05-19
Hackweek security scanning

vitoravelino:
Opening this PR to start getting some feedback. As @mssola mentioned to me, this PR became quite large. Thinking about that I'm gonna work on the data tables filtering in another branch.

There are still some missing tasks that I'll be updating as soon as I finish them.

- [x] Convert .js with templates to .vue 
- [x] Move repository to the top of the page
- [x] Move pull/push/role info somewhere else
- [x] Restyle fav icon
- [x] Restyle delete repository
- [x] Delete tag appears once a tag is selected
- [x] Tests

Extracted from @mssola comments:

There are some issues when the security scanning is not available:

- [x] No tags are returned, when it should return at least the tags with no vulnerability information.
- [x] A flashy message "Unable to fetch newer tags data" is shown for every polling. This is distracting, and we should show it once instead.
- [x] The `tags#show` panel should have a "No vulnerabilities found" instead of having an empty space.

Bits missing:

- [x] Ping @jordimassaguerpla and/or @MaximilianMeister, so they can review the updated gems (see [comment from @jordimassaguerpla](https://github.com/SUSE/Portus/pull/1319#issuecomment-315770718)) (fixed in #1321 and #1322)
- [x] Add documentation (see #1320)
- [ ] Add documentation to coreos/clair about this integration (see [this page](https://github.com/coreos/clair/blob/master/Documentation/integrations.md)).
- [x] Rebase into proper commits.


vitoravelino:
@mssola I've managed to solve the issues with the `undefined` errors by just making `poltergeist` driver use the phantomjs executable handled by the `phantomjs` gem that was installed by I guess not being used so far.

Another `portusctl` test failed after you've fixed the flags security. This new one is in `man_spec.rb`.

Looking forward to your review. Thanks! 😃

mssola:
Reviewing the functionality itself, not the code:

- I'd like to preserve the colors of the star widget. That blue color fit with the general design of Portus better.
- If I only have one tag, I can read: "Showing from 1 to 1 of 1 entries", which is quite redundant. I'd say that if there's only one page, then stating "Showing 1 entry" should be enough, or even remove this message altogether in this case.
- If there's no security backends enabled, then this feature should be entirely hidden.

vitoravelino:
> If there's no security backends enabled, then this feature should be entirely hidden.

By enabled you mean any of the 3 security options being != "", correct? Is there already a helper for that or should I write one?

mssola:
> By enabled you mean any of the 3 security options being != "", correct? Is there already a helper for that or should I write one?

Yeah exactly. There's no helper for that, but I can write it. Let me do it.

vitoravelino:
@mssola requested changes were done. 

Also, I've written more tests to hit the expected code coverage but there still 2 classes (security/zypper|clair.rb) and 1 piece of code (application_controller.rb:39) that I couldn't write any test because I had no knowledge about it. 😕

jzelinskie:
Howdy,

I'm a maintainer over at Clair. It'd be awesome if you guys added yourself to our list of documented integrations: https://github.com/coreos/clair/blob/master/Documentation/integrations.md

mssola:
> I'm a maintainer over at Clair. It'd be awesome if you guys added yourself to our list of documented integrations: https://github.com/coreos/clair/blob/master/Documentation/integrations.md

Sure thing. Once we get this merged I'll submit a PR. Thanks!

vitoravelino:
Passing! Woohoo!

app/assets/javascripts/shared/components/alert.js
```diff
@@ -0,0 +1,20 @@
+const ALERT_ELEMENT = '#float-alert';
+const TEXT_ALERT_ELEMENT = '#float-alert p';
+const HIDE_TIMEOUT = 5000;
+
+function scheduleHide() {
```
mssola: I don't think this is needed... As in, I'd prefer to inline this code where it's used.

app/assets/javascripts/utils/eventbus.js
```diff
@@ -0,0 +1,3 @@
+import Vue from 'vue';
+
+export default new Vue({});
```
mssola: What is this ?

vitoravelino: It returns a plain new Vue object with built-in methods like `$emit`, `$on`, `$off`. It's the same thing as doing `$({})` in jQuery like I tried to do [here](https://github.com/kubic-project/velum/pull/167/files#diff-3a404e107e5ad941e98771e088f6cbc1).

I did this to simulate a pub/sub object and let different components communicate without compromising everything with coupling. As an example, normally the child/inner components shouldn't handle data/services requests as a good practice. It should bubble an event and let the page/root parent component handle all of this. This way if we need to check how data is being handled at first, we will know where to take a look at. 

Not sure if I explained well enough for you. Let me know. :)

app/assets/javascripts/shared/components/alert.js
vitoravelino: What do you mean with "inline this code where it's used"? Can you give an example?

vitoravelino:
No worries about the delay, we had more important stuff going on. :)

app/assets/javascripts/shared/components/alert.js
mssola: You don't have to create a new function for something simple that it's only used once.

app/assets/javascripts/utils/eventbus.js
mssola: Thanks for the explanation 👍

app/assets/javascripts/shared/components/alert.js
vitoravelino: Ohhh... Gotcha! Will change that then. Thanks!

mssola:
@vitoravelino I've manually tested this and it looks good. Just fix my comment and we'll be good to proceed.

mssola:
Bits missing:

- [ ] Add documentation.
- [ ] Add documentation to coreos/clair about this integration (see [this page](https://github.com/coreos/clair/blob/master/Documentation/integrations.md)).
- [ ] Rebase into proper commits.

mssola:
There are some issues when the security scanning is not available:

- [ ] No tags are returned, when it should return at least the tags with no vulnerability information.
- [ ] A flashy message "Unable to fetch newer tags data" is shown for every polling. This is distracting, and we should show it once instead.

mssola:
We have to rebase on top of master because this conflicts with the merged #1319.

mssola:
Rebased

mssola:
@vitoravelino I've rebased everything into four commits. Hopefully I haven't broken anything ... Anyways, could you add your sign-off in all the commits ?

vitoravelino:
> @vitoravelino I've rebased everything into four commits. Hopefully I haven't broken anything ... Anyways, could you add your sign-off in all the commits ?

Done, @mssola.

PR #1319, opened 2017-07-13
Improved testing

vitoravelino:
- Moved from `precise` to `trusty` in travis
  - https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming
- Updated `.travis.yml`
- Updated `capybara` gem
- Updated `poltergeist` gem
- Removed `phantomjs`gem 
- Removed unnecessary `wait_until`
- Removed unnecessary `wait_for_effect_on`
- Using `have_current_path` matcher instead of `eq`
- Fixed poltergeist js error on sign in/up pages fadeIn
- Fixed two failing tests after capybara update
- Fixed `admin/activities_controller_spec.rb` flaky test
- Many minor enhancements
- Fixed webmock dependency tree 
  - It was blocking the tests after .travis.yml changes 
  - Also causing `ruby-head` to fail
- Updated `mysql2`gem
  -  It was causing conflict in travis for some reason after caching bundle

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ EXPOSE 3000
 
 WORKDIR /srv/Portus
 COPY Gemfile* ./
-RUN bundle install --retry=3 && bundle binstubs phantomjs
+RUN bundle install --retry=3
 RUN apt-get update && \
     apt-get install -y --no-install-recommends nodejs
 
--- a/app/assets/javascripts/modules/users/pages/sign-in.js
+++ b/app/assets/javascripts/modules/users/pages/sign-in.js
@@ -6,7 +6,7 @@ import { fadeIn } from '~/utils/effects';
 // the user's sign in page components and handle interactions.
 class UsersSignInPage extends BaseComponent {
   mount() {
-    fadeIn(this.$el);
+    fadeIn(this.$el.find('> .container-fluid'));
   }
 }
 
--- a/app/assets/javascripts/modules/users/pages/sign-up.js
+++ b/app/assets/javascripts/modules/users/pages/sign-up.js
@@ -6,7 +6,7 @@ import { fadeIn } from '~/utils/effects';
 // the user's sign up page components and handle interactions.
 class UsersSignUpPage extends BaseComponent {
   mount() {
-    fadeIn(this.$el);
+    fadeIn(this.$el.find('> .container-fluid'));
   }
 }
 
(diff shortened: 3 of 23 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ EXPOSE 3000
 
 WORKDIR /srv/Portus
 COPY Gemfile* ./
-RUN bundle install --retry=3 && bundle binstubs phantomjs
```
vitoravelino:
```
$ bundle binstubs phantomjs
There are no executables for the gem phantomjs.
```

app/assets/javascripts/modules/users/pages/sign-in.js
```diff
@@ -6,7 +6,7 @@ import { fadeIn } from '~/utils/effects';
 // the user's sign in page components and handle interactions.
 class UsersSignInPage extends BaseComponent {
   mount() {
-    fadeIn(this.$el);
+    fadeIn(this.$el.find('> .container-fluid'));
```
vitoravelino: Cannot be `body` (this.$el).

app/assets/javascripts/modules/users/pages/sign-up.js
```diff
@@ -6,7 +6,7 @@ import { fadeIn } from '~/utils/effects';
 // the user's sign up page components and handle interactions.
 class UsersSignUpPage extends BaseComponent {
   mount() {
-    fadeIn(this.$el);
+    fadeIn(this.$el.find('> .container-fluid'));
```
vitoravelino: Same here.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "eslint": "eslint --max-warnings 0 --ext .js .",
+    "eslint": "eslint --max-warnings 0 --ext .js ./app/assets/javascripts",
```
vitoravelino: It was considering `travis_phantomjs` folder.

vitoravelino:
The only weird thing is that after caching bundle, from time to time this happens:

```
Gem Load Error is: Incorrect MySQL client library version! This gem was compiled for 10.0.31-MariaDB but the client library is 5.5.54.
```

This never happened with Velum afaik. I was also having issues to install mysql2 gem (0.3.18 in .lock) locally but had no issues when I upgraded to ~> 0.4.x. I guess that might be related.

jordimassaguerpla:
Hi. You are updating a lot of gems which I think are not related to this change. ASFAIK, updating capybara and poltergeist do not require the updates on the rest of gems. Could you please review it?

PR #1324, opened 2017-07-18
Fix concise model annotation

Vad1mo:
This PR has no impact it just updates model annotation comment to the latest state of the annotation gem. Its quite annoying that the workspace is always dirty without actually doing anything in that particular file.


config/routes.rb
```diff
@@ -53,6 +53,13 @@
     end
   end
 
+  namespace :api do
```
mssola: Seems unrelated 😉

mssola:
Thanks 👍

PR #1334, opened 2017-07-21
fixed path for vagrant files

saurabhsurana:
recently vagrant files were moved under
examples/development/vagrant


mssola:
Thanks 👍

PR #1348, opened 2017-07-27
js: team module refactoring

vitoravelino:
This will conflict after merging #1341, so lemme know when that happens to rebase.


vitoravelino:
> If I try to create a namespace from the teams page, it tells me that "X team does not exist". Looks like a regression...

It's not a regression because the bug has been there for a while...

The scenario is: after changing team name, the input hidden fields are not being updated and it refers to the old team name. The ideal is to use ids and not the name. We can do that in another PR.

I've fixed for team user earlier but forgot the namespace case. Anyway, both cases are fixed and tests are passing. Let's goooooo! :D

Issue #1360, opened 2017-08-02
Changes to docker compose example


soedar:
I would like to propose some changes to the docker compose examples, and would like to get some feedback on whether these changes are worth implementing:

1. Move more environment variables to the `.env` file. It would allow the `docker-compose.yml` file to be more "standardized", with runtime configuration for each platform be set without modifying the `docker-compose.yml` file. For example, `PORTUS_SECRET_KEY_BASE` can be moved to the .env file.

2. Remove `links`, depending on the minimum supported version of Docker Compose. A relatively recent version of Docker Compose will create a Docker Network associated to the project, with an internal (virtual) DNS. This means that all the containers within the same Docker Network will be able to access each other via their service name.

3. Use nginx to cache static assets instead of serving assets via volume sharing. Currently a volume is shared between the `portus` and `nginx` container, and the nginx container is serving the static assets in the volume. It seems like we can reduce some complexity if we set `RAILS_SERVE_STATIC_FILES=true` for `portus`, and define nginx to cache the static assets returned by the `portus` container. We will incur a one time penalty to the rails server when we initially request for the assets, but imho that is a insignificant penalty.

If this seems okay, I'll work on this and open a PR with these changes.

mssola:
> Move more environment variables to the .env file. It would allow the docker-compose.yml file to be more "standardized", with runtime configuration for each platform be set without modifying the docker-compose.yml file. For example, PORTUS_SECRET_KEY_BASE can be moved to the .env file.

Sounds good to me.

> Remove links, depending on the minimum supported version of Docker Compose. A relatively recent version of Docker Compose will create a Docker Network associated to the project, with an internal (virtual) DNS. This means that all the containers within the same Docker Network will be able to access each other via their service name.

Looks good to me. But when doing it be sure to document the minimum required versions of both Docker and docker-compose (and make sure that they are not the latest release, but some older).

> Use nginx to cache static assets instead of serving assets via volume sharing. Currently a volume is shared between the portus and nginx container, and the nginx container is serving the static assets in the volume. It seems like we can reduce some complexity if we set RAILS_SERVE_STATIC_FILES=true for portus, and define nginx to cache the static assets returned by the portus container. We will incur a one time penalty to the rails server when we initially request for the assets, but imho that is a insignificant penalty.

[...]

PR #1366, opened 2017-08-03
Moved customizable env variables to .env file

soedar:
See #1360


examples/compose/docker-compose.insecure.yml
```diff
@@ -4,23 +4,25 @@ services:
...
 
-      # NGinx is serving the assets instead of Puma. If you want to change this,
-      # uncomment this line.
-      - RAILS_SERVE_STATIC_FILES=true
+      # SSL
+      PORTUS_CHECK_SSL_USAGE_ENABLED: false
+
+      # Since we have no nginx in insecure mode, portus have to
+      # serve the static files
+      RAILS_SERVE_STATIC_FILES: 'true'
```
mssola: Why have you turned this into a string ? Same in the other file

soedar:
It is due to a quirk in the yaml file. YAML file by default will turn a set of boolean value into "True"/"False", and docker-compose will warn (and possibly error) if there is a boolean value in the YAML.

It was okay previous because the environment variables was passed in as an array of environmental strings, which gets interpreted directly. The issue arises when we use the object syntax to pass in env variables. 

PS: I realized that I've missed out changing `PORTUS_CHECK_SSL_USAGE_ENABLED` to use a string instead. Will update this PR 😅

mssola:
Thanks for taking care of this 👍

PR #1391, opened 2017-08-17
js: added table sortable mixin; namespaces using it

### LINKED ISSUE #1373: Make namespace table sorting reachable through URL
After sorting, URL should reflect the same state as table. This way user can refresh the page or copy and paste and the table keeps the same state.

Example: `orderBy=clean_name&sortAsc=true`

Extracted from https://github.com/SUSE/Portus/pull/1371.

vitoravelino:
So far the column sorting was possible but user wasn't able
to share the state of it to someone through copying and pasting url.

Now every sorting action the user makes on the namespaces table,
for example, will create a new url state that can be visited and
navigated back and forth.

Fixes #1373

### DIFF
--- a/app/assets/javascripts/modules/namespaces/components/table.js
+++ b/app/assets/javascripts/modules/namespaces/components/table.js
@@ -1,17 +1,26 @@
-import Vue from 'vue';
 import getProperty from 'lodash/get';
 
 import Comparator from '~/utils/comparator';
+
 import TablePagination from '~/shared/components/table-pagination';
+import TableSortableMixin from '~/shared/mixins/table-sortable';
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/modules/namespaces/components/table.js
```diff
@@ -33,16 +40,16 @@ export default {
     },
 
     filteredNamespaces() {
```
vitoravelino: In the future I want the sorting part of the `filteredNamespaces` to be available from the mixin and not hardcoded as it is right now.

vitoravelino:

PR #1392, opened 2017-08-18
js: added table paginated mixin; namespaces using it

### LINKED ISSUE #1390: Make namespace table pagination reachable through URL
After changing the page, URL should reflect the same state as table. This way user can refresh the page or copy and paste and the table keeps the same state.

Example: `page=2`

Extracted from https://github.com/SUSE/Portus/pull/1371.

vitoravelino:
So far the pagination was possible but user wasn't able
to share the state of it with someone by copying and pasting url.

Now every pagination action the user makes on the namespaces table,
for example, will create a new url state that can be visited and
navigated back and forth.

Fixes #1390


vitoravelino:

PR #1394, opened 2017-08-22
js: added inline validation for namespace's new form

### LINKED ISSUE #1376: Add inline validation for fields in Namespaces#index form
Currently there's no inline feedback validation when for back-end validation. The only way of showing validation is through the floating alert which is not good for this case.

vitoravelino:
The validation so far was fully done on the server side and
the feedback was through the float alert that disappears in 5 seconds.
Not the best experience for the users.

Now, for each field (name and team), we have an inline feedback validation.
User will only see a validation message on the float alert in unexpected
situations.

Fixes #1376

inline-validation

### DIFF
--- a/app/assets/javascripts/modules/namespaces/components/new-form.js
+++ b/app/assets/javascripts/modules/namespaces/components/new-form.js
@@ -1,9 +1,10 @@
 import Vue from 'vue';
 
-import EventBus from '~/utils/eventbus';
+import { required } from 'vuelidate/lib/validators';
 
 import { setTypeahead } from '~/utils/typeahead';
 
+import EventBus from '~/utils/eventbus';
 import Alert from '~/shared/components/alert';
 import FormMixin from '~/shared/mixins/form';
 
@@ -24,8 +25,13 @@ export default {
     return {
       namespace: {
         namespace: {
+          name: '',
           team: this.teamName || '',
         },
+      },
+      timeout: {
+        name: null,
+        team: null,
       },
     };
   },
@@ -37,6 +43,7 @@ export default {
         const name = namespace.attributes.clean_name;
 
         this.toggleForm();
+        this.$v.$reset();
         set(this.namespace, 'namespace', {});
 
         Alert.show(`Namespace '${name}' was created successfully`);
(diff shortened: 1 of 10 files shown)


app/controllers/namespaces_controller.rb
```diff
@@ -11,20 +11,24 @@ class NamespacesController < ApplicationController
...
-        @namespaces = policy_scope(Namespace).order(created_at: :asc)
-
-        accessible_json = serialize_as_json(@namespaces)
-        special_json = serialize_as_json(@special_namespaces)
-        render json: {
-          accessible: accessible_json,
-          special:    special_json
-        }
+    if request.head?
+      check_namespace_by_name if params[:name]
```
vitoravelino: @mssola You probably have something to comment regarding this.

vitoravelino:

app/assets/javascripts/modules/namespaces/components/new-form.js
```diff
@@ -53,12 +60,92 @@ export default {
     },
   },
 
+  validations: {
```
mssola: Instead of doing all this, we should instead call a single endpoint that returns whether the given name is valid. This way:

- We have a single source of truth.
- We have a simpler client-side code.

app/controllers/namespaces_controller.rb
mssola: Yeah, this is not the way to go... Moreover, consider what I just said on the above comment.

I'd create a new endpoint that checks:

1. The availability of the name.
2. The name has a proper format.

This in turn could be extended to other resources (e.g. teams). This can be done with regular Rails for now, but maybe it would be a good starting point for grape on namespaces.

app/assets/javascripts/modules/namespaces/components/new-form.js
vitoravelino: > Instead of doing all this, we should instead call a single endpoint that returns whether the given name is valid.

Do you mean for all the validations? Something like `/namespaces/valid`? How would the response be?

mssola: In my mind, you would have something like this:

- You have a `/status` endpoint for each resource (this also includes the `registry` resource from #1396).
- The response is a JSON that looks like this:

```json
{
  exists: false,
  valid: true,
  msg: ""
}
```

If the resource does not exist (e.g. you are checking that the form contains valid inputs), then `exists` is false and the `valid` and `msg` fields have to be checked. Otherwise, if the resource exists (e.g. checking that the current registry is reachable), then `valid` and `msg` can be safely ignored (e.g. and then the client-side has to show an "this already exists" message or something like that).

All this is just on the top of my head, so I may be missing something 😁

vitoravelino: I'm confused now, are you talking about the `exists` endpoint or the whole validation approach?

vitoravelino: To recap our conversation: we are gonna keep the same behavior (field by field) on the client side but move all the validation to the back-end through a single `/status` endpoint. Right?

I'm just not sure about the `status` name. It can have so many meanings. But that's easy to change later. 😉

mssola: Sure, there are other possibilities for the name, like `/validate` or something like that.

PR #1396, opened 2017-08-28
Added some registry validation and status

### LINKED ISSUE #1355: Tell the user whether the configured registry is reachable or not
In either the `edit` or the `index` page there could be a hint telling the user whether the configured registry is available or not:

- This is useful in the `index` page to check whether the registry is available or not, but it's also a re-assurance when you have configured Portus with the `signup_enabled` option disabled, and so you configured the registry manually through the CLI.
- In the `edit` page, the frontend could tell you whether the entered registry is reachable or not before actually submitting the button.

This could all be done either through a button or a link that the user has to click, or it could all be done in the frontend automatically (trying to not ping to registry too much).

vitoravelino:
The registry validation was done completely on the server side. With
this, we bring some validation to the client side to antecipate and improve
the user's experience.

Also adds a hint in the registries#index pages to help the user know if the
registry that was set is reachable or not. The reachable status is cached for
30 minutes unless the user updates the registry.

Fixes #1355

### Screenshots

screenshot 2017-08-28 17 02 10
screenshot 2017-08-28 17 02 26
screenshot 2017-08-28 17 05 51
screenshot 2017-08-28 17 05 07
screenshot 2017-08-28 17 06 41

### DIFF
--- a/app/controllers/admin/registries_controller.rb
+++ b/app/controllers/admin/registries_controller.rb
@@ -7,6 +7,8 @@ class Admin::RegistriesController < Admin::BaseController
   # GET /admin/registries/
   def index
     @registries = Registry.all
+
+    redirect_to new_admin_registry_url unless Registry.any?
   end
 
   # GET /admin/registries/new
@@ -25,17 +27,9 @@ class Admin::RegistriesController < Admin::BaseController
     @registry = Registry.new(create_params)
 
     # Check the reachability of the registry.
-    unless params[:force]
-      msg = @registry.reachable?
-      unless msg.empty?
-        logger.info "\nRegistry not reachable:\n#{@registry.inspect}\n#{msg}\n"
-        msg = "#{msg} You can skip this check by clicking on the \"Skip remote checks\" checkbox."
-        hsh = { name: @registry.name, hostname: @registry.hostname,
-          use_ssl: @registry.use_ssl, external_hostname: @registry.external_hostname }
-        redirect_to new_admin_registry_path(hsh), alert: msg
-        return
-      end
-    end
+
+    check_reachability("new") unless params[:force]
+    return if @unreachable
 
     if @registry.save
       Namespace.update_all(registry_id: @registry.id)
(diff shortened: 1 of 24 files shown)


mssola:
See [my comment](https://github.com/SUSE/Portus/pull/1394#discussion_r136061253).

vitoravelino:
@mssola This is ready to be reviewed. I've had to change a little bit the registry endpoint to be more consistent with what would rails would return from its validation. 

Also, a flaky test that wasn't happening before appeared. I'll investigate it tomorrow.

vitoravelino:
Flaky test fixed! 👍

app/controllers/admin/registries_controller.rb
```diff
@@ -7,11 +7,31 @@ class Admin::RegistriesController < Admin::BaseController
...
   end
 
   # GET /admin/registries/new
   def new
     @registry = Registry.new
+    @can_change_hostname = true
+  end
+
+  # GET /admin/registries/reachable
+  def reachable
```
mssola: Are you really using this ? (you are using the API on the client side, right?)

app/controllers/admin/registries_controller.rb
```diff
@@ -60,8 +79,30 @@ def edit
   # given registry. This might change in the future.
   def update
     @registry = Registry.find(params[:id])
-    @registry.update_attributes(update_params)
-    redirect_to admin_registries_path, notice: "Registry updated successfully!"
+    @registry.assign_attributes(update_params)
+    @can_change_hostname = !Repository.any?
+
+    # Check the reachability of the registry.
+    unless params[:force]
```
mssola: There is no `force` parameter on the edit form, so this shouldn't apply here.

app/controllers/admin/registries_controller.rb
```diff
@@ -60,8 +79,30 @@ def edit
...
+      msg = @registry.reachable?
+      unless msg.empty?
+        logger.info "\nRegistry not reachable:\n#{@registry.inspect}\n#{msg}\n"
```
mssola: can we merge this code with the one on the create method in an easy way ? I feel it's basically the same thing...

config/routes.rb
```diff
@@ -70,7 +70,9 @@
   namespace :admin do
     resources :activities, only: [:index]
     resources :dashboard, only: [:index]
-    resources :registries, except: [:show, :destroy]
+    resources :registries, except: [:show, :destroy] do
+      get "reachable", on: :collection
```
mssola: Again, this is not needed.

app/controllers/admin/registries_controller.rb
vitoravelino: I was but now we have the validate endpoint.

app/controllers/admin/registries_controller.rb
vitoravelino: We didn't but when I started this I saw the opportunity to improve the edit form experience the same way I did for the create form. So we have validation on the hostname value too. What do you think?

app/controllers/admin/registries_controller.rb
mssola: Yeah, please use the validate endpoint instead 👍

app/controllers/admin/registries_controller.rb
mssola: I think it's alright, but the `force` check point was meant more like "I can't connect to my registry now and I don't know why, but I want to test Portus out anyways". Kind of bad but it might come in handy when the registry is not up when you are configuring Portus but that's really the config for it.

So, with this in mind, I don't think it makes much sense to have this same checkbox on the edit form (it does make a lot of sense to check whether is reachable or not of course, so I appreciate your changes on this regard to the `update` method 👍).

vitoravelino: It makes sense. So no force on edit but validation, right? I'll change that.

vitoravelino:
@mssola This is good to be reviewed.

PR #1419, opened 2017-09-13
Sort tags by updated_at date not by created_at

coder-hugo:
If a tag will be overwritten its created_at date stays the same. So sorting the tags by this date results in a wrong order. The updated_at date will be changed anytime a tag is overwritten. So such a tag will appear at the top of the tag list if they are sorted by this date.


vitoravelino:
LGTM. I'm not sure if `created_at` was used on purpose or not. What do think, @mssola?

mssola:
Thanks @coder-hugo

PR #1432, opened 2017-09-22
namespace: removed more legacy js and controller code

vitoravelino:


app/controllers/namespaces_controller.rb
```diff
@@ -28,6 +28,11 @@ def show
 
     authorize @namespace
     @repositories = @namespace.repositories.page(params[:page])
+    @namespace_entity = API::Entities::Namespaces.represent(
```
mssola: I think that it would be better to move this entirely to the grape API, and then simply remove the code from this controller in this regard.

lib/api/entities.rb
```diff
@@ -97,6 +97,12 @@ class Namespaces < Grape::Entity
       expose :id, documentation: { type: Integer, desc: "Namespace ID" }
       expose :clean_name, as: :name, documentation: { type: String, desc: "Namespace name" }
       expose :created_at, :updated_at, documentation: { type: DateTime }
+      expose :description, documentation: {
```
mssola: I'm not entirely sure about this. Maybe for this version of the API we can leave this as internal ?

vitoravelino: Maybe, I dunno. I added it because it was necessary to display the description on the `show` page. I think we can keep this way, maybe it's useful for someone. It's an optional field, so if not used it returns empty. No big deal. Up to you.

app/controllers/namespaces_controller.rb
vitoravelino: Are we gonna start moving the non GET actions to the API?

mssola: I think it would be a good place to start :-) I'll start with it asap

lib/api/entities.rb
mssola: Sounds reasonable. Let's leave it like this :+1:

app/controllers/namespaces_controller.rb
vitoravelino: I'll try then. It's good for me to practice and learn something that I'm not too familiar with. Don't worry! :)

lib/api/v1/namespaces.rb
```diff
@@ -27,6 +27,35 @@ class Namespaces < Grape::API
                   type:         current_type
         end
 
+        desc "Create a namespace",
```
vitoravelino: @mssola Please review this!

PR #1434, opened 2017-09-26
teams: migrated index page to vue components

vitoravelino:
I'll investigate how we can improve the bootstrap of the components in another PR to not block the current work on this one.


app/controllers/teams_controller.rb
```diff
@@ -33,13 +38,22 @@ def create
...
     @team.owners << current_user
 
-    if @team.save
-      @team.create_activity(:create,
-                            owner:      current_user,
-                            parameters: { team: @team.name })
-      respond_with(@team)
-    else
-      respond_with @team.errors, status: :unprocessable_entity
+    respond_to do |format|
```
vitoravelino: @mssola Should I move this to the API too?

mssola: I was thinking on it but I didn't want to annoy you more with API thingies 😄 I don't have that much spare time in SUSEcon, but I'll try to open up a PR with it asap 👍

vitoravelino: Same comment as the other. 😃

lib/api/v1/teams.rb
```diff
@@ -22,6 +22,30 @@ class Teams < Grape::API
           present policy_scope(Team), with: API::Entities::Teams
         end
 
+        desc "Create a team",
```
vitoravelino: @mssola Please review this!

PR #1443, opened 2017-10-02
repositories: migrated to vue and exposed on menu

### LINKED ISSUE #1420: In Repositories the [Previous,1 ,2, N and Next] Pagination link doesn't work (exist)
The Previous, Next  and 1,2,3 pagination isn't working on repositories. I only tried repositories maybe its also not working in other views.
 
Not sure if this was already addressed, if so please close.

vitoravelino:
The default pagination limit was also changed to `10`.

Fixes #1420

### DIFF
--- a/app/assets/javascripts/modules/repositories/components/table-row.js
+++ b/app/assets/javascripts/modules/repositories/components/table-row.js
@@ -1,0 +1,25 @@
+import moment from 'moment';
+
+export default {
+  template: '#js-repository-table-row-tmpl',
+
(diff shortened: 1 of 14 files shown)


vitoravelino:
We can also reuse this on the namespaces#show page where we list all the repositories. I'll make the changes. Thanks!

vitoravelino:
Just pushed the request changes. I've added `updated_at` instead of `created_at`. I think it makes more sense, no? Lemme know.

PR #1444, opened 2017-10-04
Fixed the namespace and team name clashes

### LINKED ISSUE #1269: Cannot add back a "removed user" because of pre-existing namespace
## BUG EXPLANATION
Hello Portus-devs,
During one of our tests in our Portus instance, I created a new user and then removed it.
Of course that left the namespace in place as part of other bug/feature already discussed elsewhere.

The "problem" is that if I try to add back that same user, I get back the very logical error/consequence

`Username cannot be used: there is already a namespace named 'test-username'`

## STEPS TO REPRODUCE
0- login to Portus and go in the admin section
1- create a new user
2- delete the user created at step #1
3- try to add again the user created at step #1

## VERSION INFO
Portus version 2.2 running from the official Docker containers (https://hub.docker.com/r/opensuse/portus/)

vitoravelino:
Currently we have a problem when a user is removed and added back
to Portus. The team and namespace associated with that user are not
removed. So, when an admin tries to create a user or an user tries
to create an account, an error mentioning that there's already a
namespace with that username is shown.

With this patch we avoid that by adding an incremental number to the
end of the name. For example, if username is `brian` and there's
already a namespace/team with that name, `brian0` namespace and team
will be created instead. If `brian` gets removed again, next time it
will be `brian1` for both namespace and team.

Remembering that we should remove this once we implement the
namespace/team removal.

Fixes #1269

### DIFF
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -93,16 +93,16 @@ class Namespace < ActiveRecord::Base
     [namespace, name, registry]
   end
 
-  # Tries to transform the given name to a valid namespace name. If the name is
-  # already valid, then it's returned untouched. Otherwise, if the name cannot
-  # be turned into a valid namespace name, then nil is returned.
+  # Tries to transform the given name to a valid namespace name without
+  # clashing with existent namespaces.
+  # If the name cannot be turned into a valid namespace name,
+  # then nil is returned.
+  # If the name is valid, checks if it clashes with others namespaces and
+  # finds one until it's not being used and returns it.
   def self.make_valid(name)
-    return name if name =~ NAME_REGEXP
-
     # One common case is LDAP and case sensitivity. With this in mind, try to
     # downcase everything and see if now it's fine.
     name = name.downcase
-    return name if name =~ NAME_REGEXP
 
     # Let's strip extra characters from the beginning and end.
     first = name.index(/[a-z0-9]/)
@@ -120,6 +120,19 @@ class Namespace < ActiveRecord::Base
     name = final[0..MAX_NAME_LENGTH]
 
     return nil if name !~ NAME_REGEXP
+
+    # To avoid any name conflict we append an incremental number to the end
+    # of the name returns it as the name that will be used on both Namespace
+    # and Team on the User#create_personal_namespace! method
+    # TODO: workaround until we implement the namespace/team removal
+    increment = 0
+    original_name = name
+    while Namespace.exists?(name: name)
+      name = "#{original_name}#{increment}"
+      increment += 1
+      break if increment > 1000
+    end
+
     name
   end
 
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -51,4 +51,23 @@ class Team < ActiveRecord::Base
   def self.search_from_query(valid_teams, query)
     all_non_special.where(id: valid_teams).where(arel_table[:name].matches(query))
   end
+
+  # Tries to transform the given name to a valid team name without
+  # clashing with existent teams.
+  # Checks if it clashes with others teams and finds one until it's not
+  # being used and returns it.
+  def self.make_valid(name)
+    # To avoid any name conflict we append an incremental number to the end
+    # of the name returns it as the name that will be used on both Namespace
+    # and Team on the User#create_personal_namespace! method
+    # TODO: workaround until we implement the namespace/team removal
+    increment = 0
+    original_name = name
+    while Team.exists?(name: name)
+      name = "#{original_name}#{increment}"
+      increment += 1
+    end
+
+    name
+  end
 end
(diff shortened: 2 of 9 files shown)


vitoravelino:
@mssola I'm not sure if the approach I went with is the best. Lemme know what you think.

vitoravelino:
Just realized that a feature test is missing. I'll do that now.

vitoravelino:
Done! 😃

app/models/namespace.rb
```diff
@@ -120,6 +120,20 @@ def self.make_valid(name)
     name = final[0..MAX_NAME_LENGTH]
 
     return nil if name !~ NAME_REGEXP
+
+    # To avoid any name conflict we append an incremental number to the end
+    # of the name returns it as the name that will be used on both Namespace
+    # and Team on the User#create_personal_namespace! method
+    # TODO: workaround until we implement the namespace/team removal
+    if name =~ NAME_REGEXP
```
mssola: Since you have a `return nil if name !~ NAME_REGEXP` as the previous statement, I think it's safe to skip this check.

app/models/namespace.rb
```diff
@@ -120,6 +120,20 @@ def self.make_valid(name)
...
+      increment = 0
+      original_name = name
+      while Namespace.exists?(name: name)
+        name = "#{original_name}#{increment}"
```
mssola: For the first time around, a `name0` will be created. I think it would be more elegant to have a `name` for the first case.

mssola: Plus, just to avoid completely messed up scenarios, I'd give this loop a top value. That is:

```ruby
found = false
100.times do |idx|
  if Namespace.exists?(name: name)
    name = "#{original_name}#{idx + 1}"
  else
    found = true
    break
  end
end

# Return nil for crazy scenarios...
name if found
```

This would also preserve the name for the 0 case, as mentioned in my previous comment 😉

app/models/team.rb
```diff
@@ -51,4 +51,23 @@ def member_ids
   def self.search_from_query(valid_teams, query)
     all_non_special.where(id: valid_teams).where(arel_table[:name].matches(query))
   end
+
+  # Tries to transform the given name to a valid team name without
+  # clashing with existent teams.
+  # Checks if it clashes with others teams and finds one until it's not
+  # being used and returns it.
+  def self.make_valid(name)
```
mssola: Same as for the namespace version.

app/models/user.rb
```diff
@@ -86,13 +86,6 @@ def private_namespace_and_team_available
     if ns.nil?
       errors.add(:username, "'#{username}' cannot be transformed into a " \
         "valid namespace name")
-    elsif Namespace.exists?(name: ns)
-      clar = (ns != username) ? " (modified so it's valid)" : ""
-      errors.add(:username, "cannot be used: there is already a namespace " \
-        "named '#{ns}'#{clar}")
-    elsif Team.exists?(name: username)
-      errors.add(:username, "cannot be used: there is already a team named " \
-        "like this")
```
mssola: Do these checks really hurt ? If not, then I'd leave them just to be sure.

app/models/namespace.rb
vitoravelino: > For the first time around, a name0 will be created. I think it would be more elegant to have a name for the first case.

It won't because it will not enter the `while` loop since `exists?` will return false. There's a test ensuring that what you said won't happen.

app/models/team.rb
vitoravelino: Same here.

app/models/user.rb
vitoravelino: The problem is that it will never happen now that there's no name clashes. I know this method is called before `create_personal_namespace!` but it doesn't make sense to show an error about name conflicts if we are trying to solve the conflicts.

app/models/namespace.rb
mssola: Duh... makes sense 😁 That being said, could you put it inside of a `times` block ? just to be safe.

app/models/user.rb
mssola: Gotcha

app/models/namespace.rb
vitoravelino: Do you mean to avoid infinite loop if mysql/activerecord fails by never returning `true`?

mssola: That, or even that the same user got removed and re-added 100 times 😁 (yeah, my pessimistic me is telling me that it will happen ...)

app/models/namespace.rb
```diff
@@ -120,6 +120,19 @@ def self.make_valid(name)
...
+    # To avoid any name conflict we append an incremental number to the end
+    # of the name returns it as the name that will be used on both Namespace
+    # and Team on the User#create_personal_namespace! method
+    # TODO: workaround until we implement the namespace/team removal
+    increment = 0
+    original_name = name
+    while Namespace.exists?(name: name)
+      name = "#{original_name}#{increment}"
+      increment += 1
+      break if increment > 1000
```
vitoravelino: @mssola I didn't go with the `times` approach. Just added a `break if` here. Is it okay?

Issue #1470, opened 2017-10-23
Raise exception on migrate.rake


mssola:
There are some deprecated methods since last version. Preparing for 2.3, we should either remove this methods, or raise a proper exception with info for this release.

PR #1471, opened 2017-10-23
refa: reduce usage of send

salzig:
saw issue #1470 and found `lib/portus/migrate.rb` a bit hard to follow. 

Hope you don't mind the refactoring.

### DIFF
--- a/lib/portus/deprecation_error.rb
+++ b/lib/portus/deprecation_error.rb
@@ -1,0 +1,9 @@
+module Portus
+  # DeprecationError is the exception to be raised when certain functionality has been deprecated
+  # in the latest version and we want to provide migration instructions
+  class DeprecationError < StandardError
+    def to_s
+      "[DEPRECATED] " + super
+    end
+  end
+end
--- a/lib/portus/migrate.rb
+++ b/lib/portus/migrate.rb
@@ -1,16 +1,14 @@
+# rubocop:disable Metrics/CyclomaticComplexity
+# rubocop:disable Metrics/PerceivedComplexity
 module Portus
   # The Portus::Migrate module implements some methods that are used to handle a
   # change from different versions of Portus where there is an old (and
   # deprecated) way of doing things and a preferred new one.
   module Migrate
-    # This method has to be used when migrating from an 'x.minutes' string
-    # configuration value (e.g. the `jwt_expiration_time` value).
-    #
     # The duration parameter is the given original value, and the default
     # parameter is an Integer with the value to be used whenever there is
     # an error. Otherwise, if the given duration is not an integer and has
-    # a good (old) string format, then eval will be used just as we did
-    # it before.
+    # a good (old) string format, then a deprecation error will be raised.
     #
     # TODO: (mssola) remove in the next version.
     def self.from_humanized_time(duration, default)
@@ -22,19 +20,16 @@ module Portus
       # If we got an empty String then return the default value.
       return default.minutes if duration.empty?
 
-      raw_value, method = duration.split('.')
+      raw_value, method = duration.split(".")
       value = raw_value.to_i
 
       # When we found a callable method, notify about deprecation
       if !method.nil? && value.respond_to?(method)
-        Rails.logger.tagged("deprecated") do
-          Rails.logger.warn "The 'x.minutes' format is deprecated for configuration values such " \
-                            "as `jwt_expiration_time`. From now on these values are expected to " \
-                            "be integers representing minutes."
-        end
-        value.send(method)
+        raise DeprecationError, "The 'x.minutes' format is deprecated for configuration values " \
+                                "such as `jwt_expiration_time`. From now on these values are " \
+                                "expected to be integers representing minutes."
       elsif value > 0
-        # If it's a string containing a positive number, then convert it and return it. 
+        # If it's a string containing a positive number, then convert it and return it.
         value.minutes
       else
         # Otherwise, if it has a bad (old) format, then just return the default.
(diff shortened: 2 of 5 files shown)


lib/portus/migrate.rb
```diff
@@ -14,29 +14,37 @@ module Migrate
...
-        Rails.logger.warn "The 'x.minutes' format is deprecated for configuration values such " \
-                          "as `jwt_expiration_time`. From now on these values are expected to " \
-                          "be integers representing minutes."
+      raw_value, method = duration.split('.')
+      value = raw_value.to_i
+
+      # When we found a callable method, notify about deprecation
+      unless method.nil?
+        if value.respond_to? method
+          Rails.logger.tagged("deprecated") do
```
mssola: Since you are fixing #1470, raise an exception instead of logging it. I'd create a `DeprecationError` exception class, that can be called like this:

```ruby
raise DeprecationError, "Format 'x.minutes' is no longer supported: time values are expected to be given in minutes."
```

This class would add an "[DEPRECATED]" prefix or something like that to the message.

lib/portus/migrate.rb
```diff
@@ -14,29 +14,37 @@ module Migrate
...
-      Rails.logger.tagged("deprecated") do
-        Rails.logger.warn "The 'x.minutes' format is deprecated for configuration values such " \
-                          "as `jwt_expiration_time`. From now on these values are expected to " \
-                          "be integers representing minutes."
+      raw_value, method = duration.split('.')
+      value = raw_value.to_i
+
+      # When we found a callable method, notify about deprecation
+      unless method.nil?
+        if value.respond_to? method
```
mssola: rubocop will complain about this (it doesn't now because CI is broken, working on that...). Instead you should use a guard clause:

```ruby
return unless value.respond_to? method
```

lib/portus/migrate.rb
```diff
@@ -14,29 +14,37 @@ module Migrate
...
+        end
       end
 
-      # rubocop:disable Lint/Eval
-      eval(duration)
-      # rubocop:enable Lint/Eval
+      # If it's a string containing a positive number, then convert it and return
+      # it. Otherwise, if it has a bad (old) format, then just return the default.
+      if value > 0
+        return value.minutes
```
mssola: Don't use return, simply writing `value.minutes` is enough 😉

lib/portus/migrate.rb
```diff
@@ -14,29 +14,37 @@ module Migrate
...
+      else
+        Rails.logger.warn "Unsupported time format (#{duration}), fallback to default."
+        return default.minutes
```
mssola: Same here

lib/portus/migrate.rb
mssola: While you are at it, do the same for the `self.registry_config` method 😁

salzig: wanted to do the refactoring first, as it's kinda hard to read right now. Deprecation Exception would be added in another pull request afterwards.

lib/portus/migrate.rb
salzig: you're right.

lib/portus/migrate.rb
salzig: you're right.

lib/portus/migrate.rb
salzig: btw, is the deprecation only because of the usage of `eval`?

mssola:
> wanted to do the refactoring first, as it's kinda hard to read right now. Deprecation Exception would be added in another pull request afterwards.

I understand that you wanted to do the refactoring first, but I think that you can also fix that issue on this same PR (so also implement the error class). Then this PR (when squashed) would contain two commits:

1. The refactoring itself.
2. The addition of the exception class.

I don't think it would make the PR too complicated to follow honestly.

> btw, is the deprecation only because of the previous usage of eval?

Not really. The deprecation came with the realization that it was just too flexible, and that a mere integer representing time in minutes would suffice. This in turn avoids us of an `eval` call, but it wasn't the primary motivation.

mssola:
@salzig just fixed the CI issues, and upgraded quite some stuff. Due to rubocop's new issues, I had to update the file you were touching and hence the conflict. Sorry about that 😅

salzig:
will rebase later.

lib/portus/deprecation_error.rb
```diff
@@ -0,0 +1,7 @@
+module Portus
+  class DeprecationError < StandardError
```
mssola: Could you add some documentation ? Something simple like: "DeprecationError is the exception to be raised when certain functionality has been deprecated in the latest version and we want to provide migration instructions".

lib/portus/migrate.rb
```diff
@@ -2,6 +2,7 @@ module Portus
   # The Portus::Migrate module implements some methods that are used to handle a
   # change from different versions of Portus where there is an old (and
   # deprecated) way of doing things and a preferred new one.
+
```
mssola: Remove this extra line

spec/lib/portus/deprecation_error_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+describe Portus::DeprecationError do
+  let(:message) { "example" }
+  subject(:error) { described_class.new(message) }
+
+  context "#to_s" do
+    subject { error.to_s }
```
mssola: I think it would be simpler to have just one test like this:

```ruby
it "prefixes the message with [DEPRECATED]" do
  is_expected.to eq("[DEPRECATED] #{message}")
end
```

salzig: every project is different :)

mssola:
Also on the second commit, add a "Fixes #1470" on the commit message body 😉

salzig:
the commit messages contain the progress (commits before squash). Do you want them to be removed?

mssola:
Oh I see. Yes please, remove those messages. Just leave a single and cohesive message explaining what's in the commit.

salzig:
any further change required?

PR #1474, opened 2017-10-24
Update Model and Route annotations and make them concise with defaults

Vad1mo:
Update Model and Route annotation and make them concise with default ruby annotate_models.

its related to #1324

Signed-off-by: Vadim Bauer <bauer.vadim@gmail.com>

### DIFF
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -7,9 +7,10 @@
 #  created_at  :datetime         not null
 #  updated_at  :datetime         not null
 #  team_id     :integer
-#  public      :boolean          default("0")
-#  registry_id :integer
+#  registry_id :integer          not null
 #  global      :boolean          default("0")
+#  description :text(65535)
+#  visibility  :integer
 #
 # Indexes
 #
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,108 @@
+# == Route Map
+#
+# [Mailer config] Host:     portus.test.lan
+# [Mailer config] Protocol: http://
+#                           Prefix Verb     URI Pattern                                                             Controller#Action
+#                            error GET      /errors/:id(.:format)                                                   errors#show
+#                                  GET      /teams/:id/typeahead/:query(.:format)                                   teams#typeahead {:format=>"json"}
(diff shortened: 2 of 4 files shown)


Vad1mo:
In a deluxe version one could update the travis ci file and check if model  changed and if yes fail the build. This would make sure that the annotations are always in sync.

config/routes.rb
```diff
@@ -1,3 +1,108 @@
+# == Route Map
+#
+# [Mailer config] Host:     portus.test.lan
+# [Mailer config] Protocol: http://
```
mssola: Is there any way to avoid these lines about the mailer config ?

app/models/namespace.rb
```diff
@@ -7,9 +7,10 @@
 #  created_at  :datetime         not null
 #  updated_at  :datetime         not null
 #  team_id     :integer
-#  public      :boolean          default("0")
-#  registry_id :integer
+#  registry_id :integer          not null
 #  global      :boolean          default("0")
+#  description :text(65535)
+#  visibility  :integer
```
mssola: Funnily enough, on my development environment nothing gets annotated on migration (and it used to happen), but if I call the rake task directly it will work.

*Update*: it looks like upgrading the `annotate` gem fixes my issue. I'll do this in a PR.

db/schema.rb
```diff
@@ -108,6 +108,15 @@
   add_index "repositories", ["name", "namespace_id"], name: "index_repositories_on_name_and_namespace_id", unique: true, using: :btree
   add_index "repositories", ["namespace_id"], name: "index_repositories_on_namespace_id", using: :btree
 
+  create_table "settings", force: :cascade do |t|
```
mssola: What is this ?

mssola:
> In a deluxe version one could update the travis ci file and check if model changed and if yes fail the build. This would make sure that the annotations are always in sync.

I'll do this in another PR.

app/models/namespace.rb
Vad1mo: it seems to be tricky keeping it up to date.

mssola:
LGTM, merging... I'm continuing this on #1476. Thanks for this 👏

Issue #1473, opened 2017-10-24
Raise an exception for environment variables that have changed


mssola:
Some environment variables (e.g. some on the DB configuration) have changed of name. We should add some checks in `config/initializers/environment_check.rb`.

salzig:
Is there any variable i missed?

mssola:
I think that was everything, thanks for this 👍

PR #1480, opened 2017-10-25
feat: raise DeprecationError for old Database Configuration

salzig:
@mssola hope this was meant with #1473.

### DIFF
--- a/config/initializers/environment_check.rb
+++ b/config/initializers/environment_check.rb
@@ -13,6 +13,17 @@ def mandatory_secret!(env, value)
         "mandatory in production because it's needed for the generation of secrets."
 end
 
+# Database Environment Variables have changed.
+if ENV["PORTUS_PRODUCTION_HOST"].present? && ENV["PORTUS_DB_HOST"].blank? ||
+    ENV["PORTUS_PRODUCTION_USERNAME"].present? && ENV["PORTUS_DB_USERNAME"].blank? ||
+    ENV["PORTUS_PRODUCTION_PASSWORD"].present? && ENV["PORTUS_DB_PASSWORD"].blank? ||
+    ENV["PORTUS_PRODUCTION_DATABASE"].present? && ENV["PORTUS_DB_DATABASE"].blank?
+  raise Portus::DeprecationError, "The database configuration has been extended. With this " \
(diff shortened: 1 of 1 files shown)


config/initializers/environment_check.rb
```diff
@@ -13,6 +13,16 @@ def mandatory_secret!(env, value)
         "mandatory in production because it's needed for the generation of secrets."
 end
 
+# Database Environment Variables have changed.
+if ENV["PORTUS_PRODUCTION_HOST"].present? && ENV["PORTUS_DB_HOST"].blank? ||
+    ENV["PORTUS_PRODUCTION_USERNAME"].present? && ENV["PORTUS_DB_USERNAME"].blank? ||
+    ENV["PORTUS_PRODUCTION_PASSWORD"].present? && ENV["PORTUS_DB_PASSWORD"].blank? ||
+    ENV["PORTUS_PRODUCTION_DATABASE"].present? && ENV["PORTUS_DB_DATABASE"].blank?
```
mssola: This could be extended to any environment. As in, you can replace

```ruby
ENV["PORTUS_PRODUCTION_HOST"]
```

by:

```ruby
ENV["PORTUS_#{Rails.env.upcase}_HOST"]
```

And so on.

config/initializers/environment_check.rb
```diff
@@ -13,6 +13,16 @@ def mandatory_secret!(env, value)
...
+  raise Portus::DeprecationError, "The database configuration has been extended to also support " \
+  "PostgreSQL as the database. With this change the Environment Variable Names have also been " \
+  "changed. Please consolidate config/database.yml for the new names."
```
mssola: I would avoid explaining the PostgreSQL part, since we've also done it to simplify the naming of these environment variables. I'd simply instruct that now environment variables that started with `PORTUS_#{Rails.env.upcase}_`, now they should start simply with `PORTUS_DB_`.

salzig: according to af1b8b6ca7 staging, development and test where previously hardcoded.

config/initializers/environment_check.rb
salzig: according to af1b8b6 staging, development and test where previously hardcoded.

PR #1485, opened 2017-10-26
feat: offline polyfills

### LINKED ISSUE #1482: JS Polyfills Offline
I'm currently running Portus in a disconnected environment and experiencing some client side weirdness that I think is down to a lack of the polyfills loaded from https://github.com/SUSE/Portus/blob/master/app/views/layouts/application.html.slim#L33 as I've got no connection to polyfill.io. Is there a way we can build some of the required polyfills in or does it specifically 
 require the browser detection features of polyfill.io?

salzig:
Fixes #1482

### DIFF
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -15,6 +15,7 @@ import 'vendor/lifeitup_layout';
 
 import './bootstrap';
 import './vue-shared';
+import './polyfill';
 
 // modules
 import './modules/admin/registries';
--- a/app/assets/javascripts/polyfill.js
+++ b/app/assets/javascripts/polyfill.js
@@ -1,0 +1,3 @@
+require('core-js/fn/array/some');
+require('core-js/fn/array/from');
+require('core-js/fn/array/find-index');
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -30,7 +30,6 @@ html
 
     javascript:
       window.API_ROOT_URL = '#{root_url}';
-    script src="//cdn.polyfill.io/v2/polyfill.js?features=Array.prototype.some,Array.prototype.findIndex,Array.from"
     = javascript_include_tag(*webpack_asset_paths("application"))
     = yield :js_header
 
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "dependencies": {
     "bootstrap-sass": "^3.3.6",
     "compression-webpack-plugin": "^0.3.2",
+    "core-js": "^2.5.1",
     "css-loader": "^0.28.1",
     "jquery": "^2.2.1",
     "jquery-ujs": "^1.2.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1059,6 +1059,10 @@ core-js@^2.4.0:
   version "2.4.1"
   resolved "https://registry.yarnpkg.com/core-js/-/core-js-2.4.1.tgz#4de911e667b0eae9124e34254b53aea6fc618d3e"
 
+core-js@^2.5.1:
+  version "2.5.1"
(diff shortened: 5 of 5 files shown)


salzig:
Can reproduce the CI problem with poltergeist. But i'm unable to reproduce in a browser. Any clues whats happening there?

salzig:
`webpack-polyfill-injector` uses the exact same polyfills as `polyfill.io`. Thought that would be a good idea, but maybe it wasn't. Will have a look into this later again.

vitoravelino:
> webpack-polyfill-injector uses the exact same polyfills as polyfill.io. Thought that would be a good idea, but maybe it wasn't. Will have a look into this later again.

I'm investigating why it's not working at the moment. I have a guess but I want to confirm it.

salzig:
Oh, i run into the same error message in a completely different project (gitlab+gitlab-ci+poltergeist, all coffeescript, no vue or polyfill). Maybe it's related to poltergeist?

vitoravelino:
> Maybe I'm wrong (and please @vitoravelino correct me if so), but I think tthat you shouldn't inject it on the webpack configuration. Rather, the way to go would be to create a new file that then the main.js file imports. This file would import the needed functions. I believe that these functions can be taken from core-js/fn/array.

I didn't want to go with third-party libraries because we already have something that is supported on the latest browsers IIRC.

I kinda liked the adopted approach. It's not something that we are gonna be revisiting often and it's also easy to get rid of when we feel safe.

vitoravelino:
> Oh, i run into the same error message in a completely different project (gitlab+gitlab-ci+poltergeist, all coffeescript, no vue or polyfill). Maybe it's related to poltergeist?

My guess was wrong. 😞 I thought that somehow the code wasn't being injected. But I put some logs and verified that the code is included and is being ran.

Nevertheless, I found the issue. What this webpack plugin does is: 

- wraps all the application code
- checks if browsers needs any polyfill
- if it doesn't need, just run the application code (keeps the run in sync)
- otherwise injects a script tag with the polyfill 
- after the polyfill script is loaded, runs the wrapped application code (run becomes async)

Well, in theory almost nothing changes except for this sync to async run. The problem is that there's some legacy code on the `admin/registries/_form.html.slim` view and `window.$` is returning undefined. 

Why? Because when the browser needs the polyfill, the run of application code now becomes async because it relies on the 'load' event of the script element and there's a really small chance of `window.$` being executed before the code on the bottom of the page.

So, my suggestion is to move the legacy jQuery (just the on click event) from the template to the `admin/registries/components/form.js` inside a `mounted()` callback like:

```js
  mounted() {
    $('body').on('click', '#advancedButton', function () {
      $(this).text($(this).text() === 'Show Advanced' ? 'Hide Advanced' : 'Show Advanced');
    });
  },
```

Tested locally and it worked. You can revert your last commit (e2c43ed) too. Thanks! 😄

salzig:
@vitoravelino found a similar solution after digging for a while in another project. There it was a javascript file not available in every environment, but required in every environment. The error message was the same, but as it seems the problem was different. 

Will cleanup this merge request later the day. Thanks!

vitoravelino:
> Maybe I'm wrong (and please @vitoravelino correct me if so), but I think tthat you shouldn't inject it on the webpack configuration. Rather, the way to go would be to create a new file that then the main.js file imports. This file would import the needed functions. I believe that these functions can be taken from core-js/fn/array.

@salzig, @mssola pinged me in private and explained what he meant with his statement above. I misinterpreted that and I'm sorry for the confusion. Your solution works and we are thankful for it. But I agree with him that having just a bunch of `import` on the code is simpler and more friendly than touching webpack config (even knowing that the solution webpack solution is still simple).

So, I suggest you to use `core-js` polyfills. Just create a `polyfills.js` file on `javascripts` folder and import the 3 corresponding modules for the methods we need. In the `main.js`, `import 'polyfills'` should be the first thing.

Sorry for the inconvenience and thank you for your time!

Thanks for backing me up on this, @mssola! 😄

PR #1490, opened 2017-10-29
feat: add peek for development

### LINKED ISSUE #1462: Add peek into the development environment
The [peek](https://github.com/peek/peek) gem provides a framework of utilities that can come in handy to check the performance of various components. This can be useful 👍

salzig:
Fixes #1462

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,14 @@ unless packaging?
     gem "guard", require: false
     gem "guard-rspec", require: false
     gem "guard-rubocop", require: false
+    gem "jquery"
+    gem "jquery-rails"
+    gem "peek"
+    gem "peek-gc"
+    gem "peek-git"
+    gem "peek-mysql2"
+    gem "peek-performance_bar"
+    gem "peek-pg"
     gem "pry-rails"
     gem "quiet_assets"
     gem "rack-mini-profiler", require: false
--- a/app/assets/javascripts/debug.js
+++ b/app/assets/javascripts/debug.js
@@ -1,0 +1,4 @@
+//= require jquery
+//= require jquery_ujs
+//= require peek
+//= require peek/views/performance_bar
--- a/app/helpers/peek_helper.rb
+++ b/app/helpers/peek_helper.rb
@@ -1,0 +1,13 @@
+module PeekHelper
+  def peek_available?
+    defined? Peek
+  end
+
+  def peek_enabled?
+    peek_available? && Rails.env.development?
+  end
+
+  def render_peek
+    render "peek/bar" if peek_enabled?
+  end
+end
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -3,6 +3,7 @@ html
   head
     title Portus
     = stylesheet_link_tag 'application', media: 'all'
+    = stylesheet_link_tag "debug" if peek_enabled?
     = csrf_meta_tags
     meta content="width=device-width, initial-scale=1" name="viewport"
     link href="/favicon/apple-touch-icon-57x57.png" rel="apple-touch-icon" sizes="57x57"
@@ -32,9 +33,11 @@ html
       window.API_ROOT_URL = '#{root_url}';
     script src="//cdn.polyfill.io/v2/polyfill.js?features=Array.prototype.some,Array.prototype.findIndex,Array.from"
     = javascript_include_tag(*webpack_asset_paths("application"))
+    = javascript_include_tag("debug") if peek_enabled?
     = yield :js_header
 
   body data-route="#{js_route}"
+    = render_peek
     .vue-root
       header
         = render 'shared/header'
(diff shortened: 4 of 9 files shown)


mssola:
I've tried this and I have some problems 😕 

First of all, the whole `Gem::Specification.find_all_by_name("peek").first&.activated?` doesn't work for me. It always returns false, even if I have it installed. Am I missing anything ? I think you can safely remove all this, and check only if we are in development mode.

Moreover, if I manually remove all this, all I get is a mere brown bar:

lala

So it looks like the sql and the performance bar does not work for me 😕 Also, besides Mysql note that we also support PostgreSQL now 😉

Finally, I think there are other peek plugins which might be interesting: host and gc. I think you should include them too 😃

salzig:
Didn't want to limit it to development mode, cause maybe someone is interested in running it in test/production. Using `define?` now, easier to handle :)

Gemfile
```diff
@@ -85,6 +85,14 @@ unless packaging?
     gem "guard", require: false
     gem "guard-rspec", require: false
     gem "guard-rubocop", require: false
+    gem "jquery"
```
mssola: Please remove this. We are using jQuery from the webpack configuration.

mssola: One thing that could be done is to add the peek css and js in webpack just for development.

app/assets/javascripts/debug.js
```diff
@@ -0,0 +1,4 @@
+//= require jquery
+//= require jquery_ujs
```
mssola: As said before, remove these lines.

app/views/layouts/application.html.slim
```diff
@@ -3,6 +3,7 @@ html
   head
     title Portus
     = stylesheet_link_tag 'application', media: 'all'
+    = stylesheet_link_tag "debug" if peek_enabled?
```
mssola: This should be handled in webpack instead.

app/views/layouts/application.html.slim
```diff
@@ -32,9 +33,11 @@ html
       window.API_ROOT_URL = '#{root_url}';
     script src="//cdn.polyfill.io/v2/polyfill.js?features=Array.prototype.some,Array.prototype.findIndex,Array.from"
     = javascript_include_tag(*webpack_asset_paths("application"))
+    = javascript_include_tag("debug") if peek_enabled?
```
mssola: Same

app/helpers/peek_helper.rb
```diff
@@ -0,0 +1,13 @@
+module PeekHelper
+  def peek_available?
+    defined? Peek
```
mssola: When would be Peek not be available ? Just asking because I'm still not sure about this 😕

config/initializers/assets.rb
```diff
@@ -1,2 +1,3 @@
 Rails.application.config.assets.version = "1.0"
 Rails.application.config.assets.precompile += %w[*.woff2]
+Rails.application.config.assets.precompile += %w[debug.js debug.css] if defined? Peek
```
mssola: Again, let webpack handle this

Gemfile
```diff
@@ -85,6 +85,14 @@ unless packaging?
...
+    gem "jquery-rails"
+    gem "peek"
+    gem "peek-gc"
+    gem "peek-git"
+    gem "peek-mysql2"
+    gem "peek-performance_bar"
+    gem "peek-pg"
```
mssola: Could you add `peek-host` ? It's kind of important considering JWT tokens will contain the hostname, and it wouldn't be the first time we have a problem because we forgot to set it properly 😄

mssola:
Thanks for this PR 👍 My main observations from it are:

- We are using webpack for the assets, so try to use that instead of the regular "Rails way" of keeping assets.
- I still don't see the point of `peek_available?`. If I run `bundle`, then it will be defined, so that will return true (and I need to run bundle otherwise rails will raise an exception). So it will always be available. I guess that you want developers to configure whether they want to use peek or not. For that, maybe you can use an environment variable that you have to set to true in order to use peek ?

Gemfile
vitoravelino: > One thing that could be done is to add the peek css and js in webpack just for development.

This might be tricky or not even possible because peek's assets relies on rails [0] [1]. For [0] that wouldn't be a problem because it's still a valid comment in js and we can include its dependency. But for [1] I think that might cause an error since it's not valid css. 

If that's the case, I'm fine with having the dedicated assets files to be built just for peek. I would only rename it to `peek` instead of `debug`.

[0] https://github.com/peek/peek/blob/master/app/assets/javascripts/peek.js#L1
[1] https://github.com/peek/peek/blob/master/app/assets/stylesheets/peek.scss#L1

vitoravelino: Just realized that it's a `scss` file, so the comment will still be valid but we would need to configure webpack to load `scss` files properly.

mssola:
@salzig any news on this ?

mssola:
Closing, but feel free to submit a new PR once you addressed the issues we pointed out. Thanks a lot :+1:

PR #1493, opened 2017-10-30
Registry Client should probe the /v2/ path and accept more return codes

### LINKED ISSUE #1486: Registry reachability tests are failing
On latest master the registry reachability test is always indicating that the registry is not reachable even though login/push/pull is working just fine.  Of what does the Reachable test consists?

The tooltip error is: something went wrong check your configuration

Logs:

```
Registry not reachable: #<NoMethodError: undefined method `[]' for nil:NilClass>
 Write fragment views/registry1_status/6400318743a2031dc52df318c76c5c13 (4.8ms)
 Rendered admin/registries/index.html.slim within layouts/application (80.0ms)
```

Vad1mo:
Registry Client should probe that the /v2/ path reachable and that we accept 200 responses as well.

fixes #1486

Going straight to the /v2/ allows us to know we are targeting the right Registry.

We could also check for the header

```yml
  docker-distribution-api-version:
  - registry/2.0
```

to make sure we are really talking to the registry.

makes also https://github.com/SUSE/Portus/tree/master/examples/compose usable again.

### DIFF
--- a/lib/portus/registry_client.rb
+++ b/lib/portus/registry_client.rb
@@ -32,13 +32,16 @@ module Portus
     # Returns whether the registry is reachable with the given credentials or
     # not. This might raise a RequestError on failure.
     def reachable?
-      res = safe_request("", "get", false)
+      res = safe_request("/v2/", "get", false)
 
-      # If a 401 was retrieved, it means that at least the registry has been
-      # contacted. In order to get a 200, this registry should be created and
+      # The 'Docker-Distribution-Api-Version' header indicates that we are connected to a
+      # Docker Registry endpoint.
+      # 401 means that the registry requires authentication
+      # In order to get a 200, this registry should be created and
       # an authorization requested. The former can be inconvenient, because we
       # might want to test whether the registry is reachable.
-      !res.nil? && res.code.to_i == 401
+      !res.nil? && res.header.key?("Docker-Distribution-Api-Version") &&
+        (res.code.to_i == 401 || res.code.to_i == 200)
     end
 
     # Calls the `/:repository/manifests/:tag` endpoint from the registry. It
--- a/spec/lib/portus/registry_client_spec.rb
+++ b/spec/lib/portus/registry_client_spec.rb
@@ -23,7 +23,7 @@ class RegistryPerformRequest < Portus::RegistryClient
     # We don't care about the given parameters.
 
     return nil if @status.nil?
-    OpenStruct.new(code: @status)
+    OpenStruct.new(code: @status, header: { "Docker-Distribution-Api-Version" => "registry/2.0" })
   end
 end
 
@@ -175,7 +175,7 @@ describe Portus::RegistryClient do
 
   context "is reachable or not" do
     it "returns the proper thing in all the scenarios" do
-      [[nil, false], [200, false], [401, true]].each do |cs|
+      [[nil, false], [404, false], [200, true], [401, true]].each do |cs|
         r = RegistryPerformRequest.new(cs.first)
         expect(r.reachable?).to be cs.last
       end
--- a/spec/models/registry_spec.rb
+++ b/spec/models/registry_spec.rb
@@ -113,7 +113,8 @@ describe Registry, type: :model do
     end
 
     it "returns the proper message for each scenario" do
-      GOOD_RESPONSE = OpenStruct.new(code: 401)
+      GOOD_RESPONSE = OpenStruct.new(code: 401, header:
+        { "Docker-Distribution-Api-Version" => "registry/2.0" })
 
       [
         [nil, GOOD_RESPONSE, true, /^$/],
--- a/spec/vcr_cassettes/health/clair-bad.yml
+++ b/spec/vcr_cassettes/health/clair-bad.yml
@@ -42,9 +42,11 @@ http_interactions:
       - d0865133-bb26-4eac-ac26-370f66bc457c
       X-Runtime:
       - '0.006993'
+      Docker-Distribution-Api-Version:
+      - 'registry/2.0'
     body:
       encoding: UTF-8
       string: '{"error":"You need to sign in or sign up before continuing."}'
-    http_version: 
+    http_version:
   recorded_at: Thu, 03 Aug 2017 15:50:16 GMT
 recorded_with: VCR 3.0.3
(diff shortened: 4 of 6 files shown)


lib/portus/registry_client.rb
```diff
@@ -26,13 +26,13 @@ def initialize(host, use_ssl = false, username = nil, password = nil)
...
     def reachable?
-      res = perform_request("", "get", false)
+      res = perform_request("/v2/", "get", false)
 
       # If a 401 was retrieved, it means that at least the registry has been
       # contacted. In order to get a 200, this registry should be created and
       # an authorization requested. The former can be inconvenient, because we
       # might want to test whether the registry is reachable.
-      !res.nil? && res.code.to_i == 401
+      !res.nil? && res.code.to_i == 401 || res.code.to_i == 200
```
vitoravelino: `(res.code.to_i == 401 || res.code.to_i == 200)`, right?

Vad1mo: it should work, however its more clear as you suggested.

http://ruby-doc.org/core-2.4.2/doc/syntax/precedence_rdoc.html

lib/portus/registry_client.rb
```diff
@@ -26,13 +26,13 @@ def initialize(host, use_ssl = false, username = nil, password = nil)
     # Returns whether the registry is reachable with the given credentials or
     # not.
     def reachable?
-      res = perform_request("", "get", false)
+      res = perform_request("/v2/", "get", false)
```
mssola: Could you also leave a comment like this ?

```ruby
# NOTE: be aware of this when implementing the v2 version of the REST API
# as pointed out here: https://github.com/SUSE/Portus/issues/1500.
```

Vad1mo:
alright, will do. When is the release planned? Is there going to be a beta phase?

Vad1mo:
The code checks that the header `'Docker-Distribution-Api-Version'` exists and not its actual value. 
That check should just verify we are connected to the docker registry and not with ordinary http server.

mssola:
> When is the release planned? Is there going to be a beta phase?

A first release is expected for November 8th. This will be a beta release, so we still have time 😉 How many beta releases we deliver depends on the amount of bug fixes we want to include, so there's no hard pressure on that either.

> The code checks that the header 'Docker-Distribution-Api-Version' exists and not its actual value.
That check should just verify we are connected to the docker registry and not with ordinary http server.

👍

Vad1mo:
I am trying to fix the test cases however there is one im not sure how its related to the registry_client. 

looking at the VCRs `Portus/spec/vcr_cassettes/health/ok.yml` I see that the request is to the registry but the response is from nginx.  

```
Failures:

  1) HealthController GET /_health Basic functionality has both the DB and the registry
     Failure/Error: expect(response.status).to eq 200
     
       expected: 200
            got: 503
     
       (compared using ==)
     # ./spec/controllers/health_controller_spec.rb:32:in `block (5 levels) in <top (required)>'
     # /Users/vadim/.rvm/gems/ruby-2.4.2/gems/vcr-3.0.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
     # /Users/vadim/.rvm/gems/ruby-2.4.2/gems/vcr-3.0.3/lib/vcr.rb:189:in `use_cassette'
     # ./spec/controllers/health_controller_spec.rb:30:in `block (4 levels) in <top (required)>'

  2) HealthController GET /_health Clair enabled works even if the server contains another port
     Failure/Error: expect(response.status).to eq 200
     
       expected: 200
            got: 503
     
       (compared using ==)
     # ./spec/controllers/health_controller_spec.rb:82:in `block (5 levels) in <top (required)>'
     # /Users/vadim/.rvm/gems/ruby-2.4.2/gems/vcr-3.0.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
     # /Users/vadim/.rvm/gems/ruby-2.4.2/gems/vcr-3.0.3/lib/vcr.rb:189:in `use_cassette'
     # ./spec/controllers/health_controller_spec.rb:80:in `block (4 levels) in <top (required)>'

  3) HealthController GET /_health Clair enabled handles a proper 200 response
     Failure/Error: expect(response.status).to eq 200
     
       expected: 200
            got: 503
     
       (compared using ==)
     # ./spec/controllers/health_controller_spec.rb:73:in `block (5 levels) in <top (required)>'
     # /Users/vadim/.rvm/gems/ruby-2.4.2/gems/vcr-3.0.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
     # /Users/vadim/.rvm/gems/ruby-2.4.2/gems/vcr-3.0.3/lib/vcr.rb:189:in `use_cassette'
     # ./spec/controllers/health_controller_spec.rb:71:in `block (4 levels) in <top (required)>'

```

mssola:
@Vad1mo I just checked your branch, and did some inspection. So, the trouble part is the check of the `docker-distribution-api-version` header. The health controller calls the `health` library, which will check for the reachability of the registry. The `ok.yml` VCR cassette is based on my development environment, which is a bare-metal deployment with NGinx as a proxy. Apparently my configuration blocks the `docker-distribution-api-version` header, and so the check fails. That's why you are getting a 503 on all those places.

Now, I don't think that my deployment is to blame here 😁, I don't think administrators should have to go through a list of headers that cannot be blocked, and stuff like that. Instead, I think that the proper solution is to drop the header check: we don't want to be too strict, we are just testing that the registry is reachable, that's all. If you do this, the tests will be green.

Vad1mo:
Actually, the nginx config is incorrect, or better said incomplete.  In a proper nginx reverse proxy setup the header should be forwarded. IMHO [nginx-proxy](https://github.com/jwilder/nginx-proxy/blob/master/nginx.tmpl) has a pretty decent config for reverse proxying with nginx. 

However, you are right, we don't need to test for the existence of the headers as we are testing the connectivity only and not proper working of the service. I will revert the changes so only /v2/ is tested.

mssola:
@Vad1mo could you rebase your changes ? 🤓

Vad1mo:
I hate to say that, but the tests work fine locally.

mssola:
Ok, I'll give it a spin whenever I have some time (there are some pressing issues on my queue, once I'm done with that I'll take a look 😉)

Vad1mo:
rebased on latest master, tested and fixed test problems

@mssola could you merge.

mssola:
Thanks a lot ! 👍

PR #1504, opened 2017-11-06
js: prevent app token link to be clickable when disabled

vitoravelino:
This was a missing change done while fixing the profile form.


mssola:
Thanks 👏

Issue #1500, opened 2017-11-02
Consolidate the REST API


mssola:
Someone from the community started the REST API (see #1299), which is quite a big deal. This was also important because we were starting to transition to Vue, which required some work on an API level. Because of this, we started to add more and more endpoints organically. This is good because it means that we are building an API that makes sense (at least for us), but it also means that the totality of it hasn't been designed from the very beginning. This can lead to some inconsistencies, and some areas for improvement.

This issue is a reminder that for a 3.0 version of Portus we need to:

1. Evaluate what went well and what went wrong during the process of creating the v1 version of the REST API.
2. Check whether we can provide fixes on the API without breaking anyone's workflow (e.g. creating a v1.1 version and deprecating old endpoints).
- After point 2, we should evaluate when to roll out a v2 of the API that removes the oddities of v1.

Issue #1437, opened 2017-09-27
Normalize API error responses


vitoravelino:
Currently we have the following formats:

```js
// grape validation (400)
{"errors" => {"[\"name\"]" => ["is missing"]}}

// team creation (422)
{"errors" => ["Name can't be blank"]}

// authorization (403)
{"errors" => "Authorization fails"}
```

We need to make the values more consistent.

PR #1505, opened 2017-11-07
team users: migrated to api and vue

vitoravelino:
Migrated team users away from the legacy controller and js code to
API and Vue components.


mssola:
Sounds great, thanks for this 👏 Just ping me when this is ready to be reviewed 👍

vitoravelino:
> Sounds great, thanks for this 👏 Just ping me when this is ready to be reviewed 👍

Will do. 😃

vitoravelino:
@mssola This is good to be reviewed.

We are starting to repeat a bit ourselves in the js side but I want to refactor that in another PR since this one is already large.

vitoravelino:
Forgot to mention, this one won't be part of the next release, right? 

I think we should rebase `v2.3` with current `master`. For new stuff that really needs to go into 2.3 we can create a backport. This way we can keep working on `master` without worrying about breaking stuff for 2.3. What do you think?

mssola:
> I think we should rebase v2.3 with current master. For new stuff that really needs to go into 2.3 we can create a backport. This way we can keep working on master without worrying about breaking stuff for 2.3. What do you think?

Totally agree. I'd like to fix the packaging side first though.

This PR has conflicts 😉

lib/api/entities.rb
```diff
@@ -151,6 +151,35 @@ class Teams < Grape::Entity
       }, if: { type: :internal } { |t| t.namespaces.count }
     end
 
+    class TeamMembers < Grape::Entity
```
mssola: I wonder if we can either subclass `Users` or somehow share as much as possible from the `Users` entity ...

lib/api/v1/teams.rb
```diff
@@ -84,6 +84,96 @@ class Teams < Grape::API
...
+            end
+
+            delete ":member_id" do
+              team_user = TeamUser.find_by(id: params[:member_id], team_id: params[:id])
+              raise ActiveRecord::RecordNotFound if team_user.nil?
+              authorize team_user, :destroy?
+
+              destroyed = ::TeamUsers::DestroyService.new(current_user).execute(team_user)
+
+              if destroyed
```
mssola: IIRC, simply doing the following has a similar effect:

```ruby
unless destroyed
  error!({ "errors" => team_user.errors.full_messages }, 422, header)
end
```

As in, in the positive case, then null will be returned, which is fine.

vitoravelino: I thought that I had tried that and got an exception but I double checked now and it seems to work. Anyway, thanks! :)

lib/api/entities.rb
vitoravelino: If we present the `team_user.user` instead of `team_user` and add the extra fields, I guess we could inherit from `Users` entity. I don't have a strong opinion on this one to be honest.

lib/api/v1/teams.rb
vitoravelino: I was right, for some reason this fails in the tests but not in the browser itself when I try to remove a member.

```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN">
<HTML>
  <HEAD><TITLE>Internal Server Error</TITLE></HEAD>
  <BODY>
    <H1>Internal Server Error</H1>
    no implicit conversion of nil into String
    <HR>
    <ADDRESS>
     WEBrick/1.3.1 (Ruby/2.4.2/2017-09-14) at
     127.0.0.1:57911
    </ADDRESS>
  </BODY>
</HTML>
```

No idea why.

mssola: Ah right, it returns a JSON `null` object, that you should then handle on the JS side. I think this also happens on other endpoints. This is something to consolidate either in #1437 or #1500.

lib/api/entities.rb
mssola: Maybe it's fine for now, just wondering because they share some fields. This is something we can discuss in #1500.

lib/api/v1/teams.rb
vitoravelino: But it's an internal server error when it should be a success with a no content response. 
There's nothing to handle on the client-side. 

This only happens when running the tests. If I go to the browser and try to delete a member, it works as expected.

vitoravelino:
> Maybe it's fine for now, just wondering because they share some fields. This is something we can discuss in #1500.

Makes sense. 👍

lib/api/entities.rb
vitoravelino: Makes sense. 👍

lib/api/v1/teams.rb
vitoravelino: @mssola any idea on this? ☝️

mssola:
This PR fixes part of #1755.

app/services/team_users/build_service.rb
```diff
@@ -0,0 +1,39 @@
...
+    def build_team_user
+      team = fetch_team
+      user = fetch_user
+      TeamUser.new(user: user, team: team, role: params[:role])
+    end
+
+    def fetch_user
+      user = User.find_by(username: params[:user])
+      params.delete(:user)
+      raise ActiveRecord::RecordNotFound if user.nil?
```
mssola: You don't have to do this if you call `find_by!` instead of `find_by` before :wink:

app/services/team_users/build_service.rb
```diff
@@ -0,0 +1,39 @@
...
+      user
+    end
+
+    def fetch_team
+      team = Team.find_by(id: params[:id], hidden: false)
+      raise ActiveRecord::RecordNotFound if team.nil?
```
mssola: Same

app/services/team_users/build_service.rb
vitoravelino: Done. I've also fixed this in other places regarding this PR.

PR #1518, opened 2017-11-20
js: added alert plugin

vitoravelino:
Same idea of #1517. In this case you are still able to use the `Alert` itself but not recommended.


mssola:
Merging #1517 means that this now has conflicts. Please fix them, but I like the idea 👍

vitoravelino:
Rebase.

PR #1520, opened 2017-11-21
js: fixed team name validation behaviour

vitoravelino:
Previously the inline validation was being done only over the scoped
teams current user had access too without their hidden team.

This way when the user tried to create a team with their own hidden name
or another user's name, an error happened.

With this patch the error appears inline as we expect from validation.

I also changed the polyfills syntax import.

### DIFF
--- a/app/views/teams/components/_form.html.slim
+++ b/app/views/teams/components/_form.html.slim
@@ -8,7 +8,7 @@ div ref="form" class="collapse"
           span v-if="!$v.team.name.required"
             | Name can't be blank
           span v-if="!$v.team.name.available"
-            | Name has already been taken
+            | Name is reserved or has already been taken
     .form-group
       = f.label :description, {class: 'control-label col-md-2'}
       .col-md-7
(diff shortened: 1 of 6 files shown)


vitoravelino:
This one I think we should include in 2.3.

app/views/teams/components/_form.html.slim
```diff
@@ -8,7 +8,7 @@ div ref="form" class="collapse"
           span v-if="!$v.team.name.required"
             | Name can't be blank
           span v-if="!$v.team.name.available"
-            | Name has already been taken
+            | Name reserved or has already been taken
```
mssola: I'd change it to: "Name is reserved or has already been taken"

mssola:
That being said, listing all teams is not such a security risk imho, so maybe we can remove the scoped scenario altogether ... What do you think ?

vitoravelino:
> That being said, listing all teams is not such a security risk imho, so maybe we can remove the scoped scenario altogether ... What do you think ?

The problem is not leaking but listing as an option when this same endpoint is used for creating a namespace. User might get confused with an option that he doesn't know where it came from and that is not valid when submitted. 

I know that it's kinda ugly but I wasn't able to come up with something better at the time.

vitoravelino:
@mssola Let's not forget to include this in 2.3, ok? Just reminding you.

PR #1536, opened 2017-12-12
js: namespace vue pages components cleanup

vitoravelino:
Each namespace page (index, show) now has its own component. With this
we can pass any value from controller as props and not access them as
global variables. This also makes the page components easier to test.

One thing I could have done was to move the panel inside of the show page component to its own component but I'll wait we introduce Vuex to centralize and manage the state of components. This will avoid a lot of component props pollution.


spec/features/auth/login_feature_spec.rb
```diff
@@ -86,13 +86,14 @@ module CleanRoom
...
 
-  scenario "Successful login when trying to access a page redirects back the guest" do
+  scenario "Successful login when trying to access a page redirects back the guest", js: true do
     visit namespaces_path
     expect(page).to have_content("You need to sign in or sign up before continuing.")
     fill_in "user_username", with: user.username
     fill_in "user_password", with: user.password
     find("button.classbutton").click
     expect(page).to have_current_path(namespaces_path)
+    save_screenshot(nil, full: true)
```
mssola: Remove this line 😉

app/helpers/namespaces_helper.rb
```diff
@@ -5,7 +5,7 @@ def can_manage_namespace?(namespace)
   end
 
   def can_create_namespace?
-    current_user.admin? || APP_CONFIG.enabled?("user_permission.create_namespace")
+    Registry.any? && (current_user.admin? || APP_CONFIG.enabled?("user_permission.create_namespace"))
```
mssola: Can you add a test case for this change ?

mssola:
Moreover, you have conflicts and Travis is failing

app/helpers/namespaces_helper.rb
vitoravelino: To be honest this is not needed. I guess in the past we weren't forcing the user to fill the registry form and they could visit other pages. I'll just remove it.

vitoravelino:
@mssola it's green!

mssola:
@vitoravelino why did you remove the `Registry.any?` call ? 😁

vitoravelino:
> @vitoravelino why did you remove the Registry.any? call ? 😁

Because it's always gonna be true. We won't be showing that page without forcing the user to add a registry. If they try to visit any page (namespace in this case), they will always be redirected to `/admin/registries/new`. Then `Registry.any?` will always be true. I double checked before removing it.

PR #1551, opened 2017-12-21
Make PORTUS_ROOT configurable via environment variable

mahiso:


mssola:
This is fine by me, but this version of `portusctl` is going to be removed next version, in favor of [opensuse/portusctl](https://github.com/openSUSE/portusctl) (and in there you can configure the root path). I'll merge it, but we aware of this situation 😉

PR #1563, opened 2018-01-04
Split routes.rb into several modules

lightlace:
In order to clean up the mess in the routes.rb file, we split routes.rb into several modules which ar later loaded via "draw" in the routes.rb file.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,81 +1,32 @@
 # frozen_string_literal: true
 
 # rubocop:disable Metrics/BlockLength
-# TODO: see https://github.com/SUSE/Portus/issues/1469
 Rails.application.routes.draw do
-  resources :errors, only: [:show]
-  resources :teams, only: %i[index show update] do
-    member do
-      get "typeahead/:query" => "teams#typeahead", :defaults => { format: "json" }
-    end
-  end
-  get "/teams/typeahead/:query" => "teams#all_with_query", :defaults => { format: "json" }
 
+	def draw(routes_module)
+    instance_eval(File.read(Rails.root.join("config/routes/#{routes_module}.rb")))
+	end
+  resources :errors, only: [:show]
+	draw :teams
   resources :help, only: [:index]
 
   resources :team_users, only: %i[create destroy update]
-  resources :namespaces, only: %i[index show update] do
-    put "change_visibility", on: :member
-    resources :webhooks do
-      resources :headers, only: %i[create destroy], controller: :webhook_headers
-      resources :deliveries, only: [:update], controller: :webhook_deliveries
-      member do
-        put "toggle_enabled"
-      end
-    end
-  end
-  get "namespaces/typeahead/:query" => "namespaces#typeahead", :defaults => { format: "json" }
-
-  resources :repositories, only: %i[index show destroy] do
-    post :toggle_star, on: :member
-    resources :comments, only: %i[create destroy]
-  end
-
+	draw :namespaces
+	draw :repositories
   resources :tags, only: %i[show destroy]
-
   resources :application_tokens, only: %i[create destroy]
-
-  devise_for :users, controllers: { registrations:      "auth/registrations",
-                                    sessions:           "auth/sessions",
-                                    passwords:          "passwords",
-                                    omniauth_callbacks: "auth/omniauth_callbacks" }
+	draw :user
   resource :dashboard, only: [:index]
   resources :search, only: [:index]
   resources :explore, only: [:index]
-
-  authenticated :user do
-    root "dashboard#index", as: :authenticated_root
-  end
-
-  devise_scope :user do
-    root to: "auth/sessions#new"
-    put "toggle_enabled/:id", to: "auth/registrations#toggle_enabled", as: :toggle_enabled
-  end
-
-  namespace :v2, module: "api/v2", defaults: { format: :json } do
-    root to: "ping#ping", as: :ping
-    resource :token, only: [:show]
-    resource :webhooks, only: [] do
-      resources :events, only: [:create]
-    end
-  end
-
+	draw :v2
   mount API::RootAPI => "/"
   mount GrapeSwaggerRails::Engine, at: "/documentation" unless Rails.env.production?
 
   get "users/oauth", to: "auth/omniauth_registrations#new"
   post "users/oauth", to: "auth/omniauth_registrations#create"
 
-  namespace :admin do
-    resources :activities, only: [:index]
-    resources :dashboard, only: [:index]
-    resources :registries, except: %i[show destroy]
-    resources :namespaces, only: [:index]
-    resources :teams, only: [:index]
-    resources :users do
-      put "toggle_admin", on: :member
-    end
-  end
+	draw :admin
 
   # Error pages.
   %w[401 404 422 500].each do |code|
--- a/config/routes/user.rb
+++ b/config/routes/user.rb
@@ -1,0 +1,13 @@
+devise_for :users, controllers: { registrations:      "auth/registrations",
+                                  sessions:           "auth/sessions",
+                                  passwords:          "passwords",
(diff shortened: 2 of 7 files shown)


lightlace:
@mssola

vitoravelino:
Remember to fix the rubocop offenses.

config/routes.rb
```diff
@@ -1,81 +1,32 @@
...
-  get "/teams/typeahead/:query" => "teams#all_with_query", :defaults => { format: "json" }
 
+	def draw(routes_module)
+    instance_eval(File.read(Rails.root.join("config/routes/#{routes_module}.rb")))
+	end
+  resources :errors, only: [:show]
+	draw :teams
   resources :help, only: [:index]
 
   resources :team_users, only: %i[create destroy update]
```
vitoravelino: You can add this to the `teams` file.

config/routes.rb
```diff
@@ -1,81 +1,32 @@
...
-      resources :events, only: [:create]
-    end
-  end
-
+	draw :v2
   mount API::RootAPI => "/"
   mount GrapeSwaggerRails::Engine, at: "/documentation" unless Rails.env.production?
 
   get "users/oauth", to: "auth/omniauth_registrations#new"
   post "users/oauth", to: "auth/omniauth_registrations#create"
```
vitoravelino: Move the 2 lives above to `user`.

config/routes/user.rb
```diff
@@ -0,0 +1,13 @@
+devise_for :users, controllers: { registrations:      "auth/registrations",
```
vitoravelino: Rename file to `users` instead of `user`?

config/routes.rb
```diff
@@ -1,81 +1,32 @@
...
-  get "namespaces/typeahead/:query" => "namespaces#typeahead", :defaults => { format: "json" }
-
-  resources :repositories, only: %i[index show destroy] do
-    post :toggle_star, on: :member
-    resources :comments, only: %i[create destroy]
-  end
-
+	draw :namespaces
+	draw :repositories
   resources :tags, only: %i[show destroy]
```
vitoravelino: I think this could go inside `repositories` but I'm not sure...

config/routes.rb
```diff
@@ -1,81 +1,32 @@
...
-
-  namespace :v2, module: "api/v2", defaults: { format: :json } do
-    root to: "ping#ping", as: :ping
-    resource :token, only: [:show]
-    resource :webhooks, only: [] do
-      resources :events, only: [:create]
-    end
-  end
-
+	draw :v2
```
vitoravelino: I'm not sure if `v2` is the best name here. I think `legacy_api` would fit better. @mssola can give a better opinion about this.

config/routes.rb
lightlace: i will test it. :)

config/routes.rb
```diff
@@ -1,81 +1,32 @@
...
 Rails.application.routes.draw do
-  resources :errors, only: [:show]
-  resources :teams, only: %i[index show update] do
-    member do
-      get "typeahead/:query" => "teams#typeahead", :defaults => { format: "json" }
-    end
-  end
-  get "/teams/typeahead/:query" => "teams#all_with_query", :defaults => { format: "json" }
 
+	def draw(routes_module)
```
mssola: Could you hide this inside of another file and inside of a module or class ? And, since this comes directly from Gitlab's source code, could you also write a comment stating where did this come from ? 😉

config/routes.rb
mssola: Yeah. I'd say that it should go into is own file, but since it's so simple, let's put it inside of the `repositories` file 👍

config/routes.rb
mssola: Rename this to `registries` or something like that, since `v2` can be troublesome...

config/routes.rb
```diff
@@ -1,81 +1,32 @@
...
-  resources :repositories, only: %i[index show destroy] do
-    post :toggle_star, on: :member
-    resources :comments, only: %i[create destroy]
-  end
-
+	draw :namespaces
+	draw :repositories
   resources :tags, only: %i[show destroy]
-
   resources :application_tokens, only: %i[create destroy]
```
mssola: This could go inside of `users` since it's closely related

config/routes.rb
```diff
@@ -1,81 +1,32 @@
...
-
-  devise_for :users, controllers: { registrations:      "auth/registrations",
-                                    sessions:           "auth/sessions",
-                                    passwords:          "passwords",
-                                    omniauth_callbacks: "auth/omniauth_callbacks" }
+	draw :user
   resource :dashboard, only: [:index]
   resources :search, only: [:index]
   resources :explore, only: [:index]
```
mssola: Let's group this three into a `root` file. It could also contain the `help` resource (it should be harmless to change the order of this one)

mssola:
oops, Github didn't take into account the previous review from @vitoravelino when I added him as a reviewer 😕

config/routes.rb
lightlace: done.

config/routes.rb
lightlace: done.

config/routes/user.rb
lightlace: done.

vitoravelino:
> oops, Github didn't take into account the previous review from @vitoravelino when I added him as a reviewer 😕

Wat?! Super weird... We found a bug at least. xD

config/routes.rb
lightlace: done.

vitoravelino:
> %i[teams namespaces ...].each { |f| draw f }
 
+1 on this (can't +1 the review comment itself).

config/routes.rb
lightlace: done.

config/routes.rb
lightlace: done.

config/routes.rb
vitoravelino: Is `registries` the best option we can think of? We have some admin registries thing and might have more in the future, I guess renaming it to `registries` would make things more confusing.

mssola: Yeah ... What about `events` ? Since it's all about events getting in and out.

vitoravelino: `ping`, `token`, `webhooks/events`... 🤔  

Btw, do we still need `ping`? We now have a `health` endpoint with the grape api. Without `ping` there are 2 left... They belong to `api/v2` and are used as api endpoints, I still think `legacy_api` could be a good call but I dunno.

mssola: Bear in mind that those are endpoints dealing with registry events, and ping is an action that can be performed by the registry to Portus iirc 😉 The `token` endpoint is the one responding to JWT requests from the registry.

We could move these endpoints to the portus API, now that we have one, but then you'd force people to change the config of the registry from v2 to v1 (v2 of what we had because we were stating v2 of the registry, and then v1 for the portus API). One way to alleviate this is to move these endpoints when we create the v2 of the portus API. Then this move would be transparent to administrators. What do you think ?

vitoravelino: > Bear in mind that those are endpoints dealing with registry events, and ping is an action that can be performed by the registry to Portus iirc 😉 The token endpoint is the one responding to JWT requests from the registry.

Okie.

>  We could move these endpoints to the portus API, now that we have one, but then you'd force people to change the config of the registry from v2 to v1 (v2 of what we had because we were stating v2 of the registry, and then v1 for the portus API). One way to alleviate this is to move these endpoints when we create the v2 of the portus API. Then this move would be transparent to administrators. What do you think ?

Agreed! 👍 

Maybe `registry_events`, `registry_api`, ...?

mssola: Sounds good to me 👍

mssola: I just added this point in #1500.

PR #1592, opened 2018-01-15
js: introduced unit testing

vitoravelino:
We are introducing the ability of writing javascript unit tests. Until now,
this wasn't possible.

To achieve that we decided to go with the mocha + webpack approach that
uses the `mocha-webpack` package. As the assertion library we choose
`expect`.

To run the tests you can simply type `yarn test`. And if you want the to check
the coverage, you can type `yarn coverage`.

There still stuff to be improved. The `.vue` files coverage does not work
properly and this is something that needs investigation.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,9 @@
+{
+  "presets": [
+    "es2015"
+  ],
+  "plugins": [
+    "lodash",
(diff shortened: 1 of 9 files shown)


vitoravelino:
@mssola good to be reviewed! :)

PR #1603, opened 2018-01-19
feature: local login form can be disabled

figroc:
disable login form to enable OAuth only login page

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -89,6 +89,11 @@ ldap:
 
 # OAuth support.
 oauth:
+  # If enabled, then users can login with portus credential.
+  # Otherwise, the users can only login via OAuth.
+  local_login:
+    enabled: true
+
   # If enabled, users can authenticate with their Google Account.
   # Callback url: <host>/users/auth/google_oauth2/callback
   google_oauth2:
--- a/spec/features/auth/login_feature_spec.rb
+++ b/spec/features/auth/login_feature_spec.rb
@@ -109,6 +109,36 @@ describe "Login feature" do
     expect(page).to have_current_path(new_user_session_path)
   end
 
+  it "Login form is enabled when local_login is enabled" do
+    APP_CONFIG["oauth"]["local_login"] = { "enabled" => true }
+    APP_CONFIG["ldap"] = { "enabled" => false }
+
+    visit root_path
+    expect(page).to have_current_path(root_path)
+    expect(page).to have_content("Login")
+    expect(page).to have_content("I forgot my password")
+  end
+
+  it "Login form is enabled when ldap is enabled" do
+    APP_CONFIG["oauth"]["local_login"] = { "enabled" => false }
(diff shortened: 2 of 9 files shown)


vitoravelino:
Hi!

Some [tests](https://travis-ci.org/SUSE/Portus/jobs/330702935#L1342) are failing. You forgot to enable the config for those since it's related to the user login. You also need to add this config option to the portusctl `cli.rb` to fix the `options_spec` test.

I think this idea is fine. Once you fix those issues, we can merge it. Thanks!

figroc:
Sorry for the noisy commits. I wasn't able to build the project at my local machine.

vitoravelino:
No problem with the amount of commits. Before the merge we always ask people to squash their commits. :)

spec/features/auth/login_feature_spec.rb
```diff
@@ -109,6 +109,36 @@ module CleanRoom
     expect(page).to have_current_path(new_user_session_path)
   end
 
+  it "Login form is enabled when local_login is enabled" do
+    APP_CONFIG["local_login"] = { "enabled" => true }
```
vitoravelino: The default is enabled, so you don't need to do this here.

figroc: I prefer to keep this very line. The config should be explicitly specified for this test case, for whatever default value is.

config/config.yml
```diff
@@ -162,6 +162,14 @@ oauth:
 first_user_admin:
   enabled: true
 
+# If enabled, then users can login with portus credential. Otherwise, the users
+# can only login via OAuth
+#
+# This is ignored if LDAP is enabled. Read more about this here:
+# http://port.us.org/features/disabling_local_login.html
+local_login:
```
mssola: Maybe it would make sense to place this inside of the `oauth` section ?

.rubocop.yml
```diff
@@ -43,7 +43,7 @@ Metrics/BlockLength:
 #   2. The models regarding the registry have to be modularized.
 #   3. LDAP has to be refactored.
 Metrics/ClassLength:
-  Max: 210
+  Max: 214
```
mssola: Why ? 😕

figroc: I add four lines to `spec_helper.rb`.
rubocop complained.

mssola: Ah, I see. Then simply disable this cop for the whole cli.rb instead of raising this value. This is because this `portusctl` is deprecated in favor of [openSUSE/portusctl](https://github.com/openSUSE/portusctl) in 2.3, and since we have already created the 2.3 branch, we will nuke this code in master rather soon 😉

mssola:
@figroc could you squash your commits ?

figroc:
Done. Thank you for the reviews @vitoravelino @mssola .
Looking forward for the merge and release ;)

mssola:
Thanks @figroc for this! 👍

PR #1631, opened 2018-01-25
Add the possibility to disable background tasks

lightlace:
Add possibility to disable background tasks


lib/portus/background/registry.rb
```diff
@@ -14,9 +14,13 @@ def sleep_value
       # options and because the `execute!` method will deal with valid rows
       # already.
       def work?
-        true
```
mssola: You should've left this as it was

lib/portus/background/registry.rb
```diff
@@ -14,9 +14,13 @@ def sleep_value
...
+			  true
       end
 
+			def enabled?
+			  return APP_CONFIG("background.registry")
```
mssola: Don't use `return`, ruby will implicitly return the last eval'd object. Moreover, it should be like this:

```
APP_CONFIG.enabled?("background.registry")
```

config/config.yml
```diff
@@ -280,3 +280,7 @@ security:
 # Docker Registry. Only images on public namespaces will be shown.
 anonymous_browsing:
   enabled: true
+
+background:
+  registry: true
+  sync: true
```
mssola: Prefer the convention:

```yaml
background:
  registry:
    enabled: true
  sync: true
    enabled: true
```

mssola: Plus, add documentation.

lib/portus/background/security_scanning.rb
```diff
@@ -17,6 +17,10 @@ def work?
         ::Portus::Security.enabled? && Tag.exists?(scanned: Tag.statuses[:scan_none])
       end
 
+			def enabled?
+			  return ::Portus::Security.enabled?
```
mssola: Remove the `return`.

lib/portus/background/sync.rb
```diff
@@ -18,6 +18,10 @@ def work?
         true
       end
 
+      def enabled?
+			  return APP_CONFIG("background.sync")
```
mssola: Same as with the `registry`

lib/portus/background/registry.rb
```diff
@@ -17,6 +17,14 @@ def work?
         true
       end
 
+      def enabled?
+        if APP_CONFIG("background.registry") == true
+          Rails.logger.warn("WARNING: Registry is disabled!")
+        else
+          APP_CONFIG.enabled?("background.registry")
+        end
+      end
```
mssola: I don't think `APP_CONFIG("background.registry")` will work. Do this instead:

```ruby
def enabled?
  val = APP_CONFIG.enabled?("background.registry")
  Rails.logger.warn("WARNING: Registry is disabled!") unless val
  val
end
```

After that, notice the following:

1. You don't have to start the message with `WARNING`, since the `warn` method will already cover that.
2. The message might not be clear to users. Instead, prefer something like: "Registry integration has been disabled. This is highly discouraged!". For the `sync`: "Registry synchronization has been disabled. This can lead into consistency problems!"

mssola:
This has moved into #1679.

Issue #1619, opened 2018-01-24
OpenID Connect not actually implemented


mcharris:
According to the documentation Portus should be able to support OpenID Connect when it is not supported by the current implementation. The gem that was added is omniauth-openid which only supports OpenID and not OpenID Connect. Please note, these are two different things.

Vad1mo:
Hmm, thank you for the info. Indeed it should be OIDC.

We will provide an PR.

mssola:
@Vad1mo do you think it's an easy thing to provide ? Or we should postpone this to 2.4 ?

mcharris:
@Vad1mo I am currently working on a pull request for adding OpenID Connect. The issue is that there isnt a rubgems.com package that is updated/maintained for OIDC with omniauth. The closest one is gem https://github.com/hhorikawa/omniauth-openid-connect. 

Essentially I would need to add
`gem 'omniauth-openid-connect', :git => 'https://github.com/hhorikawa/omniauth-openid-connect'` to the Gemfile. That just doesnt seem like a good solution. For my needs I've been contemplating adding omniauth-saml instead since it will fix my problem and it is maintained by omniauth.io.

Vad1mo:
alright, did you look at https://github.com/m0n9oose/omniauth_openid_connect its more or less up to date. 

Go forward with the PR, I will have some time next week for testing.

mcharris:
@Vad1mo I did but decided against it since it wasnt updated for that last 2 years. I will use that though unless I run into an issue.

Vad1mo:
Hello @mcharris, sorry it seems I messed up with cope and paste. Looking again at the openid connect situation, it seems like indeed https://github.com/hhorikawa/omniauth-openid-connect
is the best and most up to date available option. 

Sorry for the confussion

mcharris:
Pull request #1633 open. Let me know if you have any questions. I might close the pull request if it fails travis tests.

mssola:
@Vad1mo & @mcharris if you don't mind we can leave this for 2.4, since the PR seems quite big (that is, you can submit PRs as usual, but I won't cherry pick it to 2.3)

Vad1mo:
Fine for me, unless 2.4 is not in 2019 :D

mssola:
@mcharris have you worked on this anymore ?

trunet:
can you use https://github.com/m0n9oose/omniauth_openid_connect instead?

Vad1mo:
[...]

PR #1633, opened 2018-01-25
OpenID Connect & Active Session Storage

mcharris:
Integrated with `https://github.com/hhorikawa/omniauth-openid-connect` to provide a simple OpenID Connect authorization. There was a few other bugs that I ran into.

when storing the Omniauth respose into the session there was mixed code. One was treating the object as a hash and another was treating it as a omniauth object. There was only one reference to the omniauth object so I just converted the object to an hash and then changed the one reference to the omniauth object to be a hash reference.

With the pull of the omniauth-openid-connect being from github the docker file used to create the development image needs git to also be installed.

Added social link for OpenID Connect. The verbiage for hte link was too long so I made it say ODIC on the social icon. I do think we need to revist how this is generated. With the addition of OpenID Connect and OpenID there maybe a need to specify several social icons with the same type. Both the social icons and the config will need to change to accomodate this. I opted to not include this change because it could break current installations when upgrading.

Session storage - I ran into a limitation where the information stored inside of the session was exceeding 4kb. This is because currently portus uses cookie based session storage which has a limitation of 4kb. unfortunatly openid requries 4.1kb for the tokens and certs. I switched to the other common session storage to be `active_record_store` which is a database driven session storage. This really should of been done sense the first implementation of omniauth as there is some critical information in that data that could be abused if stored in a cookie. I am also surprised that the 4kb limit hasnt been broken for other things in Portus.

Added simple spec for openid_connect. Since i am not sure how to run the tests locally I think it will fail when travis-ci runs it. 

Issues #1619

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ COPY Gemfile* ./
 #      installed with the devel_basis pattern, and finally we zypper clean -a.
 RUN zypper ref && \
     zypper -n in --no-recommends ruby2.5-devel \
-           libxml2-devel nodejs libmysqlclient-devel postgresql-devel libxslt1 && \
+           libxml2-devel nodejs libmysqlclient-devel postgresql-devel libxslt1 git && \
     zypper -n in --no-recommends -t pattern devel_basis && \
     gem install bundler --no-ri --no-rdoc -v 1.16.0 && \
     update-alternatives --install /usr/bin/bundle bundle /usr/bin/bundle.ruby2.5 3 && \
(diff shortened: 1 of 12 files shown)


mcharris:
Closing to fix issues. from CI

Vad1mo:
we should ask @hhorikawa if he would like to publish his gem. Getting it from git is not the best solution imho.

trunet:
did you see https://github.com/m0n9oose/omniauth_openid_connect ?

PR #1679, opened 2018-02-12
add the possibility to disable background tasks

lightlace:
New version of #1631


mssola:
Merging directly since further work is being done in #1675

PR #1682, opened 2018-02-14
Change the default robots.txt so that Portus cannot be crawled by search engines

lightlace:


public/robots.txt
```diff
@@ -1,5 +1,11 @@
 # See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
 #
 # To ban all spiders from the entire site uncomment the next two lines:
-# User-agent: *
-# Disallow: /
+User-agent: *
+Disallow: /
```
mssola: You shouldn't be doing this I fear. It's disallowing *all* paths (so the following lines would be irrelevant then 😉)

mssola: You shouldn't be doing this I fear. It's disallowing *all* paths (so the following lines would be irrelevant then 😉)

public/robots.txt
```diff
@@ -1,5 +1,11 @@
...
+
+User-agent: *
+Disallow: /namespaces
+Disallow: /repositories
+Disallow: /teams
+Disallow: /admin/
```
mssola: The `/admin/` has an extra slash. It should be just `/admin` I think.

Moreover, this is missing:

- `/errors`
- `/search`
- `/help`
- `/team_users`
- `/v2`
- `/tags`
- `/users`
- `/toggle_enabled`
- `/application_tokens`
- `/documentation`
- `/api`

vitoravelino:
I guess the only page that would be interesting to be indexed is the explore one. The rest are internal or not relevant. Unfortunately `Allow` is only applicable to Googlebot, so we can't just allow it and ban everything else.

If we agree that the other bots (except Googlebot) are irrelevant, we could simply

```
Allow: /explore
Disallow: /
```

vitoravelino:
> If we agree that the other bots (except Googlebot) are not important enough, we could simply

All bots accept `Allow`, so my bad mentioning that. Then we can simplify using the `Allow` solution.

mssola:
> All bots accept Allow, so my bad mentioning that. Then we can simplify using the Allow solution.

That's really cool! Then we can simply list `Allow`, which is much simpler.

mssola:
Two things about `Disallow: /`:

- Are you sure this includes subpaths ? I guess so, but just to be sure.
- We don't want to disallow / itself (the main page).

vitoravelino:
> Are you sure this includes subpaths ? I guess so, but just to be sure.

Yes.

> We don't want to disallow / itself (the main page). 

But there's no main page, it redirects to the sign in page. We can allow /users/sign_in if that's the case.

lightlace:
> But there's no main page, it redirects to the sign in page. We can allow /users/sign_in if that's the case.

It redirects only to the sign in page, if you are logged out. Otherwise is redirects to the dashboard.

PR #1696, opened 2018-02-21
feature: contributors allowed to delete repositories/tags

### LINKED ISSUE #1273: Delete Image Button only available for owners of a user group
It is currently possible to configure Portus such that a recycling bin button appears to the right of each tag of an Image.
However, this button can only be configured for the "Owner" of a team. It cannot be made visible for other team members e.g. for those acting in the role "Contributor".
It would be nice to be able to either configure that on a per-user Basis or at least on a per-role basis.

vitoravelino:
### Summary

With the variety usage of Portus, the need of giving contributors more responsibilities arose. So far only admin were able to delete repositories and tags if the capability were enabled. This patch extends the feature to also allow contributors to be able delete resources.

We also took the opportunity to migrate and add two endpoints [0] [1] the API regarding this topic. Users are now able to invoke the API to perform those actions if they want to.

[0] `DELETE /api/v1/repositories/:id`
[1] `DELETE /api/v1/tags/:id`

- [X] Implementation
- [x] New API endpoints
- [x] Usage of API endpoints in the client-side
- [x] Testing

Fixes #1273


vitoravelino:
The solution is already in place but I'm going to take the opportunity to migrate these endpoints to the API and use it in the client-side.

vitoravelino:
@mssola This is ready to be reviewed. I'm gonna open an issue for the documentation part of this. I'll work on that whenever I finish my other tasks.

app/helpers/repositories_helper.rb
```diff
@@ -32,9 +32,8 @@ def render_delete_activity(activity)
   end
 
   # Returns true if the user can remove the given repository.
-  def can_destroy?(repository)
-    APP_CONFIG.enabled?("delete") &&
-      RepositoryPolicy.new(current_user, repository).destroy?
+  def can_destroy_repository?(repository)
```
mssola: Aren't you missing the `APP_CONFIG.enabled?("delete")` part ?

app/models/tag.rb
```diff
@@ -72,7 +72,11 @@ def delete_by_digest!(actor)
       return false
     end
 
-    Tag.where(digest: dig).map { |t| t.delete_by!(actor) }
+    tags_destroyed = true
+    Tag.where(digest: dig).find_each do |tag|
+      tags_destroyed &&= tag&.delete_by!(actor)
+    end
+    tags_destroyed
```
mssola: In `tags_destroyed &&= tag&.delete_by!(actor)` you may have removed tags from a previous iteration, even if this one fails. Thus, you should rename `tags_destroyed` to `success`.

app/policies/repository_policy.rb
```diff
@@ -20,7 +20,12 @@ def show?
   # Returns true if the repository can be destroyed.
   def destroy?
     raise Pundit::NotAuthorizedError, "must be logged in" unless @user
-    @user.admin? || @repository.namespace.team.owners.exists?(user.id)
+
+    delete_enabled         = APP_CONFIG.enabled?("delete")
+    is_owner               = @repository.namespace.team.owners.exists?(user.id)
+    can_contributor_delete = APP_CONFIG["delete"]["contributors"] &&
```
mssola: Change to: `APP_CONFIG.enabled?("delete.contributors")` as you are doing in the tag policy. Moreover, maybe it would be a good change to DRY this part and create a method like this:

```ruby
def allow_destroy_for?(user, repository)
  delete_enabled         = APP_CONFIG.enabled?("delete")
  is_owner               = repository.namespace.team.owners.exists?(user.id)
  can_contributor_delete = APP_CONFIG.enabled?("delete.contributors") &&
                             @repository.namespace.team.contributors.exists?(user.id)
  delete_enabled && (user.admin? || is_owner || can_contributor_delete)
end
```

Note that the tag policy basically does `tag.repository`, so the repository is the only object needed.

lib/api/v1/repositories.rb
```diff
@@ -105,6 +105,24 @@ class Repositories < Grape::API
...
+               ]
+
+          delete do
+            repository = Repository.find(params[:id])
+            authorize repository, :destroy?
+
+            destroy_service = ::Repositories::DestroyService.new(current_user)
+            destroyed = destroy_service.execute(repository)
+
+            error!({ "errors" => destroy_service.error }, 422, header) unless destroyed
```
mssola: On success a delete request should return a 204 status code.

lib/api/v1/tags.rb
```diff
@@ -45,6 +45,24 @@ class Tags < Grape::API
...
+               ]
+
+          delete do
+            tag = Tag.find(params[:id])
+            authorize tag, :destroy?
+
+            service = ::Tags::DestroyService.new(current_user)
+            destroyed = service.execute(tag)
+
+            error!({ "errors" => service.error }, 422, header) unless destroyed
```
mssola: Same as in the repository.

lib/api/v1/repositories.rb
vitoravelino: Remember that in the other PR about managing team member (still pending btw, but I need to re-review and rebase before asking another review) I was having an issue by writing just the unless? For some reason this is working now. Maybe gem version has been updated? Response seems fine as you can see below:

```
Request URL:http://localhost:3000/api/v1/repositories/12
Request Method:DELETE
Status Code:204 No Content
```

lib/api/v1/tags.rb
vitoravelino: Same here. :smile:

app/policies/repository_policy.rb
vitoravelino: > Change to: APP_CONFIG.enabled?("delete.contributors") as you are doing in the tag policy.

The tag policy was actually a leftover since it doesn't work. The `enabled?` function works only with the `enabled` child attribute. In this case we have `contributors` being a child of `delete` and the value goes directly to it and not to a `enabled` child. I should actually fix the tag policy file.

I double checked and by using `enabled?` on `delete.contributors`, it always return `true`. I think that if `enabled` child attribute does not exist it should return false, no?

vitoravelino:
>  The Travis failure is most probably that you haven't signed off your commit (-s flag) 

I'm just signing the commits via GPG (-S) but not showing the signature (-s). I'll configure my `gitconfig` to explicitly add the signature.

vitoravelino:
GREEN! :smile:

lib/api/v1/repositories.rb
mssola: Cool. As long as the tests works I'm fine then :+1:

mssola:
Thanks a lot! :+1:

PR #1699, opened 2018-02-22
Make reply_to setting optional

### LINKED ISSUE #1693: reply_to should be an optional setting
If `reply_to? is not set in the config, this error pops up during start:

```
`check_email!': Mail: bad config value for 'reply_to'. '' is not a proper email... (RuntimeError)
```

lightlace:
Fixes #1693 .

Make sure that the email settings are not checked and therefore no error is raised, when "reply_to" is empty.

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -7,7 +7,9 @@
 email:
   from: "portus@example.com"
   name: "Portus"
-  reply_to: "no-reply@example.com"
+  # The reply_to field is empty, as the user has the possibility to enable
+  # it optionally by filling in the appropriate reply address.
+  reply_to: ""
 
   # If enabled, then SMTP will be used. Otherwise 'sendmail' will be used
   # (defaults to: /usr/sbin/sendmail -i -t).
--- a/config/initializers/mail.rb
+++ b/config/initializers/mail.rb
@@ -8,7 +8,7 @@ end
 
 unless Rails.env.test?
   check_email!("from")
-  check_email!("reply_to")
+  check_email!("reply_to") if APP_CONFIG["email"]["reply_to"].present?
 
   # If SMTP was set, then use it as the delivery method and configure it with the
   # given config.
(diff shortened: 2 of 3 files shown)


config/config.yml
```diff
@@ -7,7 +7,7 @@
 email:
   from: "portus@example.com"
   name: "Portus"
-  reply_to: "no-reply@example.com"
+  reply_to: "no-reply@mycompany.com"
```
vitoravelino: I think you should leave this empty since it's optional. Leaving a comment explaining why this is useful (or not) would be nice too.

config/initializers/mail.rb
```diff
@@ -8,7 +8,7 @@ def check_email!(key)
 
 unless Rails.env.test?
   check_email!("from")
-  check_email!("reply_to")
+  check_email!("reply_to") unless APP_CONFIG["email"]["reply_to"] == ""
```
mssola: This is safer:

```ruby
if APP_CONFIG["email"]["reply_to"].present?
```

mssola: Same in `app/mailers/devise_mailer.rb`.

config/config.yml
```diff
@@ -3,11 +3,13 @@
 # (it will be ignored by git). For more info, you can read the dedicated page
 # here: http://port.us.org/docs/Configuring-Portus.html.
 
-# Settings for the Portus mailer.
+# Settings for the Portus mailer.a
+# Thenreply_to field is empty, as the user has the possibility to enable
+# it optionally by filling in the appropiate reply address.
```
mssola: Move this comment to the option itself. as it's done for the `smtp` option for example.

config/config.yml
```diff
@@ -3,11 +3,13 @@
 # (it will be ignored by git). For more info, you can read the dedicated page
 # here: http://port.us.org/docs/Configuring-Portus.html.
 
-# Settings for the Portus mailer.
+# Settings for the Portus mailer.a
```
mssola: Typo. Probably a leftover.

mssola:
Thanks :+1:

PR #1656, opened 2018-02-05
Rename swagger_doc into official naming openapi-spec.

Vad1mo:
Rename swagger_doc into official naming openapi-spec.

closes #1652

However in feature releases we remove the swagger term in code at other location as well.

PR #1704, opened 2018-02-26
The OpenAPI should appear under openapi-spec and not swagger_doc

Vad1mo:
This is a fix for https://github.com/SUSE/Portus/pull/1656 The api wasn't still at the right place but the documentation was already updates. Now both the documentation and the the api are available under `/api/openapi-spec`


Vad1mo:
@mssola ping

mssola:
Thanks :clap:

PR #1689, opened 2018-02-19
Implement administrator-only image push setting

lightlace:
Implement a environment variable in config.yml, so that only admins can push images.

PR #1705, opened 2018-02-26
Implement a environment variable in config.yml, so that only admins can push images.

lightlace:
Implement a environment variable in config.yml, so that only admins can push images.

New version of #1689.

Signed-off-by: Fabian Baumanis <fabian.baumanis@suse.com>


mssola:
Thanks :clap:

Issue #1766, opened 2018-04-05
Reorganize admin pages/tasks


vitoravelino:
### Description

Currently admin pages are kinda confusing and not too useful. For example, an admin cannot create a team when they reach the teams admin page. The same happens for namespaces. The repositories page currently list all the repositories you have access instead of only your own (like we do for namespaces and teams).

Based on those issues, we decided to unify the admin tasks into the normal pages. So we will get rid of some admin pages and improve the rest 

- [x] Get rid of admin teams page and unify with normal teams page (via #1776)
- [X] Get rid of admin namespaces page and unify with normal namespaces page (via #1783)
- [x] Split repositories into "teams member repositories" and "other repositories" (via #1785)
- [x] Migrate admin dashboard overview elements (number of teams, namespaces, etc) to normal overview (via #1786)
- [x] Group recent activities into "personal" and "administrative" tabs or something similar (via #1786)
- [x] Show "users" sidebar item if admin (via #1784)
- [x] Show "registries" sidebar item if admin (via #1784)

PR #1776, opened 2018-04-10
teams: improved team creation form with owner

### LINKED ISSUE #1765: Ability to define an owner when creating a team
### Description

Currently this is not possible. Whenever you create a team, who ever is logged becomes the team owner. This is not always the case and it requires boring and unnecessary additional work to add another user as owner and remove yourself from the team. 

The idea is to have the logged user as the pre-selected owner but make it possible to define another user as the owner.

**This will only be available for admin users.**

vitoravelino:
If you are a Portus admin, you might want to create various teams with
its respective owners. So far the path to achieve that was tedious and
not friendly. Admin had to create a team, add a member as owner and
remove himself as ownere later on.

With this patch, admin will be able to not only create a team through
the admin team page but also assign an owner different of himself.

Fixes #1765


mssola:
Looks good to me. One thing though: I'd also add this field into the main teams page (only appearing if the current user is an admin, and the default would be the same current user).

vitoravelino:
> One thing though: I'd also add this field into the main teams page (only appearing if the current user is an admin, and the default would be the same current user).

I thought about that but the problem is that the teams page lists all the teams that the user is member of. So, if we allow an admin to add a team from that page and defining a different owner, the added team wouldn't appear in the list. It would be confusing. 

An alternative though (that I really like btw), is to get rid of this admin overview page and unify the admin and normal user pages. If user is an admin, we are more flexible with the policies resolve method. Be aware that some pages might change a little bit due to that but I think it's safe. We also need to worry about the impact on the API with this change.

I started doing what I suggested on #1766 but I'm more than happy to go with the unification approach. What do you think?

vitoravelino:
Based on our internal discussion, we are going with the unification approach. Going to rework on this PR.

PR #1783, opened 2018-04-16
namespaces: unified admin page with regular page

vitoravelino:
We got rid of the admin namespaces page and unified the vision of the
admin on the regular namespace page. So, if the user is an admin, they
will be seeing all of the namespaces and will be able to manage all of
them. If the user is a regular one, everything stays the same.

API is also affected by this change.

One improvement we did was to separate the other namespaces the user had
access to without being through membership in a different panel. This
will help the user understand better how they have access to those
namespaces.


lib/api/v1/namespaces.rb
```diff
@@ -11,11 +11,12 @@ class Namespaces < Grape::API
...
 
-        helpers ::API::Helpers::Namespaces
-
         desc "Returns a list of namespaces",
              tags:     ["namespaces"],
-             detail:   "This will expose all accessible namespaces",
+             detail:   "This will expose all accessible namespaces by the user via either team
+                        membership or visibility. Keep in mind that if the user is an admin, this
+                        will return all the global, personal and other namespaces created by all
+                        the users.",
```
mssola: Nitpick: remove the "." at the end. When exporting to the documentation we expect all strings to not have a period at the end.

vitoravelino:
Good to go.

PR #1786, opened 2018-04-16
dashboard: unified admin dashboard page with regular page

vitoravelino:
We got rid of the admin dashboard page. We finaly complete the unification
of admin tasks/vision on the regular pages. If the user is an admin,
they will see more things on the page now.
    
In this case, the dashboard page will show similar components to what the
user used to see on the admin dashboard page. The main difference is that
now we have tabs grouping the types of recent activities.
    
If the user is a regular one, nothing changes and everything stays the same.

### Regular user
screenshot-20180416143406-1915x431

### Admin
screenshot-20180416143430-1915x535


vitoravelino:
> I liked the dedicated admin area, as everything there was dedicated to the administration.
Audit trails of all users. The widgets.

The widgets and audit trails (recent activities for admin) were not removed. We just unified with the regular dashboard page as you can see on the screenshot.

> You shouldn't forget that admins are users too :)
> This means that admins in small setups have two roles user & admin. Having them separated reflacts > that use case. IMHO Portus is mostly used in this scenarios

I know and that's why we are trying to improve and create new semantics in the UI. We are trying to be explicit on why the user has access to something and not only listing everything in a single table as it was before. This can always be improved and we'll try to keep working on it. 

The reason we are doing this is because the admin pages were not good enough and kinda forgotten this whole time. We also heard complains about them. That's why we decided to do this change.

Nothing prevents us from reverting this in the future if we see that it makes sense to go that way again. But for now we thought that this was a good idea. Hopefully we won't regret this decision. :sweat_smile: 

Thanks for being worried about this!

Vad1mo:
The point is we have some custom widget that display relevant informations or point to further settings. 

In the new setup will the user see the widgets or only the admin?

vitoravelino:
> In the new setup will the user see the widgets or only the admin?

Only the admin. I thought I had written that in the commit message but I didn't. My bad. Everything stays the same for the regular user. The changes are only for admins.

vitoravelino:
I'm going to reword the commit later. It's pretty bad. 

@mssola you can review this and the other PRs but lemme reword all the commit messages before merging everything, ok?

mssola:
> @mssola you can review this and the other PRs but lemme reword all the commit messages before merging everything, ok?

ugh, I've already merged a couple of them ...

vitoravelino:
Good to go.

PR #1815, opened 2018-05-09
config: added pagination options

### LINKED ISSUE #1375: Add global pagination limit config value
Table page limit is currently hard coded, it would be nice to have this configurable in the global configuration file. It would also be useful for future back-end optimizations too.

vitoravelino:
So far the pagination and number of shown pages before and after current
page were hard coded and not configurable.

With this the user can now choose from a configuration perspective how
he wants to show their tables.

Fixes #1375

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -340,3 +340,12 @@ background:
     #     empty. This is the default value since it's deemed to be the most
     #     common use-case.
     strategy: initial
+
+# Pagination configuration
+pagination:
+  # Number of entries to be listed per page
+  limit: 10
+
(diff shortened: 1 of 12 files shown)


config/config.yml
```diff
@@ -340,3 +340,12 @@ background:
     #     empty. This is the default value since it's deemed to be the most
     #     common use-case.
     strategy: initial
+
+# Pagination confgiuration
```
mssola: "confgiuration" -> "configuration"

vitoravelino: Fixed.

PR #1816, opened 2018-05-09
docs: pagination config

vitoravelino:


vitoravelino:
@mssola let's keep this on hold until we release 2.4.

mssola:
We are already adding documentation that will only make it for 2.4, so it's fine to merge it. We only need to warn users about the availability of this feature (and you are already doing it on this PR).

There's a Travis CI error though :sweat_smile:

PR #1822, opened 2018-05-15
js: migrated webhooks show page to vue

vitoravelino:
This patch does not include webhook headers and deliveries. Only the
details panel and the edit form.


app/assets/javascripts/utils/http.js
```diff
@@ -1,6 +1,7 @@
 import Alert from '~/shared/components/alert';
 
 export function handleHttpResponseError(response) {
+  console.log(response);
```
mssola: You may want to remove this :wink:

PR #1824, opened 2018-05-16
js: migrated webhook headers to vue components

vitoravelino:
Signed-off-by: Vítor Avelino <vavelino@suse.com>

PR #1825, opened 2018-05-16
js: migrated webhook deliveries to vue components

vitoravelino:
This patch also includes a small fix that was preventing the request of
the webhook to happen when no authentication were provided.


vitoravelino:
Only merge this after #1824.

PR #1826, opened 2018-05-22
ui: added visibility to namespace edit form

### LINKED ISSUE #1442: ux: change the namespace visibility from the show page

vitoravelino:
The visibility change was only possible via namespaces table showed in
the namepace#index page. User can now change the visibility while
editing a namespace.

A <visibility-chooser> component was created to abstract most of the
visibilities options. <namespace-visibility> evolved to be a wrapper of
the previous component that can be used in standalone to change a
namespace remotely.

Fixes #1442 

screenshot-20180522104823-387x430


vitoravelino:
I need to provide that info in the details panel but I'll do it in another PR.

PR #1848, opened 2018-06-06
js: refactor data -> components way on repositories#index

vitoravelino:


app/assets/javascripts/modules/repositories/pages/index.vue
```diff
@@ -0,0 +1,42 @@
...
+    data() {
+      return {
+        teamRepositories: [...this.teamRepositoriesRef],
+        otherRepositories: [...this.otherRepositoriesRef],
+      };
+    },
+  };
+</script>
+
+<style>
```
mssola: If there's no style, is this really needed ?

vitoravelino: Nope. Forgot to delete from generated snippet. Thanks!

PR #1866, opened 2018-06-25
spec: added chrome headless as default js runner

vitoravelino:
So far the feature tests with js enabled were running through
poltergeist driver. As we know, PhantomJS was discontinued and
that's bad. PhantomJS was far from perfect and we've had a lot
of flaky issues with it.

Even knowing that most of those flaky issues were our fault, it's
always good to improve our workflow whenever possible. Based on this we
are adding chrome headless as our default js runner for our feature
tests.

Unfortunately we had to keep poltergeist for one scenario that we
couldn't make chrome headless work properly. That scenario is when the
dev runs the feature tests inside of the container. For bare metal and
travis, chrome headless works like a charm.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@ RUN zypper addrepo https://download.opensuse.org/repositories/devel:languages:go
     zypper --gpg-auto-import-keys ref && \
     zypper -n in --no-recommends ruby2.5-devel \
            libmysqlclient-devel postgresql-devel \
-           nodejs libxml2-devel libxslt1 git-core go1.10 && \
+           nodejs libxml2-devel libxslt1 git-core \
+           go1.10 phantomjs && \
     zypper -n in --no-recommends -t pattern devel_basis && \
     gem install bundler --no-ri --no-rdoc -v 1.16.0 && \
     update-alternatives --install /usr/bin/bundle bundle /usr/bin/bundle.ruby2.5 3 && \
--- a/Gemfile
+++ b/Gemfile
@@ -100,10 +100,13 @@ end
 
 group :test do
   gem "capybara", "~> 2.14.3"
-  gem "codeclimate-test-reporter", group: :test, require: nil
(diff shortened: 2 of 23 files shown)


Dockerfile
```diff
@@ -33,4 +34,11 @@ RUN zypper addrepo https://download.opensuse.org/repositories/devel:languages:go
            libtool m4 make makeinfo && \
     zypper clean -a
 
+# Add Google Chrome repo and install it
+RUN zypper  -n ar https://dl.google.com/linux/chrome/rpm/stable/x86_64 Chrome && \
```
mssola: Merge with the above `RUN` block.

Gemfile
```diff
@@ -100,10 +100,13 @@ end
 
 group :test do
   gem "capybara", "~> 2.14.3"
-  gem "codeclimate-test-reporter", group: :test, require: nil
+  gem "capybara-screenshot", "~> 1.0.0"
+  gem "chromedriver-helper"
+  gem "codeclimate-test-reporter", require: false
```
mssola: Why did you remove the `test` group here ?

spec/support/capybara.rb
```diff
@@ -1,11 +1,76 @@
...
+      console = page.driver.browser.manage.logs.get(:browser)&.reject do |log|
+        log.message =~ JS_CONSOLE_FILTER
+      end
+
+      p console.map(&:message).join("\n") if console.present?
+    end
+  end
+end
+
+# HACK: when running tests inside of a container we should use poltergeist
```
mssola: Maybe this comment should go on the line where you set `JAVASCRIPT_DRIVER`.

Gemfile
vitoravelino: Because it's already inside of the test group.

vitoravelino:
> In the commit message you are talking about PhantomJS being discontinued, wasn't it Poltergeist? If not, then can you remove the lines where we download phantomjs in Travis?

:+1:

PR #1879, opened 2018-07-10
Improve detection of containerized environments - fix issue #1732

### LINKED ISSUE #1732: Failed to create bus connection: No such file or directory - when calling /api/v1/health endpoint
### Description

I've just deployed opensuse/portus:2.3.0 (with puma) connected to a postgres:9.6 pod into kubernetes.
Everything works fine except of Portus is logging "Failed to create bus connection: No such file or directory" after every Kubernetes LiveProbe calling /api/v1/health check. The Database is accessible, the software works like expected, calling the URL in the browser gives me success all over the place (database, clair, registry) - but it always logs this error. the /api/v1/_ping endpoint works like expected (no Error in Log).

### Steps to reproduce

1. Watching the POD Output of Portus
2. calling the /api/v1/health endpoint

- **Expected behavior**: No Error Message in Log
- **Actual behavior**: Error Message in Log, but everything is working

flavio:
The code determining if Portus is running containerized is broken on Kubernetes because in this case PID 1 doesn't belong to the `docker` cgroup slice.

That causes the code to invoke `hostnamectl`, that tries to access kdbus from within a container. That causes Rails to fail immediately.

fixes #1732

### DIFF
--- a/packaging/suse/portusctl/lib/constants.rb
+++ b/packaging/suse/portusctl/lib/constants.rb
@@ -4,7 +4,8 @@
 
 # Checks whether it's running inside of a Docker container or not
 def dockerized?
-  File.read("/proc/1/cgroup").include?("docker")
+  cgroup = File.read("/proc/1/cgroup")
+  cgroup.include?("docker") || cgroup.include?("kubepod")
 end
 
 # This one is set by the bash wrapper we deliver with our RPM


mssola:
Thanks a lot :clap:

Monnoroch:
@flavio also, docker is not the only container runtime. There's also containerd, rkt and probably more. Checking for kubepods is great for kubernetes, but this check has to be redesigned to handle more runtimes.

flavio:
@Monnoroch by default kubernetes places all the user PODs under this cgroup namespace, despice of the CRI in use.

Monnoroch:
There is life outside kubernetes too. Besides, what if they change the cgroup name? Current implementation looks like a hack, really. I think there should be a way for a user to force assuming a containerized environment. Or, better yet, making the software environment-agnostic.

flavio:
As far as I've seen this code is gone from the master branch of Portus. I see this just as a temporary fix.

@mssola can you confirm what I've said, we are not going to have to deal with this code anymore. Am I right?

mssola:
> As far as I've seen this code is gone from the master branch of Portus. I see this just as a temporary fix.

That's correct. This fix only applied to the `v2.3` branch. For the `master` branch we are using [this portusctl](https://github.com/openSUSE/portusctl), which does not need this kind of check.

PR #1880, opened 2018-07-12
Make containerized deployment more configurable

flavio:
A bunch of changes required to be more flexible when deploying Portus into a containerized environment.

### TODO

* [ ] Extend the documentation to mention the new environment variable added.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -31,6 +31,9 @@ default: &default
   <% if ENV["PORTUS_DB_TIMEOUT"] %>
   timeout: <%= ENV['PORTUS_DB_TIMEOUT'] %>
   <% end %>
+  <% if ENV['PORTUS_DB_SOCKET'] %>
+  socket:  <%= ENV['PORTUS_DB_SOCKET'] %>
+  <% end %>
 
 <%= Rails.env.downcase %>:
   <<: *default
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -16,7 +16,7 @@ workers ENV.fetch("PORTUS_PUMA_WORKERS") { 2 }.to_i
 #     purposes inside of a container.
 #  3. UNIX socket. This is the default and it's good for development purposes if
 #     you are not using a container setup.
-if ENV["PORTUS_PUMA_HOST"]
+if ENV["PORTUS_PUMA_HOST"] && ENV["PORTUS_PUMA_USE_UNIX_SOCKET"] != "true"
   if ENV["PORTUS_PUMA_TLS_KEY"]
     host, port = ENV["PORTUS_PUMA_HOST"].split(":")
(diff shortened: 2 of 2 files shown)


flavio:
I think this should be backported also to the 2.3 branch

docker/init
```diff
@@ -82,7 +82,6 @@ done
 update-ca-certificates
 
 # Further settings
-export PORTUS_PUMA_HOST="0.0.0.0:3000"
```
mssola: I wouldn't change this, since it will break the deployment of most users. I propose to export this env. variable to this default value if none has been provided.

flavio: This isn't going to help with my use-case: have puma bind to unix socket even when running Portus using the official container, see:

https://github.com/SUSE/Portus/blob/7358f3a81be062218cdacb209702fce089af2f23/config/puma.rb#L19-L31

Another way would be to rework the code snipped I pasted above to something like:

```ruby
if ENV["PORTUS_PUMA_HOST"] && !ENV["PORTUS_PUMA_USE_UNIX_SOCKET"]
...
else
# bind to socket
end
```

mssola: I'd prefer the code you just pasted instead of the old one because I believe that telling people to always set this env. variable to "0.0.0.0:3000" for the most common use case is confusing. Instead, if you really know what you are doing, you can provide this extra env. variable.

So a :+1: from my side.

mssola: Oh, and instead of `!ENV["PORTUS_PUMA_USE_UNIX_SOCKET"]`, I think it's safer `ENV["PORTUS_PUMA_USE_UNIX_SOCKET"].blank?`

flavio:
@mssola I've updated the code but I haven't tested it. AFAIK `.blank?` is defined by active support, I fear this is not available at that time. Can you double check?

flavio:
Also, which branch holds the documentation?

mssola:
> I've updated the code but I haven't tested it. AFAIK .blank? is defined by active support, I fear this is not available at that time. Can you double check?

I've tested it and nothing broke :smile: That's because we are calling puma from within the context of a Rails application (notice, for example, that we later call `ActiveRecord` directly :wink:). So, it should be fine.

> Also, which branch holds the documentation?

It's on the `site-source` branch, because of an issue we had with jekyll and custom plugins.

mssola:
As for the failing integration tests, maybe a rebase does the trick, otherwise we'll have to check what's broken.

flavio:
Rebased

mssola:
@flavio tomorrow we want to release `2.3.4`, which includes an important fix. Could you update this PR so we can also include this there?

flavio:
@mssola done!

mssola:
Cherry-picked and pushed into `v2.3`.

PR #1883, opened 2018-07-17
js: upgraded js dev packages

vitoravelino:
The main packages that caused bigger changes to the code were webpack and
eslint. That happened due to the distance from the upstream version.

Coverage now seems more reliable with the replacement of the
instrumentation package.

Also removed unnecessary packages.


vitoravelino:
This is the first part of the packages upgrade. Next PR will take care of the app packages itself (e.g.: vue, lodash, etc).

PR #1891, opened 2018-07-19
js: reduced bundle size

vitoravelino:
With the migration to webpack 4, the latest uglify plugin stopped
removing comments by default and some unnecessary files were being
included.

- Moment.js was replaced with dayjs
- Webpack bundle analyzed was added to be used when necessary to analyze
bundle content

Unfortunately bundle is still big but that's because vue-multiselect was
introduced and lots of templates were migrated to .vue files.

### DIFF
--- a/app/assets/javascripts/utils/date.js
+++ b/app/assets/javascripts/utils/date.js
@@ -1,7 +1,21 @@
-import moment from 'moment';
+import dayjs from 'dayjs';
 
-const isValid = date => moment(date, moment.ISO_8601).isValid();
+const isISO8601 = (date) => {
+  const type = typeof date;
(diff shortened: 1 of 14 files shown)


vitoravelino:
Next step regarding this topic is to split the bundle into two chunks: vendors and application.

vitoravelino:
This failing was killing me inside and I had to go with an approach I didn't like but I'd say it's good enough for now. If we get in trouble in the future we can get back to this.

momentjs is awesome but unfortunately a bazooka for what we've been doing.

app/assets/javascripts/utils/date.js
```diff
@@ -1,7 +1,21 @@
-import moment from 'moment';
+import dayjs from 'dayjs';
 
-const isValid = date => moment(date, moment.ISO_8601).isValid();
+const isISO8601 = (date) => {
```
mssola: This is quite dodgy, but I guess you copied it directly from `moment`. It's fine by now, I guess :sweat_smile:

PR #1903, opened 2018-07-26
Fix gitlab

kiorky:
### Summary

This allow
* to use a private gitlab instance
* to use paginated group results (actually only the first 20 groups are searched, other are ignored.)

### DIFF
--- a/app/controllers/auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/auth/omniauth_callbacks_controller.rb
@@ -24,6 +24,22 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   # Callback for Bitbucket.
   alias bitbucket google_oauth2
 
+  protected
+
+  def github_next(resp)
+    # -> gitlab: x-next-page is in headers, and not empty
+    resp.headers.key?("x-next-page") && \
+      resp.headers["x-next-page"].present?
+  end
+
+  def gitlab_next(resp)
+    # -> github: Link is in headers
+    #            and if we are not on last page, we have a last link
+    resp.headers.key?("Link") && \
+      resp.headers["Link"].include?('rel="last"') && \
+      !resp.headers.key?("x-next-page")
+  end
+
   private
 
   # If user does not exist then ask for username and display_name.
@@ -70,9 +86,8 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     when "gitlab"
       if conf["group"].present?
         # Get user's groups.
-        server = conf.fetch("server", "")
-        server = server.presence || "https://gitlab.com"
-        is_member = member_of("#{server}/api/v4/groups") do |g|
+        server = conf.fetch("server", "").presence || "https://gitlab.com"
+        is_member = member_of("#{server}/api/v4/groups", per_page: 100) do |g|
           g["name"] == conf["group"]
         end
         "The Gitlab account isn't in allowed group." unless is_member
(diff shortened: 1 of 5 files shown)


kiorky:
Ok, i rebase and it's ready for review.

app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -70,7 +70,9 @@ def check_membership
     when "gitlab"
       if conf["group"].present?
         # Get user's groups.
-        is_member = member_of("https://gitlab.com/api/v4/groups") do |g|
+        server = conf.fetch("server", "")
+        server = server.presence || "https://gitlab.com"
```
mssola: I think this is simpler, no?

```ruby
server = conf.fetch("server", "https://gitlab.com")
```

app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -97,11 +99,28 @@ def github_member?(conf)
...
+                                access_token: token }.compact
+      teams.concat JSON.parse resp.body
+      # -> gitlab: x-next-page is in headers
+      # -> github: Link is in headers
+      #            and if we are not on last page, we have a last link
+      gitlab_next = resp.headers.key?("x-next-page") && \
+                    resp.headers["x-next-page"].present?
+      github_next = resp.headers.key?("Link") && \
+                    resp.headers["Link"].include?('rel="last"') && \
+                    !resp.headers.key?("x-next-page")
```
mssola: Nitpicking: Maybe you could move these `gitlab_next` and `github_next` variables intro separate auxiliar methods, since they make the loop larger than it should.

app/controllers/auth/omniauth_callbacks_controller.rb
kiorky: Nope as "" can be in conf, and we want to avoid empty string and default to "https://gitlab.com".

kiorky:
btw, sorry for style, i'm not a RoR guy :)

kiorky:
Tell me if its good, we ll rebase at the end of the review process the incremented commits.

app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -1,5 +1,19 @@
 # frozen_string_literal: true
 
+def github_next(resp)
```
mssola: Put these methods inside of the class, after a protected keyword, just like this:

```ruby
class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController

  # code

  protected

  # your methods
end
```

app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -97,11 +112,20 @@ def github_member?(conf)
...
 
   # Get user's teams and check if one match to restriction.
-  def member_of(url)
+  def member_of(url, per_page: nil)
     # Get user's groups.
     token = request.env["omniauth.auth"].credentials["token"]
-    resp = Faraday.get url, access_token: token
-    teams = JSON.parse resp.body
+    teams = []
+    # groups are paginated !
```
mssola: Remove this comment. Instead, let's put this info in the documentation of the method to something like this:

```ruby
# Get user's teams and check if one match to restriction. This method uses pagination, and the
# caller can specify the number of teams per page in the `per_page` parameter.
def member_of(url, per_page: nil)
```

mssola:
@kiorky looks nice! Now I think you can rebase it with current master and push :wink:

mssola:
Thanks a lot :clap:

mssola:
This has been cherry picked into `v2.3` already, and it will be included in the next patch-level release.

PR #1932, opened 2018-08-08
js: set shared components as global

vitoravelino:
Shared components are often used as complementary components and it's
annoying to force developers to be importing them all the time.

To avoid this task, all the components located in the shared/components
folders will be automatically recognized by vue without needing to
import them into any component definition.


vitoravelino:
> Just one question, does this add any runtime penalty? (thinking on the globals file)

Unless we end up with hundreds of shared components or something really wrong is done in one of them, it shouldn't affect anything. So, no need to worry.

PR #1937, opened 2018-08-08
ui: created <sign-out-btn> component

### LINKED ISSUE #1931: No route matches [GET] "/users/sign_out"
Signout does not get the correct url path

No route matches [GET] "/users/sign_out"

Using the tag :latest

vitoravelino:
After the removal os jquery-ujs, the sign out button stopped working.

A component to mimic the previous behavior was created and it's called
<sign-out-btn> by default in the shared/components folder.

Fixes #1931

### DIFF
--- a/app/assets/javascripts/shared/components/sign-out-btn.vue
+++ b/app/assets/javascripts/shared/components/sign-out-btn.vue
@@ -1,0 +1,39 @@
+<template>
+  <span>
+    <a class="topbar btn btn-default" id="logout" data-placement="bottom" data-toggle="tooltip" :title="title" rel="nofollow" href="#" v-bind="$attrs" @click.prevent="onClick">
+      <i class="fa fa-sign-out"></i>
+    </a>
+    <form method="post" :action="href" class="hidden" ref="form">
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/shared/components/sign-out-btn.vue
```diff
@@ -0,0 +1,40 @@
...
+
+    computed: {
+      csrfToken() {
+        return CSRF.token();
+      },
+    },
+
+    methods: {
+      onClick() {
+        this.$alert.$schedule('Signed out successfully.');
```
mssola: Maybe we could add a test that checks that this alert is actually displayed.

vitoravelino:
Fixed.

app/controllers/auth/sessions_controller.rb
```diff
@@ -35,9 +35,4 @@ def create
       flash[:notice] = nil
     end
   end
-
-  def destroy
-    super
-    flash[:notice] = nil
-  end
```
mssola: All changes that remove hacks on devise controllers is very much appreciated. Thanks! :tada:

PR #1953, opened 2018-08-31
Fix: Parsing of unauthorized header

lukaszimmermann:
### Summary

Fixed a problem when parsing the HTTP header obtained from Docker Registry
when the authentication challenge is passed to the client.

Previously, the key-value of pairs of the `www-authenticate` section were parse by
splitting by `,`. This is not safe as  the scope key can contain values such `myrepository/push,pull`, which also contain the ` ,` character. 
I noticed this because I am currently working on an feature update of Portus that also
allows the creation of tags for repositories within the GUI. Here,  the manifest needs
to be pushed to the new tag, hence the value for the scope key.

The new strategy uses scanning with a regular expression.

### DIFF
--- a/lib/portus/http_helpers.rb
+++ b/lib/portus/http_helpers.rb
@@ -106,7 +106,7 @@ module Portus
     # registry has sent the proper "WWW-Authenticate" header value that will
     # allow us the request a new authorization token for this client.
     def request_auth_token(unhauthorized_response)
-      bearer_realm, query = parse_unhauthorized_response(unhauthorized_response)
+      bearer_realm, query = parse_unauthorized_response(unhauthorized_response)
       uri = URI("#{bearer_realm}?#{query.to_query}")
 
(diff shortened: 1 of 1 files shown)


Vad1mo:
nice one 

To pass the build  check `bundle exec rubocop --extra-details --display-style-guide --display-cop-names`

lukaszimmermann:
Thanks. How can the DCO be fixed? It does not accept my real name.

vitoravelino:
> How can the DCO be fixed?

My guess would be that you're committing with a different email from the signoff (check your gitconfig) or maybe the email is not associated with your github account. I don't know tbh.

By the way, feel free to squash the commits into one.

mssola:
That's nice! Travis is also failing on a test, could you check that? Moreover, could you add a test for the case you are fixing?

lukaszimmermann:
Sorry for the delay. Yes I will do these further changes soon

PR #1958, opened 2018-09-04
Fix: add 'yarn upgrade'

Hugo1380:
### Summary

To be sure to found webpack, otherwise you can have an error who can do crashing portus webpack service.

### DIFF
--- a/examples/development/compose/bootstrap-webpack
+++ b/examples/development/compose/bootstrap-webpack
@@ -7,4 +7,5 @@ if [ $? -ne 0 ]; then
   yarn install --no-emoji --no-progress
 fi
 
+yarn upgrade
 yarn run webpack


Hugo1380:
Without that everytime when you start your services with `docker-compose up` with the `docker-compose.yml` at the root of the directory, an error occur on the portus_webpack service. And this fix seem to correct this error and all starting properly.

I really found this fix by hazard.

Hugo1380:
I will screenshot you the error, you will maybe more understand.

(I'm not native English speaker :/)

vitoravelino:
> (I'm not native English speaker :/)

Me neither. Don't worry.

I had an issue with webpack not being found a few hours ago:

```
webpack_1     | Checking dependencies...
webpack_1     | yarn run v1.5.1
webpack_1     | $ webpack --watch --config config/webpack.js
webpack_1     | sh: 1: webpack: not found
webpack_1     | error An unexpected error occurred: "Command failed.
webpack_1     | Exit code: 127
webpack_1     | Command: sh
webpack_1     | Arguments: -c webpack --watch --config config/webpack.js
webpack_1     | Directory: /srv/Portus
webpack_1     | Output:
webpack_1     | ".
webpack_1     | info If you think this is a bug, please open a bug report with the information provided in "/srv/Portus/yarn-error.log".
webpack_1     | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

but I solved it by removing the `node_modules`.

Your suggestion has side effects because it will indeed upgrade the packages based on the `package.json` constraints causing a change on `yarn.lock`, which is not ideal.

My guess is that it was just a coincidence but not the real fix.

Hugo1380:
It seem it need to be run just one time. I removed it from my `bootstrap-webpack` and it's working correctly.

I try to clean-up all containers and restart all.

Hugo1380:
I will try to update my version, but I had an issue with the new version. I will open an issue.

vitoravelino:
Ok, I'll close this one then.

PR #1962, opened 2018-09-05
[packaging] fix gem packaging

jordimassaguerpla:
Instead of adding all the gems as sources, we will use the cpio
strategy, which does not need to change the spec file or, more
important, does not need us to add/remove the .gem files into the
package every time there is a change in the Gemfile.lock file.

However, using this strategy the gems are copied on another path, and we
need to fix this.

Then, we will need to adapt the _service file in obs to have this
section

```
  <service name="download_url">
    <param name="protocol">https</param>
    <param name="host">raw.githubusercontent.com</param>
    <param name="path">/SUSE/Portus/master/Gemfile</param>
  </service>
  <service name="download_url">
    <param name="protocol">https</param>
    <param name="host">raw.githubusercontent.com</param>
    <param name="path">/SUSE/Portus/master/Gemfile.lock</param>
  </service>

  <service name="bundle_gems">
    <param name="strategy">cpio</param>
  </service>
```

A part from that, we need to add bundler-*.gem as a SOURCE in the spec
file, since the obs-service-bundle_gems does not add this gem into the
cpio.

### DIFF
--- a/packaging/suse/portus.spec.in
+++ b/packaging/suse/portus.spec.in
@@ -47,8 +47,9 @@ Source0:        %{branch}.tar.gz
 Source1:        node_modules.tar.gz
 Source2:        yarn.lock
 __PATCHSOURCES__
-
-# Add gems as sources with the https://github.com/openSUSE/obs-service-bundle_gems service
+# We need to add bundler cause https://github.com/openSUSE/obs-service-bundle_gems
+# does not add it by default
+Source100:      bundler-1.16.4.gem
 
 Requires:       timezone
 Requires:       net-tools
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
@mssola This PR is the first step to fix the packaging in master.

We will also need to do the following in https://build.opensuse.org/package/show/Virtualization:containers:Portus/portus:

 * adapt the _service file 
 * add the bundler*.gem into the package

I can do it once you approve this PR.

jordimassaguerpla:
This is what we need to add into the _service file

  <service name="bundle_gems">
    <param name="strategy">cpio</param>
  </service>

PR #1964, opened 2018-09-06
docs: improved the "First steps" page

### LINKED ISSUE #1159: Add documentation for user workflow
Add documentation for a proposed usage workflow.

mssola:
Fixes #1159


mssola:
It's not exactly what #1159 was for, but with this I think we are covered :+1:

PR #1968, opened 2018-09-13
Extended documentation for the mailer

mssola:
See #1967


mssola:
Only merge this after #1967 is merged :wink:

PR #2026, opened 2018-11-01
add how to generate self-signed certificates

eisoku9618:
I always forget the command to generate self-signed certificates, so I'm happy if README.md teaches me it.

### DIFF
--- a/examples/compose/README.md
+++ b/examples/compose/README.md
@@ -22,6 +22,14 @@ change hostname set in this file, and also in the `nginx/nginx.conf` file.
 This example is set up in a way so you can use self-signed certificates. Of
 course this is not something you would want to do in production, but this way we
 ease up the task for those who are curious to try it out.
+In order to create self-signed certificates, you could use the following command:
+
+```bash
+openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -keyout portus.key -out portus.crt
+```
+
+After that, you can simply move the ``portus.key`` and the ``portus.crt`` files
+into the secrets directory.
 
 ## The setup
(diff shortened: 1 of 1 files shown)


examples/compose/README.md
```diff
@@ -22,6 +22,9 @@ change hostname set in this file, and also in the `nginx/nginx.conf` file.
 This example is set up in a way so you can use self-signed certificates. Of
 course this is not something you would want to do in production, but this way we
 ease up the task for those who are curious to try it out.
+```bash
+openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -keyout portus.key -out portus.crt
+```
```
mssola: I'd add some introductory text, so the whole section would look like as follows:

This example is set up in a way so you can use self-signed certificates. Of course this is not something you would want to do in production, but this way we ease up the task for those who are curious to try it out. In order to create self-signed certificates, you could use the following command:

```bash
openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -keyout portus.key -out portus.crt
```

After that, you can simply move the `portus.key` and the `portus.crt` files into the `secrets` directory.

mssola:
Also, remember to sign off your work (use the `-s` flag of `git commit`) :wink:

examples/compose/README.md
eisoku9618: Thanks! Updated.

mssola:
@eisoku9618 thanks for this :+1: Now, could you squash your commits into a single one, and then rebase it on top of current master? After that, I'll gladly merge it :yum:

PR #2027, opened 2018-11-02
lib: use TRAVIS_COMMIT_RANGE for git-validation

mssola:
This should cover pretty much all cases in Travis, and it will simplify
what we've been doing so far.


mssola:
Merging since `git-validation` finally worked. Now it's less magic :sweat_smile:

PR #2030, opened 2018-11-06
api: added pagination/ordering helpers and parameters

vitoravelino:


vitoravelino:
It was a bit tricky to think about adding an isolated test for this. So what I'll be doing is adding tests for each endpoint that use these helpers.

PR #2031, opened 2018-11-06
api: added pagination/ordering to repositories

vitoravelino:

Depends-on #2030

### DIFF
--- a/lib/api/helpers/ordering.rb
+++ b/lib/api/helpers/ordering.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module API
+  module Helpers
+    # Ordering implements a helper responsible for the resource entries ordering that
+    # can be customized through request parameters like `sort_attr` and `sort_order`.
+    module Ordering
+      def order(relation)
+        relation.order(params[:sort_attr] => params[:sort_order])
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/api/helpers/ordering.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module API
+  module Helpers
+    module Ordering
```
mssola: Could you add documentation? Even if it's a bit obvious, I believe it's good practice to write what it's expected to happen when you include this module, or you use this method (e.g. the parameters that are going to be used)

lib/api/helpers/pagination.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+# Extracted and adapted from https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/api/helpers/pagination.rb
+
+module API
+  module Helpers
+    module Pagination
```
mssola: Same here

lib/api/helpers/pagination.rb
```diff
@@ -0,0 +1,56 @@
...
+      def paginate(relation)
+        return relation if params[:all]
```
mssola: Maybe it's safer:

```ruby
return relation if params[:all].present?
```

lib/api/helpers/ordering.rb
vitoravelino: Sure.

lib/api/helpers/pagination.rb
mssola: Just checked, it is not :sweat_smile: Ignore this.

lib/api/helpers/ordering.rb
vitoravelino: Done.

PR #2042, opened 2018-11-27
ui: fixed team creation for standard user

vitoravelino:
When standard user, the UI was sending a parameter through the request
that was supposed to be sent only by admin user. This was preventing
standard users from creating teams.

### DIFF
--- a/spec/system/teams_spec.rb
+++ b/spec/system/teams_spec.rb
@@ -88,6 +88,25 @@ describe "Teams support", type: :system, js: true do
 
         expect(page).not_to have_css("#team_owner_id")
       end
+
+      it "creates a team" do
+        toggle_new_team_form
+        fill_in "Name", with: "valid-team2"
+
+        click_button "Add"
+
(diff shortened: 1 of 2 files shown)


vitoravelino:
This needs to be backported to `2.4`.

mssola:
Could you add a system test (unless there's one already) controlling that if `user_permission.create_team` is disabled, this is not possible? Just to cover all possibilities.

Thanks :+1:

spec/system/teams_spec.rb
```diff
@@ -88,6 +88,24 @@
...
+
+        expect(page).to have_content("Team 'valid-team2' was created successfully")
+        expect(page).to have_link("valid-team2")
+      end
+
+      context "permission disabled" do
+        it "doesn't see link to create team" do
+          APP_CONFIG["user_permission"]["create_team"]["enabled"] = false
+
+          expect(page).not_to have_content("Create new etam")
```
mssola: Typo: "etam" -> "team" :wink:

PR #2043, opened 2018-11-27
ui: fixed namespaces duplication

vitoravelino:
Standard users were seeing duplicated namespaces entries in the
namespaces#index page.


vitoravelino:
This needs to be backported to `2.4`.

mssola:
Backported!

PR #2049, opened 2018-12-04
Allow automatic creation of unknown namespaces

### LINKED ISSUE #1961: Allow automatic creation of unknown namespaces
### Description

In some places we've found that it's actually useful to create namespaces automatically. This happens, for example, in the `sync` task when we start portus for the first time and the registry is already full of images, in which some of them are inside of namespaces. So, Portus should be able to create these namespaces, even if a team has not created it.

##### Creating a new team of administrators

I propose to create a team of administrators. All namespaces being automatically created will be created within this team. After that, after #1241 is implemented, administrators would, in turn, assign these created namespaces into proper teams whenever they feel it's needed.

##### Re-using the hidden global team

Another possibility is to simply use the global hidden team (the one created for the portus user). This has the inconvenience that there are already quite some assumptions built in the code about this team, but at least we wouldn't create yet another global admin team.

### Dependencies

- [ ] #1241: created namespaces would be assigned to the admins team, and so they should be able to be re-assigned when all pieces are put together.

vitoravelino:
Repositories with unknown namespaces from a pre-existent registry
were being ignored while syncing.
    
Now, by default, all the repositories will be considered and orphan
namespaces will be created. These namespaces can be transfered to its
respective teams later by any admin.

Fixes #1961

### DIFF
--- a/app/models/namespace.rb
+++ b/app/models/namespace.rb
@@ -58,6 +58,11 @@ class Namespace < ApplicationRecord
     name == "portus"
   end
 
+  # Returns true if namespace is orphan (no public team assigned)
+  def orphan?
+    !global && team.name.include?("global_team")
+  end
+
   # global_namespace_cannot_be_private adds an error and returns false if the
   # visibility of the global namespace is set to private. Otherwise, it returns
   # true. This function is used to validate the visibility.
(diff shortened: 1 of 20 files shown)


app/models/team.rb
```diff
@@ -30,6 +30,7 @@ class Team < ApplicationRecord
   #
   # Right now, all the special namespaces are simply marked as hidden.
   scope :all_non_special, -> { where(hidden: false) }
+  scope :global, -> { where(hidden: true).first }
```
mssola: You are assuming that they are in order, which may not be true. If you are looking for the global hidden team, then you can just use:

```ruby
find_by(name: "portus_global_team_1")
```

Note: this `1` count is valid because we don't support multiple registries.

app/models/namespace.rb
```diff
@@ -177,4 +181,21 @@ def create_delete_activities!(actor)
       }
     )
   end
+
+  def self.create_from_name!(name)
```
mssola: If the namespace didn't persist it should return nil. Note that callers from `self.get_from_repository_name` already expect this kind of behavior.

mssola: Moreover, wouldn't it be easier to simply use the create service as we do on the API?

app/models/team.rb
vitoravelino: Yeah. This is not definitive yet. Just wanted something that worked first. I'll keep that in mind.

app/models/user.rb
```diff
@@ -77,9 +77,10 @@ class User < ApplicationRecord
   has_many :tags, dependent: :nullify
   has_many :comments, dependent: :destroy
 
-  scope :not_portus, -> { where.not username: "portus" }
-  scope :enabled,    -> { not_portus.where enabled: true }
-  scope :admins,     -> { not_portus.where enabled: true, admin: true }
+  scope :portus,     -> { where(username: "portus").first }
```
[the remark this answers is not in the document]
vitoravelino: Actually declaring `portus` as a scope is a problem because scopes are intended to be chainable and always return a relation instead of the model. I've moved it to a static class method instead. `not_portus` makes sense to be like it is.

app/models/namespace.rb
vitoravelino: > If the namespace didn't persist it should return nil. Note that callers from `self.get_from_repository_name` already expect this kind of behavior.

It's already returning, no? I've changed it to `create` instead of `create!` if that was your concern.

> Moreover, wouldn't it be easier to simply use the create service as we do on the API?

We can reuse that instead of creating the activity manually.

mssola: The thing is that with `create`, you got a namespace, even if not persisted. Therefore, this method would never return nil, but just a flawed `namespace` object. That was my issue.

Now, if you use `create!`, it will actually raise an exception if something wrong happens. This is not handled neither here nor in `get_from_repository_name`. Since in the `get_from_repository_name` method the whole logic already considers nil namespaces, I think we should stick to that. Thus, I would keep using `create` here. If it didn't persist, then I'd return nil (and possibly log a debug message). Something like:

```ruby
namespace = Namespace.create!(
    name:       name,
    registry:   Registry.get,
    visibility: Namespace.visibilities[:visibility_private],
    team:       Team.global
)
return unless namespace.persisted?

# create activity and return
```

Also note that the create service also follows this logic :wink:

app/models/namespace.rb
```diff
@@ -53,11 +53,16 @@ class Namespace < ApplicationRecord
 
   scope :not_portus, -> { where.not name: "portus" }
 
-  # Returns true if this namespace belongs to the internal user "portus".
+  # Returns true if namespace belongs to the internal user "portus".
```
mssola: I think the "this" is fine. This method applies to `self` :wink:

app/models/team.rb
```diff
@@ -49,6 +49,11 @@ def member_ids
     team_users.pluck(:user_id)
   end
 
+  # Returns the main global team
+  def self.global
+    find_by(hidden: true)
```
mssola: Please, find by name instead: "portus_global_team_1".

PR #2078, opened 2018-12-20
ui: improved tag's row content

### LINKED ISSUE #2071: Improve tag row
### Description

This is part of the re-bump process of the tag/vulnerabilities features. This issue will refer to the following items:

- [x] Present vulnerabilities in a more friendly way (highlight # of high or if it's free of vulnerabilities)
- [x] Rename `Pushed at` to `Last modified`
- [x] Humanize `Last modified at` column
- [x] Tooltip with full date when hovering `Last modified`
- [x] Style checkbox

vitoravelino:
A few improvements were done regarding the tag's row. See below:

- Present vulnerabilities in a more friendly way (highlight # of high or if it's free of vulnerabilities)
- Rename Pushed at to Last modified
- Humanize Last modified at column
- Tooltip with full date when hovering Last modified
- Style checkbox

Fixes #2071 

screenshot-20181220142245-1384x327

### DIFF
--- a/app/assets/javascripts/modules/repositories/components/tags/tags-table-row.vue
+++ b/app/assets/javascripts/modules/repositories/components/tags/tags-table-row.vue
@@ -2,13 +2,22 @@
   .selected td {
     background-color: #e8f1f6;
   }
+
+  .vulns-link {
+    text-decoration: none;
+  }
+
+  .image-id {
+    font-family: monospace;
+  }
 </style>
 
 <template>
   <tr :class="{ 'selected': selected }">
     <td v-if="canDestroy">
-      <input type="checkbox" v-model="selected" @change="toggleTag()">
+      <checkbox v-model="selected" @change="toggleTag()"></checkbox>
     </td>
+
     <td>
       <tag v-for="t in tag" :key="t.name" :tag="t" :repository="repository"></tag>
     </td>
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/modules/repositories/components/tags/tags-table-row.vue
```diff
@@ -2,13 +2,22 @@
   .selected td {
     background-color: #e8f1f6;
   }
+
+  .vulns-link {
+    text-decoration: none;
+  }
+
+  .image-id {
```
mssola: Instead of this, maybe we could use the `.text-monospace` class.

app/assets/javascripts/modules/repositories/services/vulnerabilities-parser.js
```diff
@@ -1,32 +1,23 @@
...
-      if (!vulnerabilities[backend]) {
-        return;
-      }
-
-      vulnerabilities[backend].forEach((vul) => {
-        severities[vul.severity] += 1;
-        total += 1;
-      });
+    vulnerabilities.forEach((vul) => {
+      severities[vul.severity] += 1;
```
mssola: Overstretching a bit, but what if `vul.severity` returns something which is not covered by the `severities` hash (because of an update on clair or something like that)?. I think we should have a default case (maybe using `Unknown` ?).

vitoravelino: Well, these severities are based on [CVSS rating scale](https://www.first.org/cvss/specification-document#5-Qualitative-Severity-Rating-Scale). So I don't think this will be changed that often.

app/assets/javascripts/modules/repositories/components/tags/tags-table-row.vue
vitoravelino: Indeed.

PR #2087, opened 2019-01-09
tags: added size column

### LINKED ISSUE #2073: Add tag size column
### Description

This is part of the re-bump process of the tag/vulnerabilities features. This issue will refer to the following items:

- [x] Calculate size
- [x] Add column

vitoravelino:

Fixes #2073 

screenshot-20190109104338-816x236

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -109,6 +109,7 @@ class Repository < ApplicationRecord
 
   # Add the repository with the given `repo` name and the given `tag`. The
   # actor is guessed from the given `event`.
+  # rubocop:disable Metrics/MethodLength
   def self.add_repo(event, namespace, repo, tag)
     actor = User.find_from_event(event)
     return if actor.nil?
@@ -120,10 +121,11 @@ class Repository < ApplicationRecord
       repository = Repository.create(namespace: namespace, name: repo)
     elsif repository.tags.exists?(name: tag)
       # Update digest and status if the given tag already exists.
-      id, digest = Repository.id_and_digest_from_event(event, repository.full_name)
+      data = Repository.data_from_event(event, repository.full_name)
       tag = repository.tags.find_by(name: tag)
-      tag.update_columns(image_id:   id,
-                         digest:     digest,
+      tag.update_columns(image_id:   data.id,
+                         digest:     data.digest,
+                         size:       data.size,
                          scanned:    Tag.statuses[:scan_none],
                          updated_at: Time.current)
       repository.create_activity(:push, owner: actor, recipient: tag)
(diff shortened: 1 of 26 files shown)


spec/lib/portus/registry_client_spec.rb
```diff
@@ -194,7 +194,8 @@ def fetch_link_test(header)
           password
         )
 
-        _, _, manifest = registry.manifest(repository, tag)
+        _, _, _, manifest = registry.manifest(repository, tag)
```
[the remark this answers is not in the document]
vitoravelino: Headers:

```
content-length →1159
content-type →application/vnd.docker.distribution.manifest.v2+json
date →Wed, 09 Jan 2019 17:29:19 GMT
docker-content-digest →sha256:4df16fdf7389590bf7f2bedfaf19c869fa8c14d4dd3e0154ce7c04229e76bd23
docker-distribution-api-version →registry/2.0
etag →"sha256:4df16fdf7389590bf7f2bedfaf19c869fa8c14d4dd3e0154ce7c04229e76bd23"
```

Body:

```
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "config": {
        "mediaType": "application/vnd.docker.container.image.v1+json",
        "size": 2931,
        "digest": "sha256:5750958563b979a445cbefcdf4eb8831382f15781d68e4e23c6c3e23903adc28"
    },
    "layers": [
        {
            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
            "size": 2385012,
            "digest": "sha256:49388a8c9c86a6f56d228954eede699c64fce6c671a989e3e21c391859694645"
        },
        {
            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
            "size": 1334606,
            "digest": "sha256:65422b9afc83fbf5c0bc6969342745125295c35821abb87be9267e138805c777"
        },
        {
            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
            "size": 5960492,
            "digest": "sha256:e7c32889b4829112f3394a00615c46c294d972e49c09368487b82342a2d99f08"
        },
        {
            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
            "size": 33125,
            "digest": "sha256:d9025f3f03815c28ddca71445ea52ebbac31886697ac2782edc03bba5b18db99"
        }
    ]
}
```

db/migrate/20190109112643_add_size_to_tags.rb
```diff
@@ -0,0 +1,5 @@
+class AddSizeToTags < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tags, :size, :integer
```
mssola: What's the default for this? As in, how do you know that for a given tag we have never computed its size?

vitoravelino: I thought about leaving null but it can also be 0. To be honest it doesn't matter because we don't care to whatever is there. We just compute it somewhere and overwrite the value whenever we need to. Now, if it's null we display `-` as you can see in the screenshot.

lib/portus/registry_client.rb
```diff
@@ -81,6 +83,22 @@ def manifest(repository, tag = "latest")
...
+    # The json navigation is based on Image Manifest Version 2, Schema 2 that
+    # is available at https://docs.docker.com/registry/spec/manifest-v2-2/
+    def calculate_tag_size(manifest)
+      size = manifest["config"]["size"]
+
+      if manifest["layers"].present?
+        manifest["layers"].each do |layer|
+          size += layer["size"]
+        end
+      end
```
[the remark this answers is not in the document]
vitoravelino: I think we can use `reduce`. I've done that in JavaScript and it's a great approach for summing stuff.

app/models/repository.rb
```diff
@@ -120,38 +121,49 @@ def self.add_repo(event, namespace, repo, tag)
...
-        id, = Registry.get.client.manifest(repo, digest)
+        manifest = Registry.get.client.manifest(repo, digest)
       rescue ::Portus::RequestError, ::Portus::Errors::NotFoundError,
              ::Portus::RegistryClient::ManifestError => e
         logger.warn "Could not fetch manifest for '#{repo}' with digest '#{digest}': " + e.to_s
       end
     end
 
-    [id, digest]
+    OpenStruct.new(id: manifest.id, digest: digest, size: manifest.size)
```
mssola: Couldn't you just reuse the `manifest ` object? (while updating what it's needed)

vitoravelino: I thought about that but this would be one liner and not passing the manifest along. I assumed that the manifest wasn't that important for whoever was calling this. Otherwise it's no big deal to include it.

mssola: I mean, I'd assign whatever is new, and then return the object. My point is to reuse the object that we create right at the beginning: we want to avoid creating new objects if we can. I remember that the last issue we got about performance there was something about memory going up. We can avoid these kinds of situations by reusing the object and then return it.

vitoravelino: Ok, fair enough...

PR #2095, opened 2019-01-15
config: improved garbage collector options

### LINKED ISSUE #2051: Avoid that garbage_collector deletes images that are still in use
### Description

When having the newly introduced garbage_collector enabled, all tags older than the specified time are deleted, regardless if there are any (possible) used tags left.

This results in complete repositories that get deleted when images are not build regularly.

It would be a great help to ensure that tags that are still used are not deleted.
Using the `tag` parameter does not help here.

Ideas that came to my mind for this:
* check if the tag was still accessed (pulled) during the specified  `older_than` timeframe before deleting
* add an option "keep_latest_x_tags" which lets you specify to always keep the newest X tags

**Portus version**: 2.4

- [X] Implementation #2095 
- [x] Documentation #2100

vitoravelino:
So far the garbage collector only had one simple option that was to
delete images older than X days.

In this patch we are introducing a new option and a new behavior. User
will now be able to keep the latest N images through `keep_latest`
option. And the images won't be deleted even if older than X days if
they were pulled in the latest X days.

### DIFF
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -87,6 +87,9 @@ class Webhook < ApplicationRecord
     hydra.run
   end
 
+  # Pull event is not handled on Webhook yet.
+  def self.handle_pull_event(event); end
+
   # Handle a delete event from the registry. All enabled webhooks of the provided
   # namespace are triggered in parallel.
   def self.handle_delete_event(event)
--- a/config/config.yml
+++ b/config/config.yml
@@ -63,9 +63,16 @@ delete:
   garbage_collector:
     enabled: false
 
-    # Remove images older than a specific value. This value is interpreted as
-    # the number of days.
+    # Remove images not pulled and older than a specific value. This value is
+    # interpreted as the number of days.
+    #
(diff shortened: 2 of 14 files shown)


app/models/webhook.rb
```diff
@@ -87,6 +87,9 @@ def self.handle_push_event(event)
     hydra.run
   end
 
+  # Pull event is not handled nu Webhook yet
```
mssola: Typo: "nu" -> "on"

config/config.yml
```diff
@@ -63,10 +63,17 @@ delete:
...
-    # the number of days.
+    # Remove images not pulled and older than a specific value. This value is
+    # interpreted as the number of days.
+    #
+    # e.g.: If an image wasn't pulled in the latest 30 days and the image wasn't
+    # updated somehow in the latest 30 days, the image will be deleted.
     older_than: 30
 
+    # Keep the latest X images regardless if it's older than the value set in
+    # `older_than` configuration
```
mssola: nitpicking: You should add a period at the end of the sentence.

PR #2110, opened 2019-01-18
docs: added upgrading w/ compose section

vitoravelino:
A section that explains why the UI goes into a broken state when
upgrading using docker compose with nginx container from our examples.

Also shows a solution for the issue.


_docs/upgrading-portus.md
```diff
@@ -160,3 +160,23 @@ the registry to `readonly` mode before performing this operation, just to avoid
...
+
+
+### Upgrading with docker compose
+
+If you are using docker compose containing nginx to deploy Portus, you might end
+up in a broken state after an upgrade. It's a known issue [0] [1] but you should
+not worry about it. This happens because `portus` and `nginx` containers share a
+a volume for the static assets.
+
+Until we find a way to avoid that all you gotta do is remove the volumes before
```
mssola: gotta -> have to. "gotta" is too colloquial for the documentation :wink:

PR #2118, opened 2019-01-21
ui: improved wording

### LINKED ISSUE #2113: Improve wording around
There are some legacy wording around and it would be nice to improve that to avoid redundancy and verbosity in order to make the app consistent and cleaner live we've been doing recently.

Some examples:

* Add -> Create
* Create user -> Create
* Registered users -> Users
* Number of members -> Members
* Number of namespaces -> Namespaces

vitoravelino:
Mostly on forms and panel titles.

Fixes #2113

### DIFF
--- a/app/assets/javascripts/modules/users/components/application-tokens/panel.vue
+++ b/app/assets/javascripts/modules/users/components/application-tokens/panel.vue
@@ -3,7 +3,7 @@
     <h5 slot="heading-left">Application tokens</h5>
 
     <div slot="heading-right" v-if="canCreate">
-      <toggle-link text="Create new token" :state="state" state-key="newFormVisible" class="toggle-link-new-app-token"></toggle-link>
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/modules/users/components/application-tokens/panel.vue
```diff
@@ -3,7 +3,7 @@
     <h5 slot="heading-left">Application tokens</h5>
 
     <div slot="heading-right" v-if="canCreate">
-      <toggle-link text="Create new token" :state="state" state-key="newFormVisible" class="toggle-link-new-app-token"></toggle-link>
+      <toggle-link text="Create" :state="state" state-key="newFormVisible" class="toggle-link-new-app-token"></toggle-link>
```
mssola: So, what's the logic of changing "Create" to "Save" in some forms, but keeping "Create", for example, on this one? :confused:

vitoravelino: That's the toggle button and not the form submit one.

PR #2120, opened 2019-01-21
ui: improved sidebar highlight

### LINKED ISSUE #2114: Highlight current page on sidebar
Currently we only highlight the sidebar item when user is hovering them.

The idea is to make the sidebar item "marked" when users is visiting the #index and/or children pages of that resource/section.

vitoravelino:
Previously the highlight was only happening when hovering and when the
`#index` pages where selected. When the user selected a page or were
navigating in a sub-page (e.g. /repositories/1), the highlight
disappeared.

With this patch we are enforcing the highlight in every single page the
user visits. This will give more context for the user.

Fixes #2114 

screenshot-20190121171354-546x650
screenshot-20190121171404-477x538
screenshot-20190121171411-424x550

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -65,6 +65,17 @@ module ApplicationHelper
     APP_CONFIG["pagination"]["before_after"]
   end
 
+  # Returns `active` string if current path includes the passed parameter
+  # or if paremeter is `true`.
+  def active_page?(path_or_bool)
+    case path_or_bool
+    when String
+      "active" if request.fullpath.include?(path_or_bool)
+    when TrueClass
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -65,6 +65,17 @@ def pagination_before_after
     APP_CONFIG["pagination"]["before_after"]
   end
 
+  # Returns `active` string if current path includes the passed parameter
+  # or if paremeter is `true
```
mssola: \`true should be either \`true\` or without the backtick. I'd go without the backtick. Plus, a period is missing at the end.

PR #2130, opened 2019-01-25
namespace: update regular expressions to match reference

kurthaegeman:
### Summary

When creating accounts containing hyphens, Portus replaces the hyphens by underscores for the namespace name while this is [allowed by Docker Distribution](https://github.com/docker/distribution/blob/master/docs/spec/api.md). I checked the Portus code and [compared](https://rubular.com/r/DqG8LnVV8WmAUK) the intended namespace name with the regular expressions used in namespace.rb. Looks like a backslash-escape issue.


kurthaegeman:
Close. Need to fix DCO.

vitoravelino:
You don't need to close the PR. You can push force and fix the issue while you are working on it. You can add a `[WIP]` in the beginning of the PR title too to let us know that this is not ready to be reviewed.

Another thing, provide a test with the change you are requesting if there's nothing taking care of that. That helps us understand better the issue.

kurthaegeman:
Poke.

PR #2140, opened 2019-02-03
show explore link on login page when ldap auth is enabled

ross211:
When LDAP auth is enabled the 'Explore' link on the login page will never show even when anonymous browsing is enabled.

This change makes the explore link show when LDAP is enabled and also makes the login button text just 'Login' regardless of LDAP being enabled or not as Portus now supports running local users along side LDAP auth.

On a side note maybe the signup and forgot password options should be available when LDAP auth is enabled if local logins are also enabled. 

Signed-off-by: Ross Dougherty <ross@inomial.com>

### DIFF
--- a/app/views/devise/sessions/new.html.slim
+++ b/app/views/devise/sessions/new.html.slim
@@ -9,12 +9,8 @@ section.row-0
           = f.text_field :username, class: 'input form-control input-lg first', placeholder: 'Username', autofocus: true, required: true
           = f.password_field :password, class: 'input form-control input-lg last', placeholder: 'Password', autocomplete: 'off', required: true
           = f.button id: "login-btn", class: 'classbutton btn btn-primary btn-block btn-lg' do
-            - if APP_CONFIG.enabled?("ldap")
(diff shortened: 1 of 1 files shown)


mssola:
> On a side note maybe the signup and forgot password options should be available when LDAP auth is enabled if local logins are also enabled.

I'm not sure about this one, actually. If LDAP is enabled, my assumption is that the administrator wants a tight control of who is allowed to be in through the LDAP server, which acts as the authentication service. Thus, ideally, if you want to add new users, you should add them on the LDAP server. However, for exceptional cases (and for stuff like bots), we also allow administrators to create database users. These should be the exception rather than the norm. So, allowing users to sign up would simply create a clear contradiction with the LDAP server semantically speaking. Similarly, allowing people to reset passwords is not a good idea in my opinion: if you are an LDAP user, then this is not really the place to change your password; if you are not an LDAP user, you should ask the admin to do it for you.

mssola:
Thanks for this :+1:

PR #2141, opened 2019-02-03
fix style inconsistencies on password reset pages

ross211:
This fixes:
- Button font didn't match login page on both the reset and change password pages.
- 'Go back to login page' text color was wrong on change password page.

Signed-off-by: Ross Dougherty <ross@inomial.com>


mssola:
Backported into the v2.4 branch :+1:

PR #2171, opened 2019-03-25
Support oauth against custom github server (eg github enterprise)

gtallan:
### Summary

Proposed fix for issue 2169.

This set of patches works for me but has not been formally tested - I included a guess at a possible rspec test but was not sure about the correct infrastructure setup to run it.

In this patch, I assume that the github api is located at the hostname `api.{#servername}`, which is true for our github enterprise installation, but I don't know if this is universally true.


mssola:
Sorry for not getting back to you sooner (I was really busy the weeks before [susecon](https://www.susecon.com/) and during the conference itself)

It looks like the commits are not in a proper format. For more info, run:

    $ bundle exec rake test:git

Other than that, the code looks good to me.

PR #2173, opened 2019-03-26
[WIP] ui: migrate popovers from jquery to vue component

bazilio91:
Partially moving from jquery to vue implementation of bootstrap components.

New dependency: https://github.com/wxsms/uiv

### DIFF
--- a/app/assets/javascripts/bootstrap.js
+++ b/app/assets/javascripts/bootstrap.js
@@ -18,10 +18,6 @@ $(function () {
     new Vue({ el: '.vue-root' });
   }
 
-  if ($.fn.popover) {
-    $('a[rel~=popover], .has-popover').popover();
-  }
-
   if ($.fn.tooltip) {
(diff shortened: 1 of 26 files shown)


mssola:
Is this still WIP? or is it ready for review?

vitoravelino:
Hi, @bazilio91! 

Sorry for the late feedback. I really appreciate your initiative and I wouldn't mind to use `uiv` if we didn't need to include all its component together with popover. Is it possible to import only the popover component like we were doing directly with bootstrap? If not, I'm afraid we won't be able to accept this. We've been trying to avoid adding more third party dependencies in order to decrease the bundle size.

My idea around that was to create a simple popover wrapper component that would call the jquery code behind the scenes. Nothing fancy at all.

In the future if we move to modals as we've discussed internally, the popover would be replaced with sweetalert2.

I hope you understand. :smile:

